PR #85, opened 2019-06-18
Added pretties and eslint to packjson and set default settings

RubenGeo:


services/programs-service/package.json
```diff
@@ -26,6 +26,8 @@
     "@nestjs/testing": "^6.0.1",
     "@nestjs/typeorm": "^6.0.0",
     "@nestjs/websockets": "^6.0.1",
+    "@typescript-eslint/eslint-plugin": "^1.10.2",
+    "@typescript-eslint/parser": "^1.10.2",
```
elwinschmitz: These two should be also `devDependencies`.

(Easiest is to install them again with the flag `--save-dev`)

PR #179, opened 2019-08-22
Notification toast AW#121

RubenGeo:


interfaces/PA-App/src/app/services/update.service.ts
```diff
@@ -102,4 +109,20 @@ export class UpdateService {
       itemString
     );
   }
+
+  createUpdateToast(messageKey: string) {
+    this.translate.get(messageKey).subscribe((message: string) => {
```
elwinschmitz: You could use `this.translate.instant(messageKey)` (I think) so you don't need the whole `.subscribe`-callback thing...

But I'm not 100% sure.

RubenGeo: image
Yes i tried that first :(. But when i did that i only got the message key back e.g. 'TOAST.conclusion' instead of the value. I figured it was because of the reason in the image: https://stackoverflow.com/questions/46216185/ngx-translate-instant-returns-key-instead-of-value

RubenGeo: But i will recheck

RubenGeo: It's true


>     const message = this.translate.instant(messageKey);
    const closeButtonText = this.translate.instant('TOAST.close');
    this.toastController.create({
      message,
      animated: true,
      showCloseButton: true,
      closeButtonText,
      cssClass: 'update-toast',
      position: 'bottom'
    }).then((obj) => {
      obj.present();
    });

results in: 

image

PR #280, opened 2019-10-02
Remove id schema AB#756

RubenGeo:


services/121-service/examples/program-anonymous1.json
```diff
@@ -30,6 +30,26 @@
     "en": "Please follow these instructions to join our program"
   },
   "customCriteria": [
+    {
+      "criterium": "My full name:",
```
elwinschmitz: This should be `"criterium": "name",`, right?

services/121-service/examples/program-anonymous1.json
```diff
@@ -30,6 +30,26 @@
...
+      "label": {
+        "en": "My date of birth:"
```
elwinschmitz: And this then: `"en": "My name:"`

PR #308, opened 2019-10-15
Added count down timer to the AP app AB#498

RubenGeo:


RubenGeo:
@elwinschmitz  @jannisvisser  I made some code toward task 498. However due to my dockers failing. I was not able to put it toward a final test, but I think it works and it's not pretty. Since i'm not working for the rest of this week one of you could test it and merge.

PR #326, opened 2019-10-22
AB#878 (change) to (input) for numeric/text fields

jannisvisser:


interfaces/PA-App/src/app/personal-components/enroll-in-program/enroll-in-program.component.html
```diff
@@ -39,7 +39,7 @@
                    inputmode="numeric"
                    pattern="[0-9]*"
                    [name]="question.code"
-                   (change)="changeAnswers($event)"
+                   (input)="changeAnswers($event)"
```
elwinschmitz: I don't think we should be handling EVERY input, like this.

This event gets triggered on EVERY keystroke in the field, so it also triggers POST-requests on every keystroke. That will result in some weird race conditions on slow/sloppy connections, I think.

elwinschmitz: We could separate the logic of showing-next-field/submit and storing-on-the-server...

So: `onInput`-> `changeAnswer()`, `onChange` -> `storeAnswer()`, something like that maybe.

elwinschmitz: Or we could 'throttle' or 'debounce' the input-event. 

But the whole point of the `change`-event is, that the browser does all that for you already. ;)

PR #338, opened 2019-10-25
AB#725 both apps as web-apps

jannisvisser:


interfaces/AW-App/src/environments/environment.prod.ts.template.js
```diff
@@ -10,6 +10,9 @@ export const environment = {
   showDebug: ${process.env.NG_SHOW_DEBUG || 'false'},
   useAnimation: ${process.env.NG_USE_ANIMATION || 'true'},
 
+  // Extra:
+  subDirPath: ${process.env.NG_SUB_DIR_PATH_AW || '\'/AW-app\''},
```
elwinschmitz: This default value is not safe to use for the native-builds of the app(s).

elwinschmitz: + The syntax isn't what you think it should be. ;)

interfaces/PA-App/src/environments/environment.prod.ts.template.js
```diff
@@ -12,7 +12,7 @@ export const environment = {
   localStorage: ${process.env.NG_LOCAL_STORAGE || 'false'},
 
   // Extra:
-  subDirPath: '${process.env.NG_SUB_DIR_PATH}',
+  subDirPath: ${process.env.NG_SUB_DIR_PATH_PA || '\'/PA-app\''},
```
elwinschmitz: This default value is not safe to use for the native-builds of the app(s).
+ The syntax shouldn't be what you think it should be. ;)

PR #342, opened 2019-10-29
Added sms sending logging and callback function using twilio AB#901

RubenGeo:


services/121-service/src/sms/sms/sms.service.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Injectable } from '@nestjs/common';
+import { TWILIO } from '../../secrets';
+import { DEBUG, PRODUCTION_URL } from '../../config';
+import { twilioClient } from './twilio.client';
+import { TwilioMessageEntity } from './twilio.entity';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+
+let appUrl: string;
+if (process.env.NODE_ENV == 'production') {
```
elwinschmitz: This 'juggling' of the `appUrl` should all happen in the config-file, based on the ENV-variables used there.

services/121-service/src/config.ts
```diff
@@ -5,7 +5,9 @@ export const SUBDOMAIN = process.env.NODE_ENV == 'production' ? '121-service/' :
 
 const tyknIMS = 'http://11.0.0.3:50001/api/';
 const orgIMS = 'http://11.0.0.4:50002/api/';
-const userIMS = 'http://11.0.0.5:50003/api/';
+
+export const PRODUCTION_URL = 'https://production-vm.121.global/121-service/'
+export const STAGING_URL = 'https://test-vm.121.global/121-service/';
```
elwinschmitz: Similar to the use of `config.ts` in the PA-accounts-service, these URLS don't need to be hard-coded/committed.

A single `APP_URL`/`SERVICE_URL`/`PUBLIC_URL`/`EXTERNAL_URL` or `DEPLOYMENT_URL` could be set here based on an ENV-variable.

```ts
export const PUBLIC_URL = process.env.PUBLIC_URL || 'http://localhost:3000/api/'; // That last default-value would be save, but can be omitted...
```

services/121-service/src/secrets.ts.example
```diff
@@ -9,3 +9,12 @@ export const USERCONFIG = {
...
   countryIdAidWorker: 1,
 }
+
+
+export const TWILIO = {
+         sid: 'AC<your-sid>',
+         authToken: 'AC<your-sid>',
+         testToNumber: '<your-testToNumber>',
+         testFromNumber: '<your-testFromNumber>',
+         ngrok: 'http://bf235b61.ngrok.io/',
```
elwinschmitz: It wouldn't _really_ hurt, but this should also not be committed, I think.

RubenGeo: Yes that url expires after 6 hours

RubenGeo: I changed it to http://example.ngrok.io/

PR #347, opened 2019-11-01
AB#943 added username input field

jannisvisser:


interfaces/PA-App/src/app/personal-components/create-identity/create-identity.component.ts
```diff
@@ -29,35 +36,48 @@ export class CreatePasswordComponent extends PersonalComponent {
     public paData: PaDataService
   ) {
     super();
+    // this.useLocalStorage = environment.localStorage;
+    this.useLocalStorage = environment.localStorageTest;
```
elwinschmitz: Why is there another "environment"-variable required?
Isn't it just that you need to set `environment.localStorage` to `true`?

elwinschmitz: This `localStorageTest`-environment-variable is not added in the whole chain, therefore the "`--prod`"-build is failing.  :(

jannisvisser: Ah.. my bad. It is not needed in the end indeed, but I just needed it for testing, because I wanted to test the native-scenario without also changing to (non-existing) SDK-functions.. but you're right that this fails in the short-term as well, so I'll change it. @elwinschmitz

PR #348, opened 2019-11-04
PA-app: facilitate re-login on web-app

jannisvisser:


services/PA-accounts-service/src/config.ts
```diff
@@ -1,4 +1,4 @@
 export const AUTH_DEBUG = true;
 export const PORT = 3001;
-export const SUBDOMAIN = ['production', 'staging'].indexOf(process.env.NODE_ENV) > -1 ? '121-service/' : '';
+export const SUBDOMAIN = ['production', 'staging'].indexOf(process.env.NODE_ENV) > -1 ? 'PA-accounts/' : '';
```
elwinschmitz: Does the PA-accounts-service need to 'know' where it is accessed from?
Because I think this is a detail that should be "deployment-specific' and not be hard-coded like this...

I think it could also be 'passed in' via a environment-variable... (Same for the PORT or SCHEME..)

So this file would become:

```ts
export const AUTH_DEBUG = process.env.AUTH_DEBUG  || false; // Not defaulting to 'debug-mode' seems safer to me...
export const PORT = process.env.PORT || 3000; // This could be safe to include by default, I think...
export const SCHEME = process.env.SCHEME || 'http'; // Same here, but there might be already some other flag for this available.
export const SUBDOMAIN = process.env.SUBDOMAIN || ''; // It would require a string anyway, that would be fine.
```

PR #375, opened 2019-11-19
AB#1022 view program details in HO-portal

jannisvisser:


interfaces/HO-Portal/src/environments/environment.ts
```diff
@@ -6,6 +6,7 @@ export const environment = {
   production: false,
 
   // APIs
-  url_121_service_api: 'http://localhost:3000/api',
+  // url_121_service_api: 'http://localhost:3000/api',
+  url_121_service_api: 'https://test-vm.121.global/121-service/api',
```
elwinschmitz: Changes to this file only rarely need to be committed.
They will be mostly for local development only. And it shouldn't contain any specifics about our deployment in git.

jannisvisser: I know, but I always stage and then commit it by accident, when I simply commit all without staging files individually.. even though I try to think about leaving this file out of the commit. Any tips on how to change this way-of-working?

elwinschmitz: I, personally, would suggest using a GUI for staging chunks/lines only; Like https://fork.dev/ or a VSCode-extention.

Or you could explicitly 'assume the file is unchanged', by using:
```
git update-index --assume-unchanged interfaces/PA-App/src/environments/environment.ts
git update-index --assume-unchanged interfaces/AW-App/src/environments/environment.ts
git update-index --assume-unchanged interfaces/HO-Portal/src/environments/environment.ts
```
See: https://www.git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---no-assume-unchanged
Be aware that this hides changes that you DO want to commit (when adding a new values for example...)

But even better would be that Angular offered a way to have an `environment.local.ts`-file, that would be ignored by Git but loaded when-it-exists-only(without throwing errors/warnings). So you can override the defaults only when needed. This is how it works in Python/Django/Flask-apps, with a `local_settings.py`-file, for example.

Maybe we can do something like that anyway, but the annoying thing is that we have to add that custom-code then for all 3 interfaces etc... It would be nicer that it would work that way 'out-of-the-box'...

PR #376, opened 2019-11-20
AB#1023 HO-portal: make view-program-details somewhat nicer

jannisvisser:


interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
```diff
@@ -1,32 +1,108 @@
...
     });
   }
 
+
+
   public generateArray(obj) {
-    return Object.keys(obj).map((key) => ({ key, value: obj[key] }));
+    const result = [];
+    for (const key in obj) {
+      if (this.techFeatures.indexOf(key) <= -1) {
```
elwinschmitz: I think using one of the 'newer' array-functions, like `.map()` or `.filter()` (See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Iteration_methods) would make it easier to keep this method small..

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
```diff
@@ -1,32 +1,108 @@
...
+
+        const keyNew = this.translate.instant('page.programs.program-details.' + key);
+        const valueNew = this.mapLabelByLanguageCode(obj[key]);
+
+        let isArray = false;
+        if (valueNew instanceof Array) {
+          if (typeof valueNew[0] === 'object') {
+            isArray = true;
```
elwinschmitz: I think there are a few (helper-)methods within Angular/ES6 that could simplify this 3-levels-of-indentation..

Also, I think we don't really need to 'convert' the object in this way, but we can do a lot of it just in the template; By including some `*ngIf`-blocks and `*ngFor`-loops.

Because if we start adding more logic into this one method (like mentioned in the comment below), it will all become a big tangled and indented pile of code quickly.

If we define methods to do small (generic) operations, we can re-use them in all the different places we need them. All the 'specifics' will/should be contained in an 'init' or 'setup/prepare'-method only;

interfaces/HO-Portal/src/environments/environment.ts
```diff
@@ -6,7 +6,7 @@ export const environment = {
   production: false,
 
   // APIs
-  // url_121_service_api: 'http://localhost:3000/api',
-  url_121_service_api: 'https://test-vm.121.global/121-service/api',
+  url_121_service_api: 'http://localhost:3000/api',
+  // url_121_service_api: 'https://test-vm.121.global/121-service/api',
```
elwinschmitz: 😿

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
```diff
@@ -1,32 +1,108 @@
...
+    }
+
+    if (!label) {
+      label = property;
+    }
+
+    return label;
+  }
+
+  async openModal() {
```
elwinschmitz: I would advice to name things not _too generic_...
As this will have some specific function or goal, so `showProgramProperties` or `openProgramJson` would be better. Because why would this method-name need to change if we want to open the details/json in a different component?

That way names-of-things will be long, bug not confusing. For example "ProgramModalPage": Is it a page? Is it a Modal? 🤷‍♂  -> It is a "thing that displays program-properties".

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
```diff
@@ -1,32 +1,108 @@
...
+    const modal: HTMLIonModalElement =
+      await this.modalController.create({
+        component: ProgramModalPage,
+        componentProps: {
+          program: this.program,
+        }
+      });
+
+    modal.onDidDismiss().then((detail: OverlayEventDetail) => {
```
elwinschmitz: If nothing needs to happen when it is closed, we don't really need to define anything for it, I think.

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
```diff
@@ -1,32 +1,108 @@
...
+            // Enter code here to visualize array-properties (like Criteria/Aidworkers) differently
+          }
+        }
+        result.push(({ key: keyNew, value: valueNew, isArray }));
+      }
+    }
+    return result;
+  }
+
+  private mapLabelByLanguageCode(property: any) {
```
elwinschmitz: Note-to-self: We need to find a way to easily share this code across all interfaces. As copy-&-pasting this to everywhere we need this is going to create annoying bugs one day...

interfaces/HO-Portal/src/app/programs/program-list/program-list.component.html
```diff
@@ -4,7 +4,7 @@
       <ion-menu-button></ion-menu-button>
     </ion-buttons>
     <ion-title>
-      {{ 'page.programs.PROGRAMS'|translate }}
+      {{ 'page.programs.program-list.title'|translate }}
```
elwinschmitz: \o/ Yay! Way more specific and nice.

But the other page had `pageTitle` as the last element; Could that be the same everywhere?

interfaces/HO-Portal/src/app/programs/program-modal/program-modal.module.ts
```diff
@@ -0,0 +1,29 @@
+import { NgModule } from '@angular/core';
```
elwinschmitz: Is it necessary that the `program-modal`-component is a (Angular-)'page' that needs its own `*.module.ts`-file?
It feels very verbose... :/

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
jannisvisser: adjusted

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
jannisvisser: Sure, but I'll leave this until it's actually worked on. Right now it's not used.

interfaces/HO-Portal/src/environments/environment.ts
jannisvisser: I know.. for now I have adjusted your assume-unchanged approach.. but not ideal either.

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
jannisvisser: Point taken. Changed openModal fn-name. (I took this directly from an example and didn't change it.) And changed name of programModalPage.

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
jannisvisser: removed

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.ts
jannisvisser: +1

interfaces/HO-Portal/src/app/programs/program-list/program-list.component.html
jannisvisser: changed everywhere.

interfaces/HO-Portal/src/app/programs/program-modal/program-modal.module.ts
jannisvisser: I changed it to a component. I did need to add this (empty) route to programs-routing.module.ts to get it working..

```
  {
    path: '',
    component: ProgramJsonComponent,
    canActivate: [AuthGuard],
  }
```

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.html
```diff
@@ -4,14 +4,25 @@
       <ion-menu-button></ion-menu-button>
     </ion-buttons>
     <ion-title>
-      {{ 'page.programs.PROGRAM_DETAILS'|translate }}
+      <ion-icon name="arrow-round-back" [routerLink]="['/programs']"></ion-icon>
```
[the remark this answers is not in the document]
jannisvisser: you already changed this right?

interfaces/HO-Portal/src/app/programs/program-modal/program-modal.module.ts
jannisvisser: Update: this actually made the empty route (localhost:4200) go to ProgramJSON.. I changed this now to some random route (path: 'program-json'), so it doesn't interfere. But this is a weird solution. The point is that I shouldn't need a route for a modal, but somehow it seems to require this of me.. any ideas @elwinschmitz ?

elwinschmitz: Ah, that explains 'bug' AB#1098 👍 

I'll see what I can look into.. But not a clear-cut-solution right away...

interfaces/HO-Portal/src/app/programs/program-details/program-details.component.html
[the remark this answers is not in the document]
elwinschmitz: Yeah, 80%.

But it turns out it isn't fully doing what I thought it would do... So I've refatored it into a 'proper' button into its 'proper' place....

PR #378, opened 2019-11-20
removed cordova global-121-indy plugin

jannisvisser:
@elwinschmitz I removed this - now unused (right?) - plugin, because it gave trouble in the 'npm ci' step in the webhook on the server. If I shouldn't have taken this out, or should have done so differently, we can always change it back.

### DIFF
--- a/interfaces/PA-App/package-lock.json
+++ b/interfaces/PA-App/package-lock.json
@@ -3929,10 +3929,6 @@
       "integrity": "sha512-Jb3V72btxf3XHpkPQsGdyc8N6tVBYn1vsxSFj43fIz9vonJDUThYPCJJHqk6PX6N4dJw6I4FjxkpfCR4LDYMlw==",
       "dev": true
     },
-    "cordova-plugin-global-121-indy": {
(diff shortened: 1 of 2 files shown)


elwinschmitz:
If you have run: `npm run ionic -- cordova plugin remove cordova-plugin-global-121-indy` AND `npm run cordova plugins save`; then it would be enough...

But I think there should also be changes to the `config.xml`-file...

I'll check when I update my environment.

elwinschmitz:
It currently breaks the Android-build:
https://dev.azure.com/global121/121%20Platform/_build/results?buildId=1632

You would be able to test it (before merging) by commenting on this PR with the text:
`/AzurePipelines run PA-App -- Android`; 

But I see now that the text in the README(https://github.com/global-121/121-platform/tree/master/interfaces#continuous-integration-ci) doesn't explain this,,,

elwinschmitz:
I think the cleanest would be to revert this PR and do it again with all parts removed in 1 go;

That way we can 'save' that PR for later reference if we want to enable it again. (We can attach it to the Backlog-items we already have for after the first pilot(s).

PR #379, opened 2019-11-20
remove 121-indy plugin from config.xml

jannisvisser:


elwinschmitz:
This PR now does 2 totally different things. :(

It should only do 1 and be connected the the appropriate back-log items. That way, when we need to 'redo' or 'revisit' some of this work we can find things again.

See: https://github.com/global-121/121-platform/pull/378#issuecomment-555978439

elwinschmitz:
Partly salvaged in #384.

PR #381, opened 2019-11-20
Include security/performance optimisations/best-practices in server-config

elwinschmitz:
+ Fix handling of URLs used by Angular router

Fixes [AB#1087](https://dev.azure.com/global121/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/1087)

### DIFF
--- a/tools/121-platform-https.conf
+++ b/tools/121-platform-https.conf
@@ -65,7 +65,7 @@
     <IfModule mod_headers.c>
       Header set X-UA-Compatible "IE=edge" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
       Header set X-Frame-Options "DENY" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
-      Header set Content-Security-Policy "default-src 'self'; style-src 'self' 'unsafe-inline'; base-uri 'self'; form-action 'self'; frame-ancestors 'none'; upgrade-insecure-requests" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
+      Header set Content-Security-Policy "default-src 'self' data: ; style-src 'self' 'unsafe-inline'; base-uri 'self'; form-action 'self'; frame-ancestors 'none'; upgrade-insecure-requests" "expr=%{CONTENT_TYPE} =~ m#text/html#i"
       Header set X-Content-Type-Options "nosniff"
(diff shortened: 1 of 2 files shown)


elwinschmitz:
@jannisvisser I'm not sure what the easiest/safest way to test this is? I would only merge this if it is set && tested on the test/production servers. 
But I assume the files in this repository are not 'connected' to the ones on the server automatically, right?

I also think that not ALL interfaces are 'automatically deployed on merge' yet, right?

jannisvisser:
But to test this on the servers I need to merge it right? Or can I pull an unmerged PR as well? (and is that what you imply?)

No, these files need to be copied to the proper Apache-folder first indeed.

Not all interfaces are included in the webhook indeed. I just added HO-portal, because we'll be making more changes to it now. But I left AW-app still out.

elwinschmitz:
I thought that as the files need to be updated "by hand" anyway, you could do that first, to verify everything works as we want; and then merge as a sort of 'backup/documentation' only..

And we can test the workings of the `/programs/1` and /login urls right away then; As this PR includes a fix for the 'redirect-hack' that is currently used.

jannisvisser:
@elwinschmitz 
- the https-code is not allowed by Apache2. Specifically the lines with TraceEnable and ServerTokens need to be commented out for the code to be accepted.
- I have Apache version 2.4.27 on the server, which seems to be line with what is mentioned in the link you used (https://github.com/h5bp/server-configs-apache): support for version 2.4.0+
- However, when I run the new code without these 2 lines, it doesn't seem to work. When I do Ctrl+shift+R on /login I still get a 404.

elwinschmitz:
@jannisvisser 
These two lines can be removed then. 

For the URLs to work there might be other configuration-options required then... But let’s try to fix that in another PBI/bug in the future. 

The other options are good defaults anyway. 

And maybe there is a way to easily test this locally first..

elwinschmitz:
Maybe you could also try:

```
FallbackResource /index.html
```
?

jannisvisser:
No, that doesn't work either

elwinschmitz:
@jannisvisser: I've managed to tweak the options by "trying it out" on the staging server. ¯\_(ツ)_/¯

jannisvisser:
Nice. So I should review by also trying on the production server then? (because the staging-server it's already on the new settings)

elwinschmitz:
I just did, but ran into another small tweak that is necessary..

PR #393, opened 2019-11-26
AB#1089 HO-portal process PR#376 comments

jannisvisser:


elwinschmitz:
See #376

interfaces/HO-Portal/src/app/programs/programs-routing.module.ts
```diff
@@ -16,6 +17,11 @@ const programRoutes: Routes = [
     path: 'program/:id',
     component: ProgramDetailsComponent,
     canActivate: [AuthGuard],
+  },
+  {
```
elwinschmitz: @jannisvisser  What if we "just don't include it" here?

PR #396, opened 2019-11-26
Made password hidden when returning aidworkers AB#1099

RubenGeo:


RubenGeo:
Thanks for the faith :D

services/121-service/src/programs/program/program.controller.spec.ts
```diff
@@ -35,6 +35,7 @@ const newProgramParameters = {
   highestScoresX: 500,
   meetingDocuments: JSON.parse('{"en": "documents"}'),
   customCriteria: [],
+  notifications: JSON.parse('{}')
```
elwinschmitz: Why all the `JSON.parse`-uses?

That just literally 'removes the quotes'? Why not create the whole object as a real object?

PR #400, opened 2019-12-02
#AB1090 AW-app: change registration claims

jannisvisser:


interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
```diff
@@ -15,35 +17,202 @@ import { SessionStorageService } from 'src/app/services/session-storage.service'
...
+      answer.label = this.getAnswerOptionLabelByValue(question.options, answerValue);
+    }
+
+    this.answers[questionCode] = answer;
+
+    this.checkAllQuestionsAnswered(this.answers);
+
+  }
+
+  public initialAnswers(answersProgram) {
```
elwinschmitz: We should amend `changeAnswers`, so that we can re-use it in here for the inital-answers-loop...

interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
```diff
@@ -15,35 +17,202 @@ import { SessionStorageService } from 'src/app/services/session-storage.service'
...
+    }
+    this.hasAnswered = true;
+    this.changedAnswers = false;
+    this.dobFeedback = false;
+  }
+
   public getPrefilledAnswersProgram() {
     this.programsService.getPrefilledAnswers(this.did, this.programId).subscribe(response => {
       this.answersProgram = response;
+      this.initialAnswers(this.answersProgram);
```
elwinschmitz: What type of object is `answersProgram`? Not really a `Program`, right?

We could rename or get rid of the whole public property altogether?

interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
```diff
@@ -84,3 +254,29 @@ export class ValidateProgramComponent implements ValidationComponent {
   }
 
 }
+
+
+class Question {
```
elwinschmitz: I'm not sure how much would make sens, but all these 'models' could be moved to separate files that can be (maybe) easier to keep in sync between the PA-app and the AW-app...

interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
jannisvisser: adjusted

interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
jannisvisser: adjusted

interfaces/AW-App/src/app/validation-components/validate-program/validate-program.component.ts
jannisvisser: Makes sense. But in the 'cutoff' between 'process feedback right away' and 'make new task', I think this one falls in the latter? And if agree, new task within this sprint/PBI? Or on backlog?

jannisvisser:
This was the code review, are you also doing a functional review (like discussed yesterday)?

PR #415, opened 2019-12-10
Store custom data in connection AB#1163 AB#1158  AB#1153

RubenGeo:


services/121-service/src/sovrin/create-connection/connection.entity.ts
```diff
@@ -34,4 +34,9 @@ export class ConnectionEntity {
     nullable: true,
   })
   public programsExcluded: number[];
+
+  @Column('json', {
+    default: {}
+  })
+  public customData: JSON;
```
elwinschmitz: @RubenGeo @jannisvisser: I still think we should revisit these type-declarations.
As I think it doesn't do/represent what you think it does/represents...

"`JSON`" is not part of the basic types TypeScript uses: https://www.typescriptlang.org/docs/handbook/basic-types.html

I think it should just be `object` in most (if not all) cases...
https://www.typescriptlang.org/docs/handbook/basic-types.html

elwinschmitz: Even better would be to define all these `object`s, so we can name/reference them properly and get nice autocomplete/tooltips available in our editors.

PR #437, opened 2019-12-18
121-service added endpoint that return total number of included AB#1221

RubenGeo:


services/121-service/src/programs/program/program.service.ts
```diff
@@ -500,30 +503,20 @@ export class ProgramService {
...
-    const includedConnections = [];
-    for (let connection of connections) {
-      if (connection.programsIncluded.includes(programId)) {
-        includedConnections.push(connection);
-      }
-    }
-    if (includedConnections.length < 1){
+    if (includedConnections.length < 1) {
       const errors = 'There are no included PA for this program';
       throw new HttpException({ errors }, 404);
```
elwinschmitz: "0 included" could be a valid response, right? At least not a 404... (see below)

services/121-service/src/programs/program/program.service.ts
```diff
@@ -500,30 +503,20 @@ export class ProgramService {
...
     }
 
-    const availableFunds = await this.fundingService.getProgramFunds(programId)
-    const fundsNeeded = amount * includedConnections.length
+    const availableFunds = await this.fundingService.getProgramFunds(programId);
+    const fundsNeeded = amount * includedConnections.length;
     if (fundsNeeded > availableFunds) {
       const errors = 'Not enough available funds';
       throw new HttpException({ errors }, 404);
```
elwinschmitz: This could be an error, but 404 doesn't sound the best fit...

Because technically the request would be well-formed and successfully transmitted; But the outcome of the calculation is just "not valid" or "not good enough".

So 200, with a `{ status: 'error', message: 'Insufficient funds' }`; Or similar...

services/121-service/src/programs/program/program.service.ts
RubenGeo: This part is in the payout function. So if there are 0 included nothing would happen if you press payout.

services/121-service/src/programs/program/program.service.ts
RubenGeo: Created: https://github.com/global-121/121-platform/compare/master...RubenGeo:statusPayout?expand=1

PR #478, opened 2020-01-14
AB#1291 create popup for info in select-language

jannisvisser:


interfaces/PA-App/src/app/personal-components/select-language/select-language.component.html
```diff
@@ -2,7 +2,10 @@
   <dialogue-turn>
     <p [innerHTML]="'personal.select-language.introduction'|translate"></p>
     <p>
-      <ion-button type="button" size="small">
+      <ion-button type="button"
```
elwinschmitz: Once we add a "MoreInfoButtonComponent", we should be able to include a `<more-info-button>`-element here, with an attribute of the translated message:

```
<more-info-button [message]="{{ 'personal.select-language.info-message'|translate }}">
  {{ 'personal.select-language.more-info'|translate }}
</more-info-button>
```
(So the more-info-button also contains the icon)

interfaces/PA-App/src/app/personal-components/select-language/select-language.component.spec.ts
```diff
@@ -2,6 +2,7 @@ import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { TranslateModule } from '@ngx-translate/core';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
+import { PopoverController } from '@ionic/angular';
```
elwinschmitz: This should not be necessary when using the "MoreInfoButtonComponent" via the "SharedModule"..

interfaces/PA-App/src/app/personal-components/select-language/select-language.component.ts
```diff
@@ -6,6 +6,9 @@ import { PaDataService } from 'src/app/services/padata.service';
 import { TranslateService } from '@ngx-translate/core';
 import { ConversationService } from 'src/app/services/conversation.service';
 
+import { PopoverController } from '@ionic/angular';
```
elwinschmitz: This should not be necessary when using the "MoreInfoButtonComponent" via the "SharedModule"..

interfaces/PA-App/src/app/helpers/openInfoPopup.ts
```diff
@@ -0,0 +1,13 @@
+import { InfoPopupComponent } from '../info-popup/info-popup.component';
+import { PopoverController } from '@ionic/angular';
+
+export async function openInfoPopup(popoverController: PopoverController, message: string) {
```
elwinschmitz: This should move into the MoreInfoButton-component as a 'local' method.

The `message` would be a `@Input()`-property that you can pass into the `popoverController` then

interfaces/PA-App/src/app/personal-components/select-language/select-language.component.html
elwinschmitz: The `translate`-magic should be, I think, some other syntax... But the main goal is to insert the 'already-translated-text' into the component, so it doesn't have to worry about translating it....

But on the other hand, it needs to load the TranslateService anyway to be able to translate the pop-up-title anyway... So might as well pass in only the 'key to be translated' (`personal.select-language.info-message`)

PR #531, opened 2020-02-06
Fix docker envs

elwinschmitz:
I only got it to work when I followed these steps:

- Checkout repo / git pull etc.
- Create `.env`-files, `secrets.ts`-files and `ormconfig.json`-files from examples (as described in all the READMEs)
- Create `Dockerfile`s for 121-service & PA-accounts-service (WITHOUT enabling any CMD line!)
- Run `docker-compose up -d --build` from the `services/`-folder
- Edit both `Dockerfile`s to enable the `CMD ["npm", "start"]`-line
- Run `docker-compose up -d --build 121-service PA-accounts-service`

I HAD to do the `up`-command separately for "services depending on the db";
I Couldn't get it working in one go. :(
As both services now need a working connection to the DB, but the initial 'up'/'build' command is too fast.


jannisvisser:
@elwinschmitz There's nothing in your flow which I recognize / can point out what's wrong.. For me the local environment still works fine (with latest pull from master, but without updating any of the secrets files)

elwinschmitz:
@jannisvisser My main issue/concern is, that our current set-up doesn't really work "from scratch". It only works without error if you have a previously set-up environment.

So I can't easily use: "when in doubt, just turn everything off an on again" as a fast debugging strategy.

2 things should be fixed/changed, I think:
## 1) Migrations
The command `npm run typeorm migrations:run` should be able to run on any db-setup and not break. (Should fail silently if there is nothing to migrate or something...)  
Or the order of commands should be correctly hard-coded in. As in this PR:  (pre-start): `run seed-prod` → then `run migrations` → then start: `npm start'  
If that is save to do, then we can merge this PR and call this issue 'resolved'.

## 2) Startup/build order of containers
Both the `121-service` and `PA-accounts-service`-containers now depend on the `121db`-container to be up-and-running for their start-up sequences.
We could accept the workaround and document it like that in the /services/README-file (as I outlined in this PR's description).
Or;
We should look into a way to let these 2 containers 'wait' on the database, as described in: https://docs.docker.com/compose/startup-order/.
Because the configuration `depends on: 121db`, as I added in this PR, is not enough to get it working from scratch. I haven't tried all of the suggestions with these 'wait'-scripts, but they sound like an easy solution to this.

jannisvisser:
@elwinschmitz agree, let's make a backlog-item for it then.

Idea that pops in now: shouldn't the "restart: unless-stopped" help fixing this thing? Because if 121-service keeps failing until 121db is up, then it should keep restarting until it works. (But apparently that doesn't solve it.)

elwinschmitz:
@jannisvisser Letting both services retry-untill-it-works could be a workaround indeed.

They should exit with a proper error-message explaining that you could maybe wait.
So that you would see in the logs, something like: "121-service: Cannot connect to database, exiting... Restart to retry."

Because currently it wasn't working, also not after a long time, AND it wasn't quickly clear to me what the problem was..

elwinschmitz:
Any future work will be in: AB#1448

PR #566, opened 2020-02-26
Created some sound tests to check our options AB#1482

RubenGeo:


elwinschmitz:
This will be useful for AB#823

PR #578, opened 2020-03-03
Made pa metric accumalative AB#1551 AB#1552

RubenGeo:


interfaces/HO-Portal/src/app/program/program.component.html
```diff
@@ -36,12 +36,12 @@
       </ion-col>
     </ion-row>
     <ion-row>
-      <ion-col size="auto">
+      <ion-col size=3>
```
elwinschmitz: This doesn't need to be committed together with the other changes. :|

interfaces/HO-Portal/src/app/models/program-metrics.model.ts
```diff
@@ -5,8 +5,9 @@ export class ProgramMetrics {
 }
 
 export class PeopleMetrics {
-  pendingVerification: number;
-  verifiedAwaitingDecision: number;
+  startedEnlisting: number;
```
elwinschmitz: Why "Enlisting" as the verb?

We use "Enroll" in all kinds of other places?

interfaces/HO-Portal/src/app/program/program.component.html
RubenGeo: Shall I remove it?

interfaces/HO-Portal/src/app/models/program-metrics.model.ts
RubenGeo: I did that because we have a column in the back-end that is called programsEnrolled which actually contains the programs for which you are verified. I therefore i wanted to use a different term for this field to avoid confusion.

RubenGeo: Maybe we can use register instead of enlist? That also feels like it fit's more into the conext of internet accounts.

PR #581, opened 2020-03-03
AB#1406 HO people list

jannisvisser:


interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.ts
```diff
@@ -17,15 +17,14 @@ import { ProgramPhase } from 'src/app/models/program.model';
...
   @Input()
   public selectedPhase: string;
-
   @Input()
   public transferValue: any;
-
   @Input()
   public currencyCode: string;
+  @Output()
+  emitCompleted: EventEmitter<boolean> = new EventEmitter<boolean>();
```
elwinschmitz: I would suggest to name these emitters not with an `emit` verb.
Something like `isCompleted` makes more sense when you use something to 'listen' to this event.

interfaces/HO-Portal/src/app/program/phase-next/phase-next.component.html
```diff
@@ -1,5 +1,13 @@
-  <ion-button *ngIf="btnAvailable"
-              [disabled]="selectedPhase !== activePhase"
-              (click)="advancePhase(activePhaseId)"
-              [innerHTML]="btnText"
-  ></ion-button>
+<confirm-prompt *ngIf="btnAvailable"
+                [disabled]="selectedPhase !== activePhase || !phaseReady || isInProgress || currentUserRole !== userRoleEnum.ProgramManager"
```
elwinschmitz: I would put all this logic into a specific method/property IN the `phase-next.component.ts`;
That way not everything needs to be 'made public' to the template as well. (things like the Role, the userRoleEnum, etc.)

interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.ts
```diff
@@ -196,6 +203,19 @@ export class ProgramPayoutComponent implements OnChanges {
 
     await alert.present();
   }
+
+  private checkPhaseReady() {
+    // This component only influences ready in the 'payment'-phase
+    if (this.activePhase === ProgramPhase.payment) {
```
elwinschmitz: Often using a 'return as early as possible'-approach makes for a short and readable method.

In this case: 
```
if (this.activePhase !== ProgramPhase.payment) {
  return this.emitCompleted.emit(true);
}
// from this point down there is only 1 'thruth' and no 'else' to worry about/take care of all these dangling `}`'s.
```
This can be done multiple times in every method. It saves you from having multiple levels of indentation and make for a better 'overview' of what happens inside a method.

interfaces/HO-Portal/src/app/program/program-people/program-people.component.html
```diff
@@ -1,30 +1,38 @@
 <section *ngIf="componentVisible">
   <h2>{{ 'page.program.program-details.section-people'|translate }}</h2>
   
-  <p *ngIf="!enrolledPeople?.length && !showSensitiveData">
+  <p>
+    {{ 'page.program.program-people.targetNumber'|translate }} {{program?.highestScoresX}}
+  </p>
```
elwinschmitz: Not all programs will have the property `highestScoreX` (only programs of that 'flavour').
So the template should support all possible versions of a `program`, so "only show this if this feature/property is available".

```suggestion
  <p *ngIf="program?.highestScoresX">
    {{ 'page.program.program-people.targetNumber'|translate }} {{program?.highestScoresX}}
  </p>
```

interfaces/HO-Portal/src/app/program/program-people/program-people.component.html
```diff
@@ -1,30 +1,38 @@
...
+  <p *ngIf="!enrolledPeople?.length && !showSensitiveData && selectedPhase !== 'design'">
```
elwinschmitz: Is this component even visible/available in the `design`-phase?

interfaces/HO-Portal/src/app/program/program-people/program-people.component.html
```diff
@@ -36,7 +44,7 @@ <h2>{{ 'page.program.program-details.section-people'|translate }}</h2>
     {{ 'page.program.program-people.submitPrivacy'|translate }}
   </confirm-prompt>
 
-  <confirm-prompt [disabled]="!enrolledPeople.length"
+  <confirm-prompt [disabled]="!newEnrolledPeople.length || !btnEnabled"
```
elwinschmitz: I think including the condition of `!newEnrolledPeople.length` into the decision for `btnEnabled` would keep all the logic in 1 place... Instead of here AND in the component-file with `btnEnabled`.

interfaces/HO-Portal/src/app/program/program-people/program-people.component.ts
```diff
@@ -106,86 +224,96 @@ export class ProgramPeopleComponent implements OnChanges {
...
     }
+    return paymentColumns;
   }
 
-  private createTableData(source: Person[]): Person[] {
+  private async createTableData(source: Person[]): Promise<Person[]> {
     if (source.length === 0) {
       return [];
     }
+    console.log(source);
```
elwinschmitz:
```suggestion
```

interfaces/HO-Portal/src/app/program/program-people/program-people.component.ts
```diff
@@ -200,38 +328,63 @@ export class ProgramPeopleComponent implements OnChanges {
...
           did: person.did,
-          created: formatDate(person.created, this.dateFormat, this.locale),
-          updated: formatDate(person.updated, this.dateFormat, this.locale),
+          processStarted: formatDate(person.created, this.dateFormat, this.locale),
+          digitalIdCreated: person.appliedDate ? formatDate(person.appliedDate, this.dateFormat, this.locale) : null,
+          digitalIdValidated: person.validationDate ? formatDate(person.validationDate, this.dateFormat, this.locale) : null,
+          vulnerabilityAssessmentCreated: person.appliedDate ? formatDate(person.appliedDate, this.dateFormat, this.locale) : null,
+          vulnerabilityAssessmentValidated: person.validationDate ? formatDate(person.validationDate, this.dateFormat, this.locale) : null,
+          inclusionCommunication: person.inclusionDate ? formatDate(person.inclusionDate, this.dateFormat, this.locale) : null,
+          included: person.included ? 'Included' : (person.excluded ? 'Excluded' : ''),
```
elwinschmitz: These strings should be translatable, or event only be set in the template, not here. (preferably)

interfaces/HO-Portal/src/app/program/program-people/program-people.component.ts
```diff
@@ -200,38 +328,63 @@ export class ProgramPeopleComponent implements OnChanges {
...
         };
 
+        this.paymentColumns.map((_, index2) => {
```
elwinschmitz: The lint-rule to not have `shadow variables/parameters` isn't there to force you to pick a "name-I-want+1" as a variable name. ;)

It is to avoid doing to much of the same (looking) things in 1 method.
Everything that happens in a `Array.map`-callback, could be done in its own separate, named method.

That way you can use `index` as you variable/parameter for the array-index.

elwinschmitz: Also, if a `map`-callback doesn't return a value, it isn't really a mapping, but more a `for`-loop/`forEach`-callback.

interfaces/HO-Portal/src/app/program/program-people/program-people.component.ts
```diff
@@ -200,38 +328,63 @@ export class ProgramPeopleComponent implements OnChanges {
...
 
         return personData;
       });
   }
 
-  private defaultSelectedPeople(source: Person[]): Person[] {
-    if (this.program.inclusionCalculationType === InclusionCalculationType.highestScoresX) {
-      const nrToInclude = this.program.highestScoresX;
+  private defaultSelectedPeople(source: any[]): any[] {
+    if (this.selectedPhase === ProgramPhase.inclusion) {
```
elwinschmitz: See previous "early return"-comment.

interfaces/HO-Portal/src/app/program/program-people/program-people.component.ts
```diff
@@ -200,38 +328,63 @@ export class ProgramPeopleComponent implements OnChanges {
...
+    } else {
+      return [];
+    }
+  }
 
-    return source.filter((person) => person.score >= minimumScore);
+  private defaultSelectedPeoplePrivacy(source: any[]): any[] {
+    if (
+      this.selectedPhase === this.activePhase &&
+      [ProgramPhase.inclusion, ProgramPhase.finalize, ProgramPhase.payment].includes(ProgramPhase[this.selectedPhase])
```
elwinschmitz: I wonder if we can move all these types of logic into a more central place...

We should/could then create multiple methods (side-by-side) that only do 1 thing (for example: `filterOnlyIncluded()`).
And then 1 'controller'/'init'-method picks and chooses from these available methods.

interfaces/HO-Portal/src/app/program/program.component.html
```diff
@@ -41,20 +42,23 @@
         <app-metrics [program]="program"></app-metrics>
 
       </ion-col>
+
       <ion-col size=9>
 
+        <app-program-people [selectedPhase]="selectedPhase"
+                      [programId]="program?.id"
+                      [userRole]="currentUserRole"
+                      (emitCompleted)="emitPeopleCompleted($event)"
```
elwinschmitz: Re: renaming the EventEmitter-properties:

This would then become:
```
(isCompleted)="onPeopleCompleted($event)"
```

So they both NOT include the 'emit' verb. Because the action they perform is not the same. (only 1 emits a thing actually).

features/HO-Portal/Manage_persons_affected.feature
```diff
@@ -0,0 +1,51 @@
+@ho-portal
+Feature: Manage persons affected connected to a program
```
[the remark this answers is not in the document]
jannisvisser: done

PR #636, opened 2020-04-07
Fix bug for select-language intro + Remove need for HTML

elwinschmitz:
Related [AB#1694](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/1694)


elwinschmitz:
@jannisvisser @RubenGeo 
I think we should do a similar exercise for the other interfaces (all in 1 go, maybe) to remove the HTML from the translations-file(s).

jannisvisser:
@elwinschmitz good idea. just to check: for the more-info-popups we will still have HTML in the jsons right?

elwinschmitz:
Yes, we still need to find a pretty solution for that. 
But that is already on the backlog somewhere....

PR #639, opened 2020-04-08
Show NGO logo

jannisvisser:


interfaces/PA-App/src/app/shared/dialogue-turn/dialogue-turn.component.html
```diff
@@ -14,9 +14,16 @@
     <ion-avatar *ngIf="isSelf" style="background:#2f4bfd; display:flex;">
       <ion-icon name="person" color="light" size="large" style="margin:auto;"></ion-icon>
     </ion-avatar>
-    <ion-avatar *ngIf="!isSystem && !isSelf" style="color:black; background:white; display:flex;">
-      <b style="margin:auto">{{ actor }}</b>
-    </ion-avatar>
+    <div *ngIf="!isSystem && !isSelf && actor === 'NGO A'; else noLogo">
```
elwinschmitz, line 17: When we explicitly check in the template for the string-value of `actor`, we can do anything we want inside this `ngIf`-block, so also use any filename we want.

Also hard-coding the filename is better then using the string mashup of line 19.

interfaces/PA-App/src/app/shared/dialogue-turn/dialogue-turn.component.html
```diff
@@ -14,9 +14,16 @@
...
+      <ion-avatar style="border-radius: 0 !important;">
+        <img src="./assets/{{actor}}.png">
```
elwinschmitz, line 19: For this use-case we should use `"./assets/logos/ngo-a_white_512x512.png"` (or anything that makes technical sense to manage these few logo-files)

But in general, when using variables in HTML-attributes in an Angular-template, you should use the square-brackets syntax:
```html
<img  [alt]="actor" [src]="'./assets/' + actor + '.png'">
```
Also, the `alt`-attribute should always be added to `img`-elements, to give a textual representation of the image. So in this case, the name of the NGO would be enough.
If the image is only for visual fanciness, `alt=""` should be used (so an empty string).

interfaces/PA-App/src/app/shared/dialogue-turn/dialogue-turn.component.html
elwinschmitz: Also, this error is thrown when you try to build the interface with the `--prod` flag.

```
ERROR in src/app/shared/dialogue-turn/dialogue-turn.component.html(17,10): : This condition will always return 'false' since the types 'Actor' and '"NGO A"' have no overlap.
```

So I think you need to include the ngo's we want into the `Actor`-enum in `dialogue-turn.component.ts`

elwinschmitz: This error is happening on the test-vm on deployment.

PR #642, opened 2020-04-08
fix eslint issue

jannisvisser:


services/121-service/package.json
```diff
@@ -5,7 +5,7 @@
   "homepage": "https://www.121.global/",
   "main": "index.js",
   "scripts": {
-    "lint": "npx eslint '**/*.ts'",
+    "lint": "npx eslint **/*.ts",
```
elwinschmitz, line 8: Not like this?
```suggestion
    "lint": "eslint '**/*.ts'",
```

services/PA-accounts-service/package.json
```diff
@@ -5,7 +5,7 @@
     "homepage": "https://www.121.global/",
     "main": "index.js",
     "scripts": {
-        "lint": "npx eslint '**/*.ts'",
+        "lint": "npx eslint **/*.ts",
```
elwinschmitz, line 8:
```suggestion
        "lint": "eslint '**/*.ts'",
```

services/121-service/package.json
jannisvisser: no that doesn't work for me..

jannisvisser: @elwinschmitz

PR #675, opened 2020-04-15
AB#1732 PlayText-button disappears when switching to other language then english

jannisvisser:


interfaces/PA-App/src/app/shared/play-text-audio/play-text-audio.component.ts
```diff
@@ -112,6 +112,7 @@ export class PlayTextAudioComponent implements OnInit {
     this.state = state;
     this.setLabel(state);
     this.setIcon(state);
+    this.isDisabled = false;
```
elwinschmitz, line 115:
```suggestion
```
I think a better place for it would be in the `createPlayer()`-method.
So, here: https://github.com/global-121/121-platform/pull/675/files#diff-eedc6515a85de1b46dc16ef779464ef6R70-R79
```ts
  private createPlayer() {
    this.isDisabled = false;
    this.player = new Howl({
// ...
    });
```
So you always start 'with a clean slate'.
Resetting the `isDisabled`-property _every_ time its state is set feels a bit too much.

interfaces/PA-App/README.md
```diff
@@ -54,6 +54,16 @@ Audio-file creation/processing:
...
+### NGO logo's
+Use of NGO-logo's is currently hard-coded. When either of the following changes:
+- A new NGO is participating in the 121-system
+- The logo (name) of an NGO changes
+- The NGO name (used in the 121-service program-object) changes
+
+... then this should be reflected in code-changes in:
+- `./src/app/shared/dialogue-turn/dialogue-turn.component.html`
+- and `./src/app/shared/dialogue-turn/dialogue-turn.component.ts`
+
```
elwinschmitz, line 66: 🎉 👍

elwinschmitz:
This PR is now an interesting combination of all kinds of things, but all individual commits are connected to back-log items, so ✅  :)

jannisvisser:
yes, it was a mistake.. I wanted to turn it back, but I'll leave it for now then.

PR #721, opened 2020-04-28
AB#1804 handle error in case of cancel retry

jannisvisser:


interfaces/AW-App/src/app/validation-components/download-data/download-data.component.ts
```diff
@@ -21,8 +21,11 @@ class ValidationAnswer {
 })
 export class DownloadDataComponent implements ValidationComponent {
   public downloadReady = false;
+  public downloadAborted = false;
   public nrDownloaded: number;
 
+  public validationData: ValidationAnswer[];
```
elwinschmitz, line 27: Properties or members of a class only need to be declared `public` if they will be used from 'outside' the class. (So most times only when used in the template)

If that is not the case (as it is here), you should set these properties to `private`.

jannisvisser: Good point. I know this, but my default-mode-without-thinking is still to set 'public'. I will improve!

PR #724, opened 2020-04-29
Use offline

RubenGeo:


interfaces/AW-App/package-lock.json
```diff
@@ -3935,7 +3935,7 @@
         },
         "shelljs": {
           "version": "0.5.3",
-          "resolved": "https://registry.npmjs.org/shelljs/-/shelljs-0.5.3.tgz",
+          "resolved": false,
```
elwinschmitz, line 3938: @RubenGeo How was this change made? Automatically by `npm`? Or something else?

RubenGeo: @elwinschmitz I think by a npm install? Is there another option?

PR #739, opened 2020-05-06
AB#1840 Added test endpoints for notifications  with admin auth

RubenGeo:


services/121-service/src/notifications/sms/sms.controller.ts
```diff
@@ -10,6 +12,22 @@ export class SmsController {
...
+    description: 'Test controller to test sending sms',
+  })
+  @ApiImplicitParam({ name: 'number' })
+  @Get(':number')
+  public async sendSms(@Param() params): Promise<void> {
+    return await this.smsService.notifyBySms(
+      params.number,
+      'en',
+      'included',
+      1,
```
elwinschmitz, line 27: @RubenGeo Are there any 'named' variables we could use here? Like a "InclusionStatus"-enum?

It 'feels' a bit brittle to just use these string-values here, and the 'overhead' of the extra `.enum.ts`-file seems minimal...

PR #760, opened 2020-05-12
AB#1725 feedback

jannisvisser:


jannisvisser:
There are some old but empty commits in here.. when looking at files change, it's only about the latest (feedback).

interfaces/PA-App/src/app/shared/qr-scanner/qr-scanner.component.html
```diff
@@ -47,15 +47,15 @@
...
-                    [title]="'qr-scanner.toggle-flash|translate'"
+                    [title]="toggleFlashTitle"
                     (click)="torchEnabled=!torchEnabled"
                     [color]="torchEnabled ? 'success' : 'light'"
                     class="ion-margin-vertical"
     >
       <ion-icon [name]="torchEnabled ? 'flash' : 'flash-off'"></ion-icon>
     </ion-fab-button>
     <ion-fab-button *ngIf="switchCamerasAvailable"
-                    [title]="'qr-scanner.switch-camera|translate'"
```
elwinschmitz, line 58: Ah, I think the type is with the last quote...

```
[title]="'qr-scanner.switch-camera'|translate"
```

I think that would be a better fix.

@jannisvisser

jannisvisser: go ahead @elwinschmitz ;)

PR #787, opened 2020-05-20
Get funds

RubenGeo:


services/121-service/src/config.ts
```diff
@@ -1,13 +1,18 @@
-
 export const DEBUG = (['production', 'staging'].indexOf(process.env.NODE_ENV) === -1);
 export const PORT = (process.env.PORT) ? process.env.PORT : 3000;
 export const BASE_PATH = (DEBUG) ? 'api' : '121-service/api';
 export const SCHEME = (DEBUG) ? 'http' : 'https';
 
 const tyknIMS = 'http://11.0.0.3:50001/api/';
 const orgIMS = 'http://11.0.0.4:50002/api/';
+const diberse_url = 'https://b2b.demo.disberse.com/project/​'
```
elwinschmitz, line 8: This could be a fine default-value, but make sure you can also set it via a ENV-variable.
So we don't need to change code in a different environment to talk to a different third-party server.

services/.env.example
```diff
@@ -2,3 +2,4 @@ POSTGRES_PASSWORD=<password>
 WALLET_PASSWORD_ENCRYPTION_KEY=<key>
 
 EXTERNAL_121_SERVICE_URL=
+DISBERSE_API_KEY=<key>
```
elwinschmitz, line 5: Don't forget to mention this addition in the CHANGELOG! ;)

services/121-service/src/funding/disberse-api.service.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  public async get(path: string): Promise<any> {
+    const endcodedURI = encodeURI(API.disberse.url +  path)
+    console.log(API.disberse.key)
+    return Promise.resolve(12345); // This needs to be removed after we get the API key
+    return this.httpService
+      .get(endcodedURI
+      ).pipe(
+        tap(response => console.log(`ApiService GET: ${API.disberse.url}${path}`, `\nResponse:`, response)),
```
elwinschmitz, line 20: This is not the "ApiService", but the DisberseApiService.
```suggestion
        tap(response => console.log(`DisberseApiService GET: ${encodedURI}${path}`, `\nResponse:`, response)),
```

jannisvisser:
@RubenGeo this PR can be closed I think right?

PR #792, opened 2020-05-20
AB#1918 Streamline scan/check/retrieve flow

elwinschmitz:


elwinschmitz:
@jannisvisser Because I used the test-vm as my 'local' back-end, I couldn't really test everything. And I didn't want to trigger a deploy/build during the demo ;)

elwinschmitz:
@jannisvisser I've now also changed the way the "validate-program"-component works.
At first sight I thought they should not be "interconnected", so my change would only need changes in the first part.

The whole connection between the two components should be based on only 1 "contract" (i.e. the "PaData"-object or only the "PaIdentifier/did", not both.

But the whole "sessionStorageService" thats sitting inbetween them both also complicates things. But my guess is that we can refactor that part out when we implement the "browser-based-QR-scanner"... so To be Continued... ;)

PR #812, opened 2020-05-27
Store did in PA accounts user table AB#1962

RubenGeo:


services/PA-accounts-service/src/user/user.service.ts
```diff
@@ -73,6 +74,13 @@ export class UserService {
     }
   }
 
+  public async setDid(id: number, didDto: DidDto): Promise<UserRO> {
+    let user = await this.userRepository.findOne(id);
+    user.did = didDto.did;
+    const updatedUser = await this.userRepository.save(user);
+    return this.buildUserRO(updatedUser);
```
elwinschmitz, line 81: Does it need to return the full User-object?
It is only a "POST"-endpoint right, we only need confirmation that it succeeded, nothing more.

interfaces/PA-App/src/app/personal-components/create-identity/create-identity.component.ts
```diff
@@ -169,6 +169,9 @@ export class CreateIdentityComponent extends PersonalComponent {
     this.paData.store(this.paData.type.wallet, wallet);
     this.paData.store(this.paData.type.didShort, didShort);
     this.paData.store(this.paData.type.did, did);
+
+    // 9. Store did in user table of PA-account
```
elwinschmitz, line 173: Can we also replace the line 171:
```
this.paData.store(this.paData.type.did, did);
```

We would need a `this.paData.getDid()`-method to replace all the existing `paData.retreive('did')` calls.
But it keeps everything more 'together'...

Also, it doesn't need to be the responsibility of the `create-identity`-component to do this. It could be managed by the `PaDataService` as that is already aware of the "offlineStorage"-state etc.

RubenGeo: We discussed this and decided not to change this

PR #887, opened 2020-06-16
AB#1137 AW-app stay signed in upon refresh

jannisvisser:


interfaces/AW-App/src/app/services/programs-service-api.service.ts
```diff
@@ -13,29 +14,43 @@ import { JwtService } from './jwt.service';
 export class ProgramsServiceApiService {
   constructor(private apiService: ApiService, private jwtService: JwtService) {}
 
-  login(email: string, password: string): Promise<any> {
+  // login(email: string, password: string): Promise<any> {
```
elwinschmitz, line 17: If it isn't necessary, it can be removed, right?

PR #890, opened 2020-06-17
AB#1872 refactor Person Affected Status

jannisvisser:


interfaces/HO-Portal/src/app/models/person.model.ts
```diff
@@ -14,17 +14,29 @@ export class Person {
...
   inclusionDate?: string;
+  status: PaStatus;
 }
 
 // Model for display (in table)
 export class PersonRow {
   did: string;
   checkboxVisible: boolean;
   pa: string; // Display label
+  status: PaStatus;
```
elwinschmitz, line 25: This one should not be of type `PaStatus`, as it would containt the translated string-version..

PR #911, opened 2020-06-23
AB#1554 HO-Portal header/menu refactoring

elwinschmitz:


interfaces/HO-Portal/src/app/pages/inclusion/inclusion.page.html
```diff
@@ -12,6 +12,11 @@
     [phaseReady]="isReady"
   ></app-phase-next>
 
+  <app-export-inclusion
```
jannisvisser, line 15: @elwinschmitz this component should be in review-inclusion page, not in inclusion. (Apart from this I was expecting to pick this up within that other PBI, and not within this one, but fine as well obviously.)

elwinschmitz: Yeah, I needed to resolve the conflict, so had to touch is anyway...

PR #984, opened 2020-07-15
Resize scan in debug mode AB#23119

RubenGeo:


interfaces/AW-App/src/app/validation-components/scan-qr/scan-qr.component.ts
```diff
@@ -57,9 +57,15 @@ export class ScanQrComponent implements ValidationComponent {
   "programId": 1 }`,
           }
         : {};
+    const cssClass =
+      environment.isDebug || environment.showDebug
+        ? ''
```
elwinschmitz, line 62: `cssClass` will always be an empty string now. I think you want to add `'debugModal'` to the first `''`.

interfaces/AW-App/src/global.scss
```diff
@@ -9,3 +9,13 @@
 @import '~@ionic/angular/css/text-alignment.css';
 @import '~@ionic/angular/css/text-transformation.css';
 @import '~@ionic/angular/css/flex-utils.css';
+
+
+.debugModal
+{
```
elwinschmitz, line 15: This syntax isn't "Pretty" enough. ;)

interfaces/AW-App/src/global.scss
```diff
@@ -9,3 +9,13 @@
...
+    .modal-wrapper
+    {
+        height: 60%;
```
elwinschmitz, line 18: Does this work on all kind of window-sizes/devices?

PR #1002, opened 2020-07-16
AB#2341 [PoC] Refactor PA-data handling

elwinschmitz:
Conclusion... not a real improvement. :(

### DIFF
--- a/interfaces/PA-App/src/app/services/pa-account-api.service.ts
+++ b/interfaces/PA-App/src/app/services/pa-account-api.service.ts
@@ -66,7 +66,13 @@ export class PaAccountApiService {
       )
       .pipe(
         map((value) => {
-          return value.data ? JSON.parse(value.data) : value;
+          let data;
+          try {
+            data = JSON.parse(value);
+          } catch {
+            data = value;
+          }
+          return data;
         }),
         catchError((error) => {
           if (error.error instanceof ErrorEvent) {
--- a/interfaces/PA-App/src/app/services/pa-account-api.service.ts
+++ b/interfaces/PA-App/src/app/services/pa-account-api.service.ts
@@ -55,7 +55,7 @@ export class PaAccountApiService {
       .toPromise();
   }
 
-  retrieve(type: string): Promise<any> {
+  retrieve(type: string): Promise<undefined | string | number | object> {
     console.log('PaAccountApiService : retrieve()');
 
     return this.apiService
@@ -65,6 +65,9 @@ export class PaAccountApiService {
         false,
       )
       .pipe(
+        map((value) => {
+          return value.data ? JSON.parse(value.data) : value;
+        }),
         catchError((error) => {
           if (error.error instanceof ErrorEvent) {
             // Client-side error:
--- a/interfaces/PA-App/src/app/services/padata.service.ts
+++ b/interfaces/PA-App/src/app/services/padata.service.ts
@@ -112,9 +112,7 @@ export class PaDataService {
       return this.ionStorage.get(type);
     }
 
-    const paAccountApiOutput = await this.paAccountApi.retrieve(type);
-
-    return paAccountApiOutput;
+    return await this.paAccountApi.retrieve(type);
   }
 
   /////////////////////////////////////////////////////////////////////////////
--- a/interfaces/PA-App/src/app/services/padata.service.ts
+++ b/interfaces/PA-App/src/app/services/padata.service.ts
@@ -114,11 +114,7 @@ export class PaDataService {
 
     const paAccountApiOutput = await this.paAccountApi.retrieve(type);
 
-    // Only JSON.parse the output if it could result in something:
-    if (!paAccountApiOutput) {
-      return undefined;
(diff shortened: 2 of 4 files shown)


interfaces/PA-App/src/app/services/pa-account-api.service.ts
```diff
@@ -55,7 +55,7 @@ export class PaAccountApiService {
       .toPromise();
   }
 
-  retrieve(type: string): Promise<any> {
+  retrieve(type: string): Promise<undefined | string | number | object> {
```
elwinschmitz, line 58: This never hurts... to be specific.
But also a bit 'redundant'...

interfaces/PA-App/src/app/services/pa-account-api.service.ts
```diff
@@ -65,6 +65,9 @@ export class PaAccountApiService {
         false,
       )
       .pipe(
+        map((value) => {
+          return value.data ? JSON.parse(value.data) : value;
```
elwinschmitz, line 69: The logic of parsing the returned string is moved into the API-handler.
As the API returns the `{ "data": "..." }`-style object, this handler 'should know' what to do with it...

interfaces/PA-App/src/app/services/padata.service.ts
```diff
@@ -114,11 +114,7 @@ export class PaDataService {
 
     const paAccountApiOutput = await this.paAccountApi.retrieve(type);
 
-    // Only JSON.parse the output if it could result in something:
-    if (!paAccountApiOutput) {
-      return undefined;
-    }
-    return JSON.parse(paAccountApiOutput);
+    return paAccountApiOutput;
```
elwinschmitz, line 117: We 'trust' the API-handler to do everything for us and only return sane results...

services/PA-accounts-service/src/data-storage/data-storage.controller.ts
```diff
@@ -44,7 +44,9 @@ export class DataStorageController {
   public async get(
     @User('id') userId: number,
     @Param() params,
-  ): Promise<string> {
-    return await this.dataStorageService.get(userId, params.type);
+  ): Promise<{ data: string | number | object }> {
+    const data = await this.dataStorageService.get(userId, params.type);
+
+    return { data };
```
elwinschmitz, line 50: This ensures us that the endpoint always returns JSON.

But the contents of the `data`-property is still a string. (A stringified-JS-object/string).

services/PA-accounts-service/src/data-storage/data-storage.service.ts
```diff
@@ -35,11 +35,11 @@ export class DataStorageService {
...
     if (!data || data.length === 0) {
-      const errors = { Data: ' not found' };
-      throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
+      throw new HttpException('', HttpStatus.NOT_FOUND);
     }
 
     data[0].data = this.cryptr.decrypt(data[0].data);
-    return JSON.stringify(data[0].data);
+
+    return data[0].data;
```
elwinschmitz, line 43: This extra 'stringification' isn't really necessary to do explicitly here..

But I think that NestJS will do it for us anyway when we do:
`return { data };`

So no _real_ win in the end.

services/PA-accounts-service/src/data-storage/data-storage.service.ts
```diff
@@ -35,11 +35,11 @@ export class DataStorageService {
       order: { created: 'DESC' },
     });
     if (!data || data.length === 0) {
-      const errors = { Data: ' not found' };
-      throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
+      throw new HttpException('', HttpStatus.NOT_FOUND);
     }
 
     data[0].data = this.cryptr.decrypt(data[0].data);
```
elwinschmitz, line 41: This will always be a string... A normal one or a valid JSON-string.

elwinschmitz:
Conclusion: I don't think we can improve this. ;)

jannisvisser:
@elwinschmitz I trust your conclusion. But isn't it beneficial to just merge this set of changes, even if it doesn't improve much? Or is this not a "complete PR" in the sense that you're not sure what the effects are? (And I guess the test needs to be fixed, which might already be more cost than the added value of these changes?)

elwinschmitz:
"work to get the tests working" is one extra 'cost' that I didn't want to spend yet. (But it shouldn't be impossible + it will add some value in testing this "more specifically typed" version).

But the biggest downside of this change is that it will invalidate/brake any existing data in all environments (production, test, locals, etc). Local dev-environment + test-vm is quickly 'fixable', but production-env will be some extra 'cost' later, when we update the environment.

So that was a bit too much hassle/uncertainty for my taste now.
But I think a compromise can be to 'water down' this PR and only change the types etc, but not change the actual data being stored/sent.

jannisvisser:
@elwinschmitz clear. then go ahead with the 'watered down' PR I would say.

PR #1008, opened 2020-07-21
AB#2649 adapt code to new NODE_ENV values

jannisvisser:


services/PA-accounts-service/src/config.ts
```diff
@@ -13,7 +13,7 @@ export const URL_121_SERVICE = process.env.URL_121_SERVICE_API;
 export const URL_USERIMS = 'http://11.0.0.5:50003/api';
 
 let _walletPasswordEncryptionKey: string;
-if (process.env.NODE_ENV == 'production' || process.env.NODE_ENV == 'staging') {
+if (['production', 'staging', 'test'].indexOf(process.env.NODE_ENV) > -1) {
```
elwinschmitz, line 16: I think we should remove this decision/dependency-on-NODE_ENV entirely.
We could base it on the value/existence of the `WALLET_PASSWORD_ENCRYPTION_KEY`-variable.
(Or add another like `ENABLE_ENCRYPTION`, but that doesn't make a lot of sense).

services/PA-accounts-service/src/config.ts
```diff
@@ -4,7 +4,7 @@ export const AUTH_DEBUG = false;
 export const PORT = process.env.PORT_PA_ACCOUNTS_SERVICE;
 export const BASE_PATH = process.env.SUBDOMAIN_PA_ACCOUNTS_SERVICE + '/api';
 export const SCHEME =
```
elwinschmitz, line 6: We should make the SCHEME itself configurable via an ENV-variable, so we don't need to have logic in our code for it.

PR #1045, opened 2020-07-29
AB#2684 After validation of PA, not all data is removed from /download-data enpoint

jannisvisser:


interfaces/AW-App/src/app/validation-components/upload-data/upload-data.component.ts
```diff
@@ -93,18 +108,13 @@ export class UploadDataComponent implements ValidationComponent {
...
-      this.ionicStorageTypes.credentials,
-      currentCredentials,
-    );
+  public async removeLocalStorageData(
+    did: string,
+    ionicStorageType: IonicStorageTypes,
+  ): Promise<void> {
+    let data = await this.storage.get(this.ionicStorageTypes[ionicStorageType]);
+    data = data.filter((item) => item.did !== did);
+    await this.storage.set(this.ionicStorageTypes[ionicStorageType], data);
```
elwinschmitz, line 117: The value of the parameter `ionicStorageType` is already the correct value, I think.

```suggestion
  public async removeLocalStorageData(
    did: string,
    type: IonicStorageTypes,
  ): Promise<void> {
    let data = await this.storage.get(type);
    data = data.filter((item) => item.did !== did);
    await this.storage.set(type, data);
```

Should also work. (Haven't tested it though..)

interfaces/AW-App/src/app/validation-components/upload-data/upload-data.component.ts
```diff
@@ -45,7 +45,22 @@ export class UploadDataComponent implements ValidationComponent {
...
+          this.ionicStorageTypes.validationProgramData,
+        );
+        await this.removeLocalStorageData(
+          credential.did,
+          this.ionicStorageTypes.validationFspData,
+        );
+        await this.removeLocalStorageData(
+          credential.did,
+          this.ionicStorageTypes.qrDidMapping,
+        );
```
elwinschmitz, line 63: You don't have to use the local instance of these variables, as they are just string-constants,

```suggestion
        await this.removeLocalStorageData(
          credential.did,
          IonicStorageTypes.credentials,
        );
        await this.removeLocalStorageData(
          credential.did,
          IonicStorageTypes.validationProgramData,
        );
        await this.removeLocalStorageData(
          credential.did,
          IonicStorageTypes.validationFspData,
        );
        await this.removeLocalStorageData(
          credential.did,
          IonicStorageTypes.qrDidMapping,
        );
```

PR #1069, opened 2020-08-04
Ken staging

jannisvisser:


services/121-service/src/scripts/scripts.controller.ts
```diff
@@ -5,13 +5,14 @@ import { Connection } from 'typeorm';
 import { SeedSingleProgram } from './seed-program-single';
 import { SeedDemoProgram } from './seed-program-demo';
 import { SeedPilotNLProgram } from './seed-program-pilot-nl';
+import SeedPilotKenProgram from './seed-program-pilot-ken';
```
elwinschmitz, line 8: Not like this?
```suggestion
import { SeedPilotKenProgram } from './seed-program-pilot-ken';
```

jannisvisser: @elwinschmitz it automatically added it like this, and it works.. But I agree that it looks ugly ;)

PR #1102, opened 2020-08-18
Task 2926

omendram:
Adding sample tests for Unit Tests.

In order to specifically run the tests, within the framework of Jasmin and Angular, we can only target by using the focus(?) prefix. So while we make changes we should prefix the describe with fdescribe (As shown within this change of account-page test spec).

Additionally, we can write tests for following scenarios which include:

- declarations of class wide variables
- calling of methods within the class / non-private objects
- checking the if and else statements


omendram:
Additionally, speaking I have added phantom-js also as a headless browser for executing the unit test cases. It is helpful to be able to see the results of the test cases on the command line as well, and not always having to launch new web browsers every time a test cases needs to be executed.

PR #1118, opened 2020-08-25
AB#3011 README update for unit test cases

omendram:
Hey Elwin, 

README within interfaces contains instructions and details regarding unit test cases, and why we should write them and how.

PR #1119, opened 2020-08-25
AB#2926  Unit tests sample file

omendram:
A sample of test cases @elwinschmitz  and @jannisvisser . 

Just wanted to add reviewers as to check for the potential errors and provide your thoughts in case you have any feedback towards those test cases and their applicability.

The detailed documentation can be found withing another PR: https://github.com/global-121/121-platform/pull/1118


elwinschmitz:
I think these examples are illustrative, but not really 'functional' as real tests.
So I would suggest to separate the inclusion of "PhantomJS" + documentation about its use into a separate PR and close this one to only use this for its 'illustrational purpose'. ;)

interfaces/AW-App/src/app/account/account.page.spec.ts
```diff
@@ -30,4 +42,46 @@ describe('AccountPage', () => {
...
   });
+
+  it('ngOnInit: should set up variables', () => {
+    expect(component.isLoggedIn).toBeDefined();
+  });
+
+  it('doLogin: should call login of authService', () => {
+    spyOn(event, "preventDefault");
+    component.doLogin(event);
+    expect(event.preventDefault).toHaveBeenCalled();
```
elwinschmitz, line 53: Is the `event`-object chosen as 'just an example of a parameter', or is it the suggestion to use this default browser-behaviour as a feature in our tests?

Is it to check if "is the first line of code in the method executed"? Or something else/more?

As the

PR #1174, opened 2020-08-27
AB#3001 Update program-questions for Kenya pilot

elwinschmitz:


services/121-service/examples/program-pilot-ken.json
```diff
@@ -47,7 +47,7 @@
     {
       "criterium": "name",
       "label": {
-        "en": "Confirm full name:"
+        "en": "My full name:"
```
jannisvisser, line 50: @elwinschmitz this change is fine with me. But please be aware that Orla and me changed it the other way around right before the user-test. The reason being that we changed 'username' to 'first and second name'. Users would perceive this name-question as a duplicate, and therefore we arrived at 'confirm'. Not sure what the real solution around username+name will be (there's a spike for that), but maybe best to not change this back in the meantime (so leave 'confirm').

PR #1190, opened 2020-09-01
AB#3101 Create tests for payment details

RubenGeo:
@omendram 

1.  Do you have any comments on these tests?
2. Do you have any insights with regards to the amount of effort we can put into these vs the value they add to the 121 project? Does it makes sense to work backwards and create these sort of tests for all our functions?

@jannisvisser So you can also see how the more extended tests look

### DIFF
--- a/services/121-service/src/programs/fsp/fsp.service.ts
+++ b/services/121-service/src/programs/fsp/fsp.service.ts
@@ -64,14 +64,20 @@ export class FspService {
 
     if (paymentResult.status === StatusEnum.succes) {
       for (let connection of details.connectionsForFsp) {
-        await this.storeTransaction(amount, connection, fsp, program, installment);
+        await this.storeTransaction(
+          amount,
+          connection,
+          fsp,
+          program,
+          installment,
+        );
       }
     }
 
     return { paymentResult, nrConnectionsFsp: details.paymentList.length };
   }
 
-  public async createPaymentDetails(
+  private async createPaymentDetails(
     includedConnections: ConnectionEntity[],
     amount: number,
     fspId: number,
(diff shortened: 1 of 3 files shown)


omendram:
> @omendram
> 
> 1. Do you have any comments on these tests?
> 2. Do you have any insights with regards to the amount of effort we can put into these vs the value they add to the 121 project? Does it makes sense to work backwards and create these sort of tests for all our functions?
> 
> @jannisvisser So you can also see how the more extended tests look

Hey @RubenGeo 

1. I think the tests look good. 
2. In regards to how much effort we should put in vs what they add to 121-project, unit test cases just help ensure the functionality of an individual unit to work independently once we take care of external classes/method dependencies. It adds value to the quality of our development project in terms of DevOps traditions and software development standards. Unit testing doesn't slow us down, once we get used to it, it is a fast implementation, and after that it should not take more than an hour or so. Also, when we make major changes to the code, it helps us make sure the individual functionalities are working as we have already written the tests for different related units. It just provides an additional layer of security of how things are working and what individual units are supposed to do. When major changes are ahead, a unit test based test driven approach can also help pre-define detailed goals and work efficiently. I understand the transition and integrating the habit of detailed unit testing can take a bit of time, but it helps in the long run.

services/121-service/src/programs/fsp/fsp.service.ts
```diff
@@ -64,14 +64,20 @@ export class FspService {
...
+          installment,
+        );
       }
     }
 
     return { paymentResult, nrConnectionsFsp: details.paymentList.length };
   }
 
-  public async createPaymentDetails(
+  private async createPaymentDetails(
```
omendram, line 80: is this expected? if so, then all good.

RubenGeo: Yes, i was showing/explaining Jannis the tests and we noticed that that method should be private.

PR #1236, opened 2020-09-10
AB#3223 Cypress tests for HO Portal

omendram:


interfaces/HO-Portal/cypress/fixtures/user.json
```diff
@@ -0,0 +1,9 @@
+{
```
omendram, line 1: Stub data for API calls, for Login!

package-lock.json
```diff
@@ -0,0 +1,4 @@
+{
```
elwinschmitz, line 1: I don't think we need this file to be committed?
We could ignore it via `.gitignore` maybe?

interfaces/HO-Portal/cypress.json
```diff
@@ -0,0 +1,3 @@
+{
+    "baseUrl": "http://localhost:4200"
```
elwinschmitz, line 2: I think you're local set-up hasn't used/configured to use Prettier on these files, as the whitespace differs here and there...

If you open the folder `interfaces/HO-Portal` as a 'workspace' in VSCode (with the Prettier-plugin installed), it will automatically pick up the configuration to "format on save" for all supported files.

Or manually running `npm run fix:syntax` before committing.

interfaces/HO-Portal/cypress/fixtures/login.json
```diff
@@ -0,0 +1,5 @@
+{
+  "portal": "http://localhost:4200/login",
```
elwinschmitz, line 2: Where is `portal` used?
And could it be a relative URL? (So only `/login`)

interfaces/HO-Portal/cypress/integration/src/login.spec.js
```diff
@@ -0,0 +1,46 @@
+describe('Login Page', () => {
+    beforeEach(() => {
+        cy.server();
+    });
+
+    // Real API call
+    it('default user can log in', function () {
+        cy.visit('/login');
+        cy.get('input[name="email"]').type('test@example.org');
```
elwinschmitz, line 9: This is not using the fixture-file?

elwinschmitz (changes requested):
Nice straight-forward example.

It does require some initial set-up/installing that isn't included in this PR yet, right?
I think the current set-up of all 4 separate interfaces would require to install `cypress` as a dev-dependency in each individual interface with the same npm-scripts for all.

elwinschmitz (approved):

elwinschmitz:
@omendram Apparently I've changed this PR from Draft to a real one... :/ 😨 

Apart from the installation-instructions to set-up Cypress on the local dev-environment, this looks good to me. 👍

elwinschmitz (approved):

PR #1252, opened 2020-09-15
AB#3181 moving configs

omendram:


services/.env.example
```diff
@@ -28,6 +28,26 @@ URL_PA_ACCOUNTS_SERVICE_API=http://11.0.0.8:3001/api
 EXTERNAL_121_SERVICE_URL=
 
 
+# SECRETS
+121_SERVICE_SECRETS_SECRET=secret-key
```
elwinschmitz, line 32: For previous ENV-vars we've come up with some tricks to not-start-with-a-digit, but maybe we need to go with something more generic... like `A121_ENV_VARIABLE*` (and all in uppercase)

See: https://stackoverflow.com/a/2821183

elwinschmitz (changes requested):

services/.env.example
omendram: Good spot. Thanks

PR #1324, opened 2020-10-05
AB#3014 Remove "country"-related elements from PA-app

elwinschmitz:
Related [AB#3690](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/3690)


interfaces/HO-Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -220,15 +220,14 @@ export class ProgramsServiceApiService {
...
-    countryId: number,
   ): Promise<any> {
     return this.apiService
       .post(environment.url_121_service_api, `/user`, {
         email,
         password,
         role,
         status,
-        countryId,
+        countryId: 1,
```
jannisvisser, line 230: @elwinschmitz did you leave this in here, because this can only be removed once the back-end is changed as well?

PR #1327, opened 2020-10-06
AB#3138 Log errors for interfaces

omendram:


interfaces/PA-App/src/app/services/logging.service.ts
```diff
@@ -0,0 +1,99 @@
...
+      return false;
+    }
+
+    // filter/sample data
+
+    return false;
+  }
+
+  logPageView(name?: string, url?: string) {
+    if (environment.ai_ikey && environment.ai_endpoint) {
```
elwinschmitz, line 47: I think this repeated condition could be put in a separate method...
Or maybe even stop at the "constructor"-level, as nothing will work without these values...

elwinschmitz (changes requested):

elwinschmitz (approved):

PR #1339, opened 2020-10-07
AB#3741 replace play with ear icon

jannisvisser:
@elwinschmitz please review the way I set the size of the ear-icon-button to align with the rest.


elwinschmitz:
I think the size could/should be set via a `font-size`-property, but this also works.

PR #1341, opened 2020-10-07
AB#3729

omendram:
@elwinschmitz  We will also need to add ENV related changes too.


elwinschmitz:
@omendram These changes: https://github.com/global-121/121-platform/pull/1312/files need to be applied for the AW-app as well then.

elwinschmitz (approved):

PR #1343, opened 2020-10-07
AB#3728

omendram:
need to assign env variables for the interface.


elwinschmitz:
@omendram These changes: https://github.com/global-121/121-platform/pull/1312/files need to be applied for the HO-portal as well then.

elwinschmitz (approved):

PR #1345, opened 2020-10-07
AB#3747

omendram:
Need to add env variables for Referral App


elwinschmitz:
@omendram These changes: https://github.com/global-121/121-platform/pull/1312/files need to be applied for the Referral-app as well then.

omendram:
> @omendram These changes: https://github.com/global-121/121-platform/pull/1312/files need to be applied for the Referral-app as well then.

I have added changed to env.example and environment.ts file, however we'd still need to add values to .env file.

elwinschmitz:
@omendram Is it the same value for all interfaces? Or has each interface its own iKey?

omendram:
> @omendram Is it the same value for all interfaces? Or has each interface its own iKey?

It will be the same value for all, however it has changed from the old one. Can be accessed in azure portal under overview of 121-application-insights resource.

elwinschmitz (approved):

PR #1361, opened 2020-10-12
Refs AB#3139 AB#3749 Adding app insights

omendram:


services/121-service/src/main.ts
```diff
@@ -3,6 +3,7 @@ import { ApplicationModule } from './app.module';
 import { SwaggerModule, DocumentBuilder } from '@nestjs/swagger';
 import { ValidationPipe } from '@nestjs/common';
 import { PORT, BASE_PATH, SCHEME } from './config';
+const appInsights = require("applicationinsights");
```
elwinschmitz, line 6: Hmm.. why doesn't Prettier complain about these `"` instead of `'` here? Weird.

omendram: yeah, have been fixed anyway.

PR #1386, opened 2020-10-14
AB#3837 feat: added analytics events

gulfaraz:


interfaces/Referral-App/src/app/components/offer/offer.component.html
```diff
@@ -26,7 +26,10 @@
     <ion-item *ngIf="offer.offerLink">
       <ion-icon name="globe" slot="start"></ion-icon>
       <ion-label>
-        Website: <a href="{{ offer.offerLink }}">{{ offer.offerLink }}</a>
+        Website:
+        <a href="{{ offer.offerLink }}" (click)="logClick('website')">{{
```
elwinschmitz, line 30: Maybe good to add the attributes:
```
target="_blank" rel="noopener"
```
to these external links. It's not the best-best-practice perse, but I think that it preserves the 'app-like' feel of the Referral-app a bit better.

interfaces/Referral-App/src/app/components/offer/offer.component.ts
```diff
@@ -16,5 +17,12 @@ export class OfferComponent {
   @Input()
   goBack;
 
-  constructor() {}
+  constructor(private loggingService: LoggingService) {}
+
+  public logClick(name) {
+    this.loggingService.logEvent('referral-offer-click', {
```
elwinschmitz, line 23: I think we should put these event-names into a centrally stored Enum, so that we can see how many different ones there are and that we prevent typo's and get nice auto-complete in our editors.

interfaces/Referral-App/src/app/help/help.page.html
```diff
@@ -17,25 +17,39 @@
...
+            >{{ help.helpPhone }}</a
+          >
         </p>
       </ion-col>
     </ion-row>
     <ion-row>
       <ion-col>
-        <a href="{{ help.helpWhatsApp }}" *ngIf="help.helpWhatsApp">
+        <a
+          href="{{ help.helpWhatsApp }}"
```
elwinschmitz, line 29: For WhatsApp its possible to create a smoothe 'app link' by using the URL https://wa.me/31600000000
If PA have the app installed, it will automatically open, others will land on a (sort of) helpful page.

(And for these external links I would also suggest to add the `target` and `rel`-attributes.)

omendram (approved):

interfaces/Referral-App/src/app/help/help.page.html
gulfaraz: the href value can be set via the google sheet, new tab attributes are added

elwinschmitz (approved):

PR #1388, opened 2020-10-14
AB#3874 AB#3875 Double phone

RubenGeo:


services/121-service/seed-data/program/program-pilot-nl.json
```diff
@@ -61,8 +61,20 @@
       "label": {
         "en": "My V number: (It needs to be 10 digits long)"
       },
+      "answerType": "number",
```
elwinschmitz, line 64: @RubenGeo There is no such `answerType`, the correct value is `numeric`. (it's already fixed in a later PR, but just to let you know.

PR #1414, opened 2020-10-20
AB#4029 only return published programs in PA

jannisvisser:


services/121-service/src/programs/program/program.service.ts
```diff
@@ -769,6 +769,7 @@ export class ProgramService {
 
     const connectionsResponse = [];
     for (let connection of selectedConnections) {
+      console.log;
```
elwinschmitz, line 772: @jannisvisser Whooops. ;)

services/121-service/src/programs/fsp/financial-service-provider.entity.ts
```diff
@@ -54,6 +57,7 @@ export class FinancialServiceProviderEntity {
 }
 
 export enum fspName {
-  intersolve = 'Intersolve',
+  intersolve = 'Intersolve-whatsapp',
+  intersolveNoWhatsapp = 'Intersolve-no-whatsapp',
   mpesa = 'M-Pesa',
 }
```
elwinschmitz, line 63: @jannisvisser  Where are these strings used? Do they need to match 1-on-1 with the `fsp`-property in these json-files?

Or are the the "human-readable"-versions to show in the HO-portal?

jannisvisser: @elwinschmitz both
- as discussed with you yesterday, I use the old fsp-name also as the somewhat human-readable fsp-name for HO-portal.
- but your 1st bullet is a valid remark/concern: yes, this needs to match with the name as given in the seed-script. At the time of coding this I didn't see how to better solve this.

PR #1474, opened 2020-10-28
AB#4108 return 1 voucher based on did+installment

jannisvisser:


services/121-service/src/programs/fsp/fsp.controller.ts
```diff
@@ -76,17 +76,18 @@ export class FspController {
...
+    const blob = await this.intersolveService.exportVouchers(
+      exportVoucherDto.did,
+      exportVoucherDto.installment,
+    );
+    var bufferStream = new stream.PassThrough();
+    bufferStream.end(Buffer.from(blob, 'binary'));
+    response.writeHead(200, {
+      'Content-Type': 'image/png',
+    });
+    bufferStream.pipe(response);
```
elwinschmitz, line 91: I have the feeling that these lines can be simplified using some of the Nest.js/Express build-in methods..

(like `response.contentType('png)`)

But I think we can refactor this in multiple places at once when we get to it...

PR #1496, opened 2020-11-03
AB#4272 bulk-action notify-included

jannisvisser:


jannisvisser:
@elwinschmitz as for the 2nd (remove) commit: I chose to be really minimal, and leave the notification component in, because it might be re-used, but at the very least, code within it will be re-used (adding timestamp stuff to e.g. export-buttons is quite high on the backlog).

PR #1522, opened 2020-11-11
AB#4378 don't show vnumber in kenya

jannisvisser:
@elwinschmitz see this unmerged PR which solves this
- I realize it's super-ugly (filtering on whether the 1st PA has a vnumber)
- but will work in practice (in kenya no-one has a vnumber, in nld it's a mandatory field)
- I tried cleaner solutions, by putting loadData before loadColumns, and then being able to loadColumns based on data, but this didn't seem possible.
- another downside is that we have less control of in what position the columns comes exactly
- anyway: if you think it's not full-proof and/or too ugly, feel free to not merge (and either improve or decide it's not worth it)

### DIFF
--- a/interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.html
+++ b/interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.html
@@ -102,6 +102,17 @@
   ></ngx-datatable-column>
 
   <ngx-datatable-column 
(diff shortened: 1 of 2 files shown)


elwinschmitz:
@jannisvisser We will change this around anyway when we're going to work on the "dynamic data based on program-questions in HO-portal"-item, so for now a quick&dirty fix seems good enough. 👍

PR #1542, opened 2020-11-18
AB#4291 ReadMe Edits

omendram:


README.md
```diff
@@ -105,22 +109,26 @@ Download/Clone the Git repository.
...
 
 Or to install 1 specific interface's dependencies, run: (where `<interface-name>` is one of `pa`, `aw`, `ho`, `referral`)
 
     npm run install:<interface-name>
 
 ## Setup Services
+Install dependencies for the interfaces, from the individual interface directories (`services/*`) run:
+
+    npm install
+
```
elwinschmitz, line 124: This is only required if you want to work on a service in a 'standalone', local way.
It is not the most convenient way and needs some manual adjustments.

Services can/should run within the Docker environment by default. Any other way should only be done "when you know what you're doing".

```suggestion
```

README.md
```diff
@@ -105,22 +109,26 @@ Download/Clone the Git repository.
 
 
 ## Setup Interfaces
-Install dependencies for the interfaces, from the root of this repository, run:
+Install dependencies for the interfaces, from the individual interface directories (`interfaces/*`) run:
 
-    npm run install:interfaces
+    npm install
```
elwinschmitz, line 114:
```suggestion
    npm run install:interfaces
```

README.md
```diff
@@ -105,22 +109,26 @@ Download/Clone the Git repository.
 
 
 ## Setup Interfaces
-Install dependencies for the interfaces, from the root of this repository, run:
+Install dependencies for the interfaces, from the individual interface directories (`interfaces/*`) run:
```
elwinschmitz, line 112:
```suggestion
Install all dependencies for all interfaces. From the root of this repository, run:
```

README.md
```diff
@@ -105,22 +109,26 @@ Download/Clone the Git repository.
 
 
 ## Setup Interfaces
-Install dependencies for the interfaces, from the root of this repository, run:
```
elwinschmitz, line 108: This feels like a step back.
I think it helps to "stay out of the specific parts and start-from-the-root" especially in this initial Getting-Started README.

README.md
```diff
@@ -105,22 +109,26 @@ Download/Clone the Git repository.
...
+
+    npm install
+
 
 Follow the "[Getting started / installation](services/README.md#getting-started--installation)"-section in the [services/README](services/README.md)-file.
 
-After that, you can start in development mode by running from the root folder:
-- `npm run start:services`
+## Start Interfaces
+To start all interfaces at once, execute `npm run start:interfaces` from the root of the project folder.
```
elwinschmitz, line 129:
```suggestion
After that, you can start in development mode by running from the root of this repository:
- `npm run start:services`

## Start Interfaces
To start all interfaces at once, execute `npm run start:interfaces` from the root of this repository.
```

README.md
```diff
@@ -131,6 +139,116 @@ To start either of the interfaces in development mode:
   - Unit-tests and UI-tests for all interfaces; Run with `npm test` in each `interfaces/*`-folder.
   - Unit-tests and integration-tests for all services; Run with `npm test` in each `services/*`-folder.
 
+## Common problems with Local Environment set-up
+If the swagger UI is not accessible after installing docker and setting up services, we can take following steps to debug:
+1. `docker ps` to list running containers and their statuses
+2. `docker logs -f <container-name>` to check their statuses
```
elwinschmitz, line 145:
```suggestion
1. `docker-compose ps` to list running containers and their statuses
2. `docker-compose logs -f <container-name>` to check their statuses
```

We should use `docker-compose` as that is aware of all the other services in our `docker-compose.yml`-file, and not on the individual containers.

README.md
```diff
@@ -131,6 +139,116 @@ To start either of the interfaces in development mode:
...
+
+- If the errors are related to not being able to access/connect to 'database' then resetting the database by using `dropSchema: true` in `ormconfig.ts`.
+
+- In order to access docker container itself, execute command `sudo docker exec -it <container-name> sh` and verify installation of npm modules.
```
elwinschmitz, line 149:
```suggestion
- In order to access an individual docker container, execute command `docker-compose exec -it <container-name> sh` and verify installation of npm modules.
```

`sudo` shouldn't be necessary here. (Currently on our servers, it _is_; But that should be fixed there)

README.md
```diff
@@ -131,6 +139,116 @@ To start either of the interfaces in development mode:
...
+
+
+## Unit Tests
```
elwinschmitz, line 152: There is a section-header "Testing" at line 136: https://github.com/global-121/121-platform/pull/1542/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R136

I think that should be 'moved down here', so we get the order:

1. getting started / setup dev-environment
2. common issues (or as a sub-section of the previous)
3. testing
4. unit-tests (or as a sub-section, might be even better)

elwinschmitz (changes requested):
I think we should split the whole README into individual files, as it gets quite big now...
But "having everything in 1 place" also has its benefits... ;)

diderikvw:
Testing "To set up a local development-environment:", "On Linux":
- After installing NVM the command in the README does not work:
> nvm install && nvm install-latest-npm
results in: "No .nvmrc file found"
I believe it would be better to:
- First update NVM to the latest version by using the command (unless we need a specific NVM-version for the install to work):
> nvm install-latest-npm
- Then have the user clone the repo
- Then have the user go to the repo folder and run the command:
> nvm install
In this way nvm "sees" which version of NodeJS needs to be installed.

elwinschmitz:
> Testing "To set up a local development-environment:", "On Linux":
> 
> After installing NVM the command in the README does not work:
> 
> > nvm install && nvm install-latest-npm
> 
> results in: "No .nvmrc file found"
> 
> I believe it would be better to:
> First update NVM to the latest version by using the command (unless we need a specific NVM-version for the install to work):

No, that's not what it does. You would need _a_ version of the tool NVM.
In our code-base is a `.nvmrc`-file which specifies which version of Node.js is required.
So this will install that version AND the latest-version-of-npm-for-that-node-version.

> > nvm install-latest-npm
> 
> Then have the user clone the repo
> Then have the user go to the repo folder and run the command:
> 
> > nvm install
> 
> In this way nvm "sees" which version of NodeJS needs to be installed.

These commands need to be run AFTER the repo is checked out. Installing NVM itself can happen at another time before that.

README.md
omendram: @elwinschmitz  I wrote this because it was observed that when installing from root folder, repeatedly I have observed errors such as:
`Local package.json exists, but node_modules missing, did you mean to install?`

diderikvw:
In section "Run in Development" in file "services/README.md#getting-started--installation":
> docker-compose -f  docker-compose.yml -f  docker-compose.development.yml up -d --build

Results in:
ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.

omendram:
> In section "Run in Development" in file "services/README.md#getting-started--installation":
> 
> > docker-compose -f  docker-compose.yml -f  docker-compose.development.yml up -d --build
> 
> Results in:
> ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
> 
> I ran the command with sudo instead, which seems to work (and which is bad practice, see Elwin's comment, Omendra and I also concluded that last week)

This is an issue that we encountered last time as well, at least in my observation installing docker only seems to work with "sudo" commands. Which is why i have also added that if facing problems with Docker, such as can not connect to daemon, we should try "sudo" prefix.

omendram:
> > In section "Run in Development" in file "services/README.md#getting-started--installation":
> > > docker-compose -f  docker-compose.yml -f  docker-compose.development.yml up -d --build
> > 
> > 
> > Results in:
> > ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
> > I ran the command with sudo instead, which seems to work (and which is bad practice, see Elwin's comment, Omendra and I also concluded that last week)
> 
> This is an issue that we encountered last time as well, at least in my observation installing docker only seems to work with "sudo" commands, I was not able to install and run docker without sudo privileges. Which is why i have also added that if facing problems with Docker, such as can not connect to daemon, we should try "sudo" prefix.

diderikvw:
In section "Run in Development" in file "services/README.md#getting-started--installation":

After the docker-compose command finished, the applications inside the container are still 'starting up' and this is not visible to the user. I suggest to add text to let the user wait, or have them use the command:

> sudo docker logs -f 121-service

So they can inspect what is going on.
And then let the user test if the services are running, by e.g. going to (which the user may already do as it is in the API section that follows, still good to be explicit, especually becayse they need to wait):
http://localhost:3000/docs/

So just like we let the user test for successful docker install etc., also have them test this.

diderikvw:
I completed the local dev install and can access/open all 4 interfaces in a web browser.

However, I cannot log into the HO Portal with the account specified in the .env, so there seems to be something wrong there. Perhaps the database seeding again, will look into that.

Furthermore, it would be helpful for the user if the README explained how to access each interface after starting them, do they can test / play around. And then suggesting "next steps" in development. Maybe pointing to the manual to get to understand the HO Portal and how to run a program from start to end. And of course how they can contribute as (volunteer) developer by pointing to 510.global for example.

README.md
```diff
@@ -35,20 +35,17 @@ To set up a local development-environment:
...
-      nvm install && nvm install-latest-npm
-
-- Install Git
-
-      sudo apt install git-all
+- Install Git: `sudo apt install git-all`
+- Clone the Git repository: `git clone https://github.com/global-121/121-platform.git`
+- Install Node.js: To install the required version of Node.js and to prevent conflicts between projects or components using other versions of Node.js, use [NVM - Node Version Manager](http://nvm.sh/).  
+  After installing NVM run: `nvm install-latest-npm`
+- `cd` to the directory of the cloned repository above and execute `nvm install`. In this way nvm "sees" which version of NodeJS needs to be installed.
```
elwinschmitz, line 48:
```suggestion
  After installing NVM run, from the root of this repository:  
  `nvm install && nvm install-latest-npm`
  So that the required version of Node.js and `npm` get installed.
```

README.md
```diff
@@ -87,27 +84,21 @@ To set up a local development-environment:
...
+- Clone the Git repository: `git clone https://github.com/global-121/121-platform.git`
 - Install Node.js  
   To install the required version of Node.js and to prevent conflicts between projects or components using other versions of Node.js, use [NVM - Node Version Manager](http://nvm.sh/).  
   After installing NVM run:
 
       nvm install && nvm install-latest-npm
-
-- Install Git  
-  <https://git-scm.com/book/en/v2/Getting-Started-Installing-Git> > Installing on macOS (section)
+- `cd` to the directory of the cloned repository above and execute `nvm install`. In this way nvm "sees" which version of NodeJS needs to be installed.
```
elwinschmitz, line 94: See previous comment:

```
  After installing NVM run, from the root of this repository:  
  `nvm install && nvm install-latest-npm`
  So that the required version of Node.js and `npm` get installed.
```

elwinschmitz (changes requested):
Some small tweaks to the assumed functionality of NVM and setting up the Node.js environment.

PR #1582, opened 2020-11-24
AB#4604 payment request per PA + cleanup

jannisvisser:


jannisvisser:
@RubenGeo @elwinschmitz I believe I'm done with these code-changes + successfully testing various scenarios I can think off. Given the importance I was wondering if (next to normal reviewing) you have input on which scenario's we should test to be complete..

I have tested with 1 PA, multiple PA's, all error-reasons I can think off (Dutch phone number, simulator not open, wrong API-key, etc.). I have tested that NLD still works (it does).

services/121-service/src/programs/fsp/africas-talking.service.ts
```diff
@@ -59,67 +62,22 @@ export class AfricasTalkingService {
...
-          installment: String(installment),
-          did: String(item.did),
-          amount: String(amount),
-        },
-      };
-      payload.recipients.push(recipient);
-    }
+    const recipient = {
+      phoneNumber: paymentData.paymentAddress,
+      currencyCode: process.env.AFRICASTALKING_CURRENCY_CODE,
```
elwinschmitz, line 67: Isn't the currency defined on the program?

So it would involve to query that data from here, based on the `programId`, or send only the value along all the way along the 'request-chain'...
I think it would depend on "how we handle other program-specific things" in these services to decide wich approach would be better scalable/reusable...

PR #1608, opened 2020-12-01
AB#4975 feat: support multiple locations

gulfaraz:


interfaces/Referral-App/src/app/referral/referral.page.html
```diff
@@ -81,11 +82,31 @@
...
 
-<ion-footer>
+<ion-content *ngIf="!isSupportedRegion()">
+  <ion-card *ngIf="!category">
+    <ion-card-header>
+      <ion-card-title>
+        The Netherlands Red Cross provides helpful information for
+        <span
+          *ngFor="let supportedRegion of supportedRegions; first as isFirst; last as isLast"
+          >{{ isFirst ? '' : (isLast ? ' and' : ',')}}
```
elwinschmitz, line 92: Fancy. :)

We'll need to rework this in a more "scalable" way when we want to make the interface multi-lingual though...

interfaces/Referral-App/src/app/referral/referral.page.html
```diff
@@ -81,11 +82,31 @@
...
+          <a routerLink="/{{toSnakeCase(supportedRegion)}}"
```
elwinschmitz, line 93: The right syntax would be:
```suggestion
          <a [routerLink]="['/', toSnakeCase(supportedRegion)]"
```

interfaces/Referral-App/src/app/referral/referral.page.html
```diff
@@ -81,11 +82,31 @@
...
+    </ion-card-header>
+  </ion-card>
+</ion-content>
+
+<ion-footer *ngIf="isSupportedRegion()">
   <ion-toolbar>
     <ion-buttons slot="start" *ngIf="category">
       <ion-back-button
-        defaultHref="/tabs/referral"
+        defaultHref="{{getRegionHref()}}"
```
elwinschmitz, line 109: Same syntax issue here:
```suggestion
        [defaultHref]="getRegionHref()"
```

interfaces/Referral-App/src/app/referral/referral.page.ts
```diff
@@ -28,34 +31,65 @@ export class ReferralPage {
...
+      )
+    );
+  }
+
+  public toSnakeCase = (string) => {
+    return string
+      .replace(/\W+/g, ' ')
+      .split(/ |\B(?=[A-Z])/)
+      .map((word) => word.toLowerCase())
+      .join('-');
```
elwinschmitz, line 70: Isn't this "kebab-case"? https://en.wikipedia.org/wiki/Letter_case#Special_case_styles

interfaces/Referral-App/src/app/referral/referral.page.ts
```diff
@@ -17,7 +17,10 @@ import { ReferralPageDataService } from 'src/app/services/referral-page-data.ser
   templateUrl: 'referral.page.html',
   styleUrls: ['referral.page.scss'],
 })
-export class ReferralPage {
+export class ReferralPage implements OnInit {
+  public region: string;
+  public supportedRegions: string[] = ['amsterdam', 'utrecht'];
```
elwinschmitz, line 22: There isn't really a more convenient place 'higher up' then this, right?

As I feel it should be defined somewhere in the `environment`-file, but we would have the 'hard' connection to these 2 strings anyway in other places...

interfaces/Referral-App/src/app/referral/referral.page.ts
```diff
@@ -28,34 +31,65 @@ export class ReferralPage {
...
+                this.subCategories = subCategories;
+                this.offersService.getOffers(this.region).then((offers) => {
+                  this.offers = offers;
+                  this.readQueryParams();
+                });
+              });
           });
         });
-      });
+    }
```
elwinschmitz, line 92: Hmm.. 🤔  This could use some un-indenting with `async`/`await` some day... ;)

interfaces/Referral-App/src/app/services/spreadsheet.service.ts
```diff
@@ -15,7 +15,10 @@ import { environment } from 'src/environments/environment';
 })
 export class SpreadsheetService {
   private spreadsheetURL = environment.google_sheets_api_url;
-  private spreadsheetId = environment.google_sheets_sheet_id;
+  private spreadsheetId = {
+    amsterdam: environment.google_sheets_sheet_id_amsterdam,
+    utrecht: environment.google_sheets_sheet_id_utrecht,
+  };
```
elwinschmitz, line 21: Wouldn't it be possible to supply the "spreadsheetId" from 'outside'?
So that this `SpreadsheetService` doesn't need to be aware of the 'fixed' list of ids?

elwinschmitz (changes requested):

interfaces/Referral-App/src/app/referral/referral.page.ts
gulfaraz: Do env vars support JSON values? Then we can move these there.

interfaces/Referral-App/src/app/services/spreadsheet.service.ts
gulfaraz: I had to choose between supplying the region name and supplying the sheet id. The region named felt cleaner so rolled with that.

interfaces/Referral-App/src/app/referral/referral.page.ts
elwinschmitz: No, not really...

We use a "comma-separated-list" for the LOCALES:
https://github.com/global-121/121-platform/blob/master/interfaces/PA-App/src/app/personal-components/select-language/select-language.component.ts#L46

So the ENV would be: `SUPPORTED_REGIONS=amsterdam,utrecht`; And you could use that as an array somewhere in the code then...

But that would make it actually more complex and harder to understand, I think.
As there needs to be a 'hard' connection between the SpreadsheetId and the URL-path...

But putting a 'real' JSON-object like this in an ENV-var, feels weird:
```
REGIONS={ "amsterdam": "abcDeFGhI", "utrecht": "JklMnOPqRst" }
```

interfaces/Referral-App/src/app/services/spreadsheet.service.ts
elwinschmitz: Yeah, otherwise the 'mapping'-logic would have to live somewhere else as wel..

Hmm.. For now, only these 2 regions.... Lets see if we can get rid of it soon altogether. ;)

elwinschmitz:
> Azure Static Web Apps: Your stage site is ready! Visit it here: https://ashy-water-006b5e603-1608.westeurope.azurestaticapps.net

These real-time-up-to-date staging-environments are REALLY nice! @gulfaraz :D

elwinschmitz (approved):

PR #1610, opened 2020-12-02
AB#4978 feat: use skeleton text and loaders when fetching data

gulfaraz:


elwinschmitz (approved):
It looks 😭 in the template, but 🤩 in the user's browser. :D 🚀

PR #1617, opened 2020-12-02
feat: move region level config to env files AB#4349

gulfaraz:


elwinschmitz (approved):

elwinschmitz:
Do you have access to the test-vm already? The `.env`-file there also needs to be updated with these changes, so it continues to work there...

elwinschmitz:
@gulfaraz Nevermind, I've changed the `.env`-file already. So when this is merged it will be used immediately.

PR #1618, opened 2020-12-02
Add Typedoc and generated doc files

diderikvw:
Adds the Node Module [Typedoc](http://typedoc.org/guides/installation/) for automatically generating HTML documentation from Typescript source code of the 121 service (only).

To Install: 
> npm install

To use:
> npx typedoc --out docs src

This generates files in the docs folder bases on source code in the src folder. Open by double clicking index.html inside the folder (=open index.html in your web browser).
All [Typedoc config](http://typedoc.org/guides/options/) left to default.  

We can "inspect" the resulting docs to see if it is helpful and what can be improved.
It is also possible to [generate the documentation in a pipeline](http://typedoc.org/guides/installation/#third-party-tools), and then always have the most up-to-date version available under a URL of the 121 platform, like the Swagger docs. I would need help setting this up.


services/121-service/package.json
```diff
@@ -67,7 +67,8 @@
     "eslint-plugin-prettier": "^3.1.4",
     "jest": "^24.9.0",
     "prettier": "^1.18.2",
+    "ts-jest": "^24.0.0",
     "tsc-watch": "^2.4.0",
-    "ts-jest": "^24.0.0"
+    "typedoc": "^0.19.2"
```
elwinschmitz, line 72: Are these changes (and in `package-lock.json`) generated via the command run on the Docker-environment?

So via `docker-compose exec 121-service npm install typedoc --save-dev`?
Otherwise the dependencies-of-this-dependency might be specific for your platform (it could be the same in the end, but the whole point of the Docker-env is predictability and reproducibility...)

elwinschmitz: I would suggest to:

- [ ] add the `docs/`-folder to the `.gitignore`-file (so nothing of these files need to be committed)
- [ ] Add a "npm-script" to this file (in the "scripts"-block above) like:
  ```
  "docs:generate": "typedoc --out docs src",
  ```
- [ ] Add a section to the README.md file for this service (or in the services/README, as we want to enable this in all services, I guess.)  
  In it we could explain why you want to generate this and when/how you can look at it. Including that you need to open some specific file/URL in your browser etc...

elwinschmitz (changes requested):
Thanks for your first PR! 🥳

elwinschmitz:
We could even include this 'build-task' in the deployment-pipeline, for example by including:

```
  "postinstall": "npm run docs:generate",
```

That way, it will automatically be generated the first time (and any future time) you run "`npm install`", so that also included the times the deploy-script would run that on the test-vm. So the documentation will automatically be available on the test-VM... (But we do need to enable a public URL for it as well then... for example by setting up an alias in Apache or something..)

services/121-service/package.json
elwinschmitz: You can make new commits to this branch, but maybe better to "amend"  the previous commit and "force-push" that new commit(s) to this branch.

That way the git-history will not be filled with all these (useless) changes to all these files; as it will only occupy space and make the repository slower over time.

diderikvw:
Hi @elwinschmitz, I made the changes as you suggested, but still have a few open ends. And I also do not know how to do the amend/force-push thing (now I just created a new branch locally fresh from the master branch). Let's do some "pair programming" soon and get this done together.

PR #1638, opened 2020-12-08
feat: font should fallback to Open Sans if Montserrat doesn't work AB…

gulfaraz:
…#4972


elwinschmitz:
@gulfaraz Are there no `*.woff` files available?
And there are some extra CSS-properties we can/should use to ease the user-experience.

See: https://css-tricks.com/font-display-masses/ etc.

gulfaraz:
@elwinschmitz I was only able to find the `*.ttf` formats. Please feel free to add other formats for better UX.

PR #1646, opened 2020-12-09
AB#4617 see if actual voucher was sent to PA

jannisvisser:


jannisvisser:
This is old and included now in https://github.com/global-121/121-platform/pull/1667

PR #1720, opened 2021-01-07
Unused voucher export frontend AB#5107

RubenGeo:


interfaces/HO-Portal/src/app/pages/payment/payment.page.ts
```diff
@@ -10,18 +12,33 @@ import { ProgramPhase } from 'src/app/models/program.model';
...
   constructor(
     private route: ActivatedRoute,
     private authService: AuthService,
+    private programsService: ProgramsServiceApiService,
   ) {}
 
   ngOnInit() {}
 
+  async isIntersolve() {
+    const intersolveString = 'intersolve';
```
elwinschmitz, line 32: I think it's not really useful to have this very verbose extra string here; as it isn't reused anywhere anyway..

interfaces/HO-Portal/src/app/pages/payment/payment.page.html
```diff
@@ -12,12 +12,20 @@
     [phaseReady]="isReady"
   ></app-phase-next>
 
+  <ion-row *ngIf="!isIntersolve">
```
elwinschmitz, line 15: You'd want to show this ONLY if the program uses Intersolve as an FSP, right? (So without the `!`)?

interfaces/HO-Portal/src/app/program/export-list/export-list.component.ts
```diff
@@ -84,7 +84,9 @@ export class ExportListComponent implements OnChanges {
           this.exportType === ExportType.selectedForValidation) ||
         (activePhase.name === ProgramPhase.payment &&
           this.exportType === ExportType.payment &&
-          this.paymentExportAvailable)) &&
+          this.paymentExportAvailable)) ||
+      (activePhase.name === ProgramPhase.payment &&
+        this.exportType === ExportType.unusedVouchers) &&
       this.userRole === UserRole.ProgramManager
     );
   }
```
elwinschmitz, line 92: Hmm.. I think all this logic shouldn't be handled inside this component..

But for now, this seems like the next-best-place to combine everything...

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -275,4 +279,4 @@
     "show": "Show password",
     "hide": "Hide password"
   }
-}
+}
```
elwinschmitz, line 282: Something with your local settings/config of Prettier isn't working as it should. :/

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -226,6 +226,10 @@
         "payment": {
           "btn-text": "Export list",
           "confirm-message": "This will download a CSV-file with all People Affected included in this payment, to be shared with institutions requiring this data."
+        },
+        "unused-vouchers": {
+          "btn-text": "Export list of unused vouchers",
+          "confirm-message": "This will download a CSV-file with all unused voucher."
```
elwinschmitz, line 232:
```suggestion
          "btn-text": "Export unused vouchers",
          "confirm-message": "This will download a CSV-file with all unused voucher(s)."
```

elwinschmitz (changes requested):
I think it is good to base the decision "to show the export/download-button or not" on the FSP of the Program.

Maybe we can create some more structured helper methods around it, so that we don't have to do a lot of these steps "by hand" everywhere. (Like doing the request for the program-object etc.)

PR #1745, opened 2021-01-14
AB#5575 Store lookup response

RubenGeo:


elwinschmitz (approved):

services/121-service/src/sovrin/credential/credential.service.ts
```diff
@@ -159,7 +161,9 @@ export class CredentialService {
     const cleanedAnswers = [];
     for (let answer of answers) {
       if (phonenumberTypedAnswers.includes(answer.attribute)) {
-        answer.answer = answer.answer.replace(/\D/g, '');
+        answer.answer = await this.lookupService.lookupAndCorrect(
+          answer.answer,
+        );
         cleanedAnswers.push(answer);
       } else {
         cleanedAnswers.push(answer);
```
elwinschmitz, line 169: Small refactoring:

```
      if (phonenumberTypedAnswers.includes(answer.attribute)) {
        answer.answer = await this.magic();
      }
      cleanedAnswers.push(answer);
```

So you only do some _extra_ magic if the condition applies, but in any scenario you store `answer` in the array...
Easier to extend later with multiple other extra conditions etc...

services/121-service/src/notifications/lookup/lookup.service.ts
```diff
@@ -19,4 +19,19 @@ export class LookupService {
     }
     return { result: numberCorrect };
   }
+
+  public async lookupAndCorrect(phoneNumber: string): Promise<string> {
+    try {
+      const lookupResonse = await twilioClient.lookups
```
elwinschmitz, line 25:
```suggestion
      const lookupResponse = await twilioClient.lookups
```

Missing `p`. ;)

PR #1766, opened 2021-01-20
Created extra npm method to develop locally without dockerized services

RubenGeo:
Related [AB#3091](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/3091)


services/docker-compose.development.local.yml
```diff
@@ -0,0 +1,16 @@
...
+# The settings from this file will override the values from `docker-compose.yml`
+# When run via:
+#     docker-compose -f docker-compose.yml -f docker-compose.development.local.yml
+
+# This makes sure that the 121-service and PA-accounts service immidiately shutdown after being created
+
+services:
+  121-service:
+    command: "true"
+    entrypoint: "true"
```
elwinschmitz, line 12: These values aren't really valid `commands` are they?
https://docs.docker.com/compose/compose-file/compose-file-v3/#command
https://docs.docker.com/compose/compose-file/compose-file-v3/#entrypoint

I think setting it to `""` (empty string) or `{}`(empty object) is safer (if that provides the desired functionality)..

A valid "not really doing anything"-command could also be `cwd` or `ls`, maybe...

elwinschmitz (changes requested):

PR #1803, opened 2021-02-02
AB#5902 Prevent unnecessary throws of 401-errors

elwinschmitz:


jannisvisser (commented):
@elwinschmitz have you tested anything? (I guess send a payment, and then check in Twilio logs that there are no more 401 errors?)

PR #1809, opened 2021-02-03
AB#5913  multiple roles per user

jannisvisser:


services/121-service/src/user/dto/create-user.dto.ts
```diff
@@ -8,17 +8,14 @@ export class CreateUserDto {
...
 
-  @ApiModelProperty()
+  @ApiModelProperty({ example: 'password' })
   @IsNotEmpty()
   @MinLength(8)
   public readonly password: string;
 
   @ApiModelProperty({
-    example: 'admin / aidworker / project-officer / program-manager',
+    example: ['admin', 'aidworker', 'project-officer', 'program-manager'],
```
elwinschmitz, line 17: Can you include them here from the "enum", instead of as hard-coded stings?

services/121-service/src/user/user.entity.ts
```diff
@@ -5,11 +5,13 @@ import {
   BeforeInsert,
   OneToMany,
   ManyToMany,
+  JoinTable,
 } from 'typeorm';
 import { IsEmail } from 'class-validator';
-const crypto = require('crypto');
+import crypto from 'crypto';
```
elwinschmitz, line 11: The `'crypto'`-dependency can also be removed from the `package.json`; If this was the only usage for it...

jannisvisser: @elwinschmitz all of this is already changed on my local machine ;)

jannisvisser: @elwinschmitz In general, you can assume with me that you should start reviewing only when I move the devops task to review. Otherwise it is just intermediate work.

elwinschmitz: These comments are in a large part "notes to my future self" ;)

services/121-service/src/user/user.service.ts
```diff
@@ -183,7 +172,7 @@ export class UserService {
       {
         id: user.id,
         email: user.email,
-        role: user.role,
+        roles: user.roles,
```
elwinschmitz, line 175: Here the JWT should only include the "text"-version of the roles, not their label/description...

services/121-service/src/user/user.service.ts
```diff
@@ -183,7 +172,7 @@ export class UserService {
       {
         id: user.id,
```
elwinschmitz, line 173: I think we can leave the (internal) user-id out of the JWT.

elwinschmitz (commented):
Basically some to-do's for myself on Monday. :)

services/121-service/src/roles.guard.ts
```diff
@@ -44,7 +47,11 @@ export class RolesGuard implements CanActivate {
       );
       const user = await this.userService.findById(decoded.id);
 
-      hasAccess = roles.includes(user.user.role);
+      const userRoles = user.user.roles.map(role => role.role);
+      const overlappingRoles = endpointRoles.filter(role =>
+        userRoles.includes(role),
+      );
+      hasAccess = overlappingRoles.length > 0;
```
elwinschmitz, line 54: This can be moved into the `user.service.ts`, I think.
Similar to how the HO-Portal handles it.

services/121-service/src/user/user.controller.ts
```diff
@@ -48,26 +48,20 @@ export class UserController {
...
-    } else if (_user.status == 'inactive') {
-      throw new HttpException(
-        'Account deactivated. Contact organization administration.',
-        HttpStatus.UNAUTHORIZED,
-      );
     }
 
     const token = await this.userService.generateJWT(_user);
-    const { email, role, status } = _user;
+    const { email, roles } = _user;
```
elwinschmitz, line 54: I think the only thing that needs to be returned to the interface is the `token`, as that contains all the data we'd need already (in a verifiable way).
And the other data isn't used.

So the only thing to return would be:
```
{ 
  token: '...'
}
```

services/121-service/src/user/user.service.ts
```diff
@@ -197,8 +186,7 @@ export class UserService {
       id: user.id,
       email: user.email,
       token: this.generateJWT(user),
-      role: user.role,
-      status: user.status,
+      roles: user.roles,
```
elwinschmitz, line 189: (See elsewhere; the only thing to return would be the `token` (which would contain the whole `user`).

interfaces/HO-Portal/src/app/auth/auth.service.ts
```diff
@@ -71,11 +71,6 @@ export class AuthService {
         if (user.role) {
           user.roles = [user.role];
         }
-
-        // 'Clean' roles-object into flat list of roles
```
jannisvisser: @elwinschmitz this code is not needed any more I think with the back-end change below. You need to remove this same piece of code also from the AW-app PR.

services/121-service/src/user/user.service.ts
jannisvisser: done in next commit

PR #1817, opened 2021-02-09
Refactor example-values in Swagger-UI (use enums, etc), remove unused FSP-property

elwinschmitz:


elwinschmitz:
@jannisvisser  Some small changes I stumbled upon with the roles-refactoring..

I think 'just removing' the `apiUrl`-property doesn't really need any special attention more then this, right?

jannisvisser:
@elwinschmitz no, that should be OK like this.

PR #1866, opened 2021-03-02
AB#6205 Add "program with validation" seed-script

jannisvisser:


interfaces/PA-App/src/app/personal-components/preprinted-qrcode/preprinted-qrcode.component.html
```diff
@@ -1,4 +1,4 @@
-<section *ngIf="isCanceled === false">
+<section *ngIf="!isCanceled">
```
elwinschmitz, line 1: I think the `=== false` was put there on purpose, because it doesn't match when it it is `undefined`...

services/121-service/seed-data/program/program-validation.json
```diff
@@ -0,0 +1,119 @@
...
+      "whatsappPayment": "This is a message from the Red Cross.\n\nYour €35 Albert Heijn supermarket voucher is ready to be sent.\n\nPlease reply “yes” to this message in order to receive the voucher.",
+      "whatsappVoucher": "This is a message from the Red Cross.\n\nWith this message you received 2 pictures.\nOne is the Albert Heijn supermarket voucher (cadeaukaart) with the barcode.\nThe other is an explanation of how you can use the voucher to pay in Albert Heijn.\n\n The value of this voucher is €35.\nYou can spend the voucher at any Albert Heijn supermarket in the Netherlands.\n\nYou don’t have to spend everything at once. So keep this voucher in your phone and re-use it until the amount is finished.\n\nYou will receive a €35 Albert Heijn voucher every week on Tuesday until 31 May 2021 the latest.\n\nDo you have questions? Send us a message on WhatsApp: https://wa.me/31614458781",
+      "whatsappReply": "This is an automated message, for questions please contact the Red Cross Helpdesk on WhatsApp: +31614458781"
```
elwinschmitz, line 59: Maybe we want more 'generic' texts for this test-program?

And maybe not all different languages...

interfaces/PA-App/src/app/personal-components/preprinted-qrcode/preprinted-qrcode.component.html
jannisvisser: @elwinschmitz sure.. but it wasn't working like this, because the component wasn't showing while it should.

jannisvisser: @elwinschmitz just found out that the same change is causing also the store-credential component to not show.

that being said:
- I do admit that I haven't tested the reverse yet with this change (program without validation)
- and I am interested to hear more about it. Because if you think it's needed for some purpose, while I think the reverse is needed for some purpose, that should be straightened out.

interfaces/PA-App/src/app/personal-components/preprinted-qrcode/preprinted-qrcode.component.ts
```diff
@@ -76,7 +76,9 @@ export class PreprintedQrcodeComponent extends PersonalComponent {
   async checkValidation() {
     if (!this.program.validation) {
       this.cancel();
+      return;
     }
+    this.isCanceled = false;
```
jannisvisser, line 81: @elwinschmitz there are 3 more components (besides these 2) where "isCanceled === false" is used. I haven't  touched those, because nothing is going wrong apparently. Feel free to touch those as well if you think it's needed though.

elwinschmitz: @jannisvisser Some of those components have a line of `this.isCanceled = false` already somewhere else in the code. I think that was where this component was 'out of line'..

PR #1867, opened 2021-03-03
AB#6478 isSpoken in qr-component

jannisvisser:


elwinschmitz:
Weird that in all the other instances where we use `ngIf="isCanceled === false"` all the "dialogue-turn"-children do NOT have any explicit `isSpoken`-attributes... 😕 

See: https://github.com/search?q=org%3Aglobal-121+%22ngIf%3D%22isCanceled%22&type=code

PR #1988, opened 2021-04-20
AB#7204 Make "export-voucher"-endpoint always return the most-recent image

elwinschmitz:
For "installment #17" PA will have multiple images, in `getVoucher` an array-method `find` is used;
This will return at the _first_ successful match (and thus the 'old/initial' voucher image of the correct installment-17).

By reversing the array, the most-recent image of the INcorrect-installment-17(should be 23) will be returned.

### DIFF
--- a/services/121-service/src/programs/fsp/intersolve.service.ts
+++ b/services/121-service/src/programs/fsp/intersolve.service.ts
@@ -375,7 +375,7 @@ export class IntersolveService {
       );
     }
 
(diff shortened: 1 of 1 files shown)


elwinschmitz:
@jannisvisser This is to document the current state of the LVV-production server (where I changed this by hand).

I don't think we need to merge this change per se, but as documentation this was most convenient. ;)

PR #1992, opened 2021-04-21
AB#7209 Fix: Add missing import in Module

elwinschmitz:


elwinschmitz:
@jannisvisser  Weird that this missing import didn't fail the tests before...

It showed as though the 121-service had successfully started, but as soon as you would make a request, it would fail/crash/stop... Apparently starting up the test-version of the app doesn't catch that...

jannisvisser:
@elwinschmitz yes, I also suddenly had this locally.. really weird.

PR #1995, opened 2021-04-21
Refactor.replace did

jannisvisser:


jannisvisser:
@elwinschmitz 
- I tested a full happy flow of this 
- not all sidepaths though like the specific change FSP workflows etc. It should be pretty straightforward though right..
- I imposed a length of (29,36) on referenceId, where 29 is the (minimum) length of (old) dids, while 36 is the length of (new) uuids.
- BTW: I have now imposed this same constraint at like 10 different places, instead of at one place and then referenced that in all .dto files (suggestion welcome)
- I came across an error with 'whatsappPaymentMultiple' not being available in my seeded validation program, so I thought it a good idea to impose a fallback to the single message
- I also came across (many) issues relating to the handleProof component, but that will be handled in the other subtask. 
- I haven't waited for check results yet ;)

features/AW-App/Create_program_credential.feature
```diff
@@ -1,13 +1,13 @@
 @aw-app
-Feature: Create program credential
+Feature: Upload validation data
```
elwinschmitz, line 2: The file itself needs a new name now too. ;)

interfaces/PA-App/src/app/services/padata-types.enum.ts
```diff
@@ -1,6 +1,6 @@
 export enum PaDataTypes {
   conversationHistory = 'conversationHistory',
-  did = 'did',
+  referenceId = 'referenceId',
```
elwinschmitz, line 3: This will require a migration of all existing PA's data.

elwinschmitz (approved):
I haven't seen _every_ line of code... but 👍

PR #2027, opened 2021-05-04
Feat.upload test pas

jannisvisser:


services/121-service/src/connection/connection.service.ts
```diff
@@ -126,7 +134,100 @@ export class ConnectionService {
...
+      connection.customData = JSON.parse(JSON.stringify({}));
+      connection.customData[CustomDataAttributes.nameFirst] = record.nameFirst;
+      connection.customData[CustomDataAttributes.nameLast] = record.nameLast;
+      connection.customData[CustomDataAttributes.phoneNumber] =
+        record.phoneNumber;
+      connection.customData[CustomDataAttributes.whatsappPhoneNumber] =
+        record.whatsappPhoneNumber;
+      const fsp = await this.fspRepository.findOne({
+        where: { fsp: record.fspName },
+      });
```
elwinschmitz, line 168: Maybe we can do this query outside of the for-loop?

Or at most 1 time per FSP? So we create a method to `getFSPByName`, that has a local cache of every `fspName` it has already requested before... So it can quickly return `_that_ value instead of doing the whole query again...

elwinschmitz: Yeah, I've tried it locally. And it is a major difference (almost twice as fast); but the other steps after that (so `calculateInclusionScore` and `storePrefilledAnswersTestRegistrationsNL`) take up a much bigger chunk of the total time anyway...

And for 1000 PA it is only a matter of 41 seconds versus 97 seconds. So not nearly close to the 10 minutes or so...

PR #2062, opened 2021-05-12
Feat.payment amount multiplier AB#7740

jannisvisser:


jannisvisser:
@elwinschmitz some context on this last (3rd) commit
- This is unfinished (but working) code, because I'm wrapping up, so you can potentially continue with it
- My idea was to - instead of creating yet another endpoint for updating one specific connection attribute - to move towards some endpoint-grouping by making a generic update(Connection)Attribute endpoint
- which (in current version of code) only handles the paymentAmountMultiplier
- but if extended, could extend other attributes ..
- .. including also customData attributes > the if / else could be extended to check for connection.customData[attribute] instead of for connection[attribute]
- which could get rid of the customData endpoint(s) > I also see no reason keep a separate custom-data and custom-data/overwrite endpoint
- but also of the change phone / change fsp / etc. endpoints.. 
- In the new refactoring Diderik and I have already discussed that we want to move to endpoint minimization, where basically per entity you have e.g. one GET, one POST and one PUT endpoint, instead of the whole extravaganza of current endpoints. They could largely be grouped by using parameters/body input for this 1 endpoint.
- this particular endpoint would then become the PUT endpoint for the connection entity > it's still a POST now, but this could be easily changed already I think.

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -245,7 +245,7 @@
         "pageTitle": "Program JSON"
       },
       "bulk-import": {
-        "import-error": "Something went wrong with the import. Check the required extension (.csv) and format of the CSV-file.",
+        "import-error": "Something went wrong with the import. <ul><li>Check the required extension (.csv)</li><li>Check the format and column-names of the CSV-file</li><li>The 'paymentAmountMultiplier' column should contain only whole numbers >= 1</li></ul>",
```
elwinschmitz, line 248: I think we should split this up into separate translatable sentences and format them into a list in the template.
Having this type of HTML in the translation-process is very annoying/complex.

```suggestion
        "import-error": {
          "intro": "Something went wrong with the import.",
          "check": {
            "extension": "Check the required extension (.csv)",
            "format": "Check the format and column-names of the CSV-file",
            "column": {
              "paymentAmountMultiplier": "The 'paymentAmountMultiplier' column should contain only whole numbers >= 1",
              "anotherFutureFeature": "The X-column requires Y-values"
            }
          }
        },
```

Something like that...

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -254,7 +254,7 @@
         },
         "timestamp": "Latest import has been done on: {{dateTime}}",
         "btn-text": "Import People Affected",
-        "explanation": "The required format of the CSV-file is: <pre class='ion-color-primary'>phoneNumber,namePartnerOrganization\n+0000000000,Example-Name</pre>",
+        "explanation": "The required format of the CSV-file is: <pre class='ion-color-primary'>phoneNumber,namePartnerOrganization,paymentAmountMultiplier\n+0000000000,Example-Name,1</pre>",
```
elwinschmitz, line 257: I think we need to move the actual file-format/syntax to some other place in the code as well..

That way we can ensure it will stay 'in sync' with the actual functionality..

services/121-service/src/connection/connection.service.ts
```diff
@@ -465,6 +469,36 @@ export class ConnectionService {
...
+    const connection = await this.findOne(referenceId);
+    console.log('typeof connection[attribute]: ', typeof connection[attribute]);
+    let newConnection;
+
+    if (typeof connection[attribute] === 'undefined') {
+      const errors = 'This attribute is not known.';
+      throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
+    }
+
+    if (typeof connection[attribute] === 'number' && Number.isInteger(value)) {
```
elwinschmitz, line 486: A type of "`number`" would not always have to be an integer... I think..

But for now this will work...

services/121-service/src/connection/connection.service.ts
```diff
@@ -465,6 +469,36 @@ export class ConnectionService {
     return note;
   }
 
+  public async updateAttribute(
```
elwinschmitz, line 472: I have the feeling that there _must_ be some helper-methods/tools available in Nest.js to handle this "safely update an entity"-endpoint kind of method...

As it will be a very common thing in CRUD-APIs. I think this 'manual' approach will run 'out of hand' pretty quickly in this one method.

services/121-service/src/connection/dto/update-attribute.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  public readonly referenceId: string;
+  @ApiModelProperty({
+    enum: attributesArray,
+    example: attributesArray.join(' | '),
+  })
+  @IsIn(attributesArray)
+  public readonly attribute: AttributesEnum;
+  @ApiModelProperty({ example: 'new value' })
+  @IsString()
+  public readonly value: string;
```
elwinschmitz, line 21: Not all values need to be `string`s right?

elwinschmitz (changes requested):

services/121-service/src/connection/connection.service.ts
jannisvisser: Good point. I have just committed a new cleaner solution, which makes use of the fact that TypeOrm's .save() functionality also allows partial input and also automatically handles type conversion where necessary.

This new solution already handles customData attributes as well. So if you're on board with this, we could at least replace/remove the customDataOverwrite function with this one already. (Although we might need to check more things like phone number sanitization etc.)

I do realize this is scope creep and not actually part of this item..

services/121-service/src/connection/dto/update-attribute.dto.ts
jannisvisser: changed

services/121-service/src/connection/connection.service.ts
jannisvisser: outdated now

interfaces/HO-Portal/src/assets/i18n/en.json
jannisvisser: can you make this change?

interfaces/HO-Portal/src/assets/i18n/en.json
jannisvisser: can you make this change?

interfaces/HO-Portal/src/assets/i18n/en.json
elwinschmitz: I'll make a separate PR 👍

interfaces/HO-Portal/src/assets/i18n/en.json
elwinschmitz: We'll pick it up in a future PR.

elwinschmitz (approved):

jannisvisser:
@elwinschmitz I have fixed the whatsapp messages
- and tested with local ngrok + twilio sandbox
- but not yet for the (edge) use case of simulanteously having multiple PAs on the same phone number
- and also not yet on test-vm, because only there you can test if the template is passed (the yes reply will fail there, but that doesn't matter)

PR #2157, opened 2021-06-02
Enable payments for 3000+ PAs incl Twilio and Intersolve mocks 

jannisvisser:


jannisvisser:
@elwinschmitz I have written this Twilio-mock
- As explained I could not find any mock-options for the Twilio SDK, and I gathered that it was relatively easy to make something myself, given the limited features we actually use.
- You have to switch TWILIO_MOCK in .env 
- This is working as intended for sending whatsapp & SMS, which are the only functionalities we're actually using
- I have additionally made some beginnings for 
* Voice (although that's not used anymore, and I find that the non-mock version of Voice is actually breaking currently)
* Lookup > this is not working as intended yet
* validateRequest > this seems to work, although I'm not sure what it actually does

services/121-service/src/notifications/twilio.mock.ts
```diff
@@ -0,0 +1,93 @@
...
+        numSegments: twilioMessagesCreateDto.mediaUrl ? '1' : '0',
+        direction: 'outbound-api',
+        from: twilioMessagesCreateDto.from,
+        to: twilioMessagesCreateDto.to,
+        dateUpdated: new Date(),
+        price: null,
+        errorMessage: null,
+        uri:
+          '/2010-04-01/Accounts/ACc6b5fa2276ff6b64cdbed0099fd9bc0b/Messages/SM2659aa9303394e01b30f33cbd92c4cf3.json',
+        accountSid: 'ACc6b5fa2276ff6b64cdbed0099fd9bc0b',
```
elwinschmitz, line 35: These values are not super-sensitive, but they don't belong in code.
You can use their ENV-versions instead...

```suggestion
          `/2010-04-01/Accounts/${process.env.TWILIO_SID}/Messages/SM2659aa9303394e01b30f33cbd92c4cf3.json`,
        accountSid: process.env.TWILIO_SID,
```

services/121-service/src/notifications/twilio.mock.ts
```diff
@@ -0,0 +1,93 @@
...
+        numMedia: twilioMessagesCreateDto.mediaUrl ? '1' : '0',
+        status: 'accepted',
+        messagingServiceSid: twilioMessagesCreateDto.messagingServiceSid,
+        sid: 'SM2659aa9303394e01b30f33cbd92c4cf3',
```
elwinschmitz, line 39: This value is something unique Twilio creates right?
Maybe we can move this to a constant (on top of the file), so we can use the variable throughout each session. (or change it for testing)

services/.env.example
```diff
@@ -101,6 +101,8 @@ TWILIO_PROGRAM_TOKEN_SECRET=
 TWILIO_AUTHTOKEN=
 TWILIO_WHATSAPP_NUMBER=
 TWILIO_MESSAGING_SID=MG
+TWILIO_TEST_FROM_NUMBER_VOICE=+31600000000
+MOCK_TWILIO=True/False
```
elwinschmitz, line 105: If you use 'real' Boolean values here(like `TRUE` or `1`) `dotenv` will convert it to real Boolean variables; so you don't need to compare the string like below.

elwinschmitz (dismissed):

services/.env.example
jannisvisser: @elwinschmitz I tried this, but this doesn't work for me.. When I have MOCK_TWILIO=False and use `if (process.env.MOCK_TWILIO)` then it still evaluates to true.. (in line with it being read as a string and not as a boolean).

Which is also what I remembered from earlier occurrences, which is why I used the string comparison now

jannisvisser:
@elwinschmitz I'm not managing to fix the test. Can you maybe look at it tomorrow?

services/.env.example
elwinschmitz: Hmm.. I thought that I committed it, but now I can't seem to find it anymore..

I think because the values `0`, `1`, `TRUE`, `FALSE` are also valid Javascript, but `False` isn't valid JS and therefore, 'just a string'.

So I added a comment to the `.env.example` file etc about it... But it got lost in some conflict-resolving/rebaseing or something... 😕 

Nevermind for now. ;)

interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.html
```diff
@@ -45,10 +45,38 @@ <h3>{{ 'page.program.program-payout.make-export'|translate }}</h3>
...
-  <ion-col size="4">
-    <h3>{{ 'page.program.program-payout.last-payment'|translate:{ nr: lastInstallmentId ? '#' + lastInstallmentId : '-' } }}</h3>
+  <ion-col
+    *ngIf="!!lastInstallmentId"
+    [sizeXs]="12"
+    [sizeMd]="4"
+  >
+    <h3>{{ 'page.program.program-payout.last-payment.last-payment'|translate:{ nr: lastInstallmentId ? '#' + lastInstallmentId : '-' } }}</h3>
+    <p *ngIf="lastPaymentResults">
+      {{ lastPaymentResults.amount }}<br>
```
elwinschmitz, line 55: This shouldn't be here, only for debugging...

PR #2158, opened 2021-06-03
AB#6672 Add Detect-Duplicates export

elwinschmitz:


interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -287,6 +287,10 @@
         "unused-vouchers": {
           "btn-text": "Export list of unused vouchers",
           "confirm-message": "This will download a CSV-file with all unused voucher(s).<br /><br />Note that this action can take a while to complete. Do not leave this page in the meantime."
+        },
+        "duplicate-phone-numbers": {
+          "btn-text": "Export list of duplicate phone-numbers",
+          "confirm-message": "This will download a CSV-file with all PA that share any phone-number."
```
jannisvisser, line 293: all PAs?

jannisvisser: And maybe add a short explanation like 'You can subsequently filter on these PAs and reject them"?

services/121-service/src/programs/program/program.service.ts
```diff
@@ -1117,14 +1130,81 @@ export class ProgramService {
...
+    });
+
+    const duplicates = allConnections.filter(connection => {
+      const others = without(allConnections, connection);
+      const currentPaNumbers = compact([
+        connection.customData[CustomDataAttributes.phoneNumber],
+        connection.customData[CustomDataAttributes.whatsappPhoneNumber],
+      ]);
+
+      const hasDuplicateProgramNr = this.hasDuplicateCustomDataValues(
```
jannisvisser, line 1157: ProgramNr?

elwinschmitz: At first I had `*PhoneNumber` everywhere but I renamed them all to be a bit shorter...

But this is called "program" as it is a program-specific attribute/question.
The PA-property "phoneNumber" is a more generic property, but currently these 2 'columns' are always the same, so I've settled on using this one only.

interfaces/HO-Portal/src/assets/i18n/en.json
elwinschmitz: PA = People Affected is already plural right?

services/121-service/src/programs/program/program.service.ts
jannisvisser: Aha.. but then maybe ProgramPhoneNr? Because I was confused with ProgramId or something.

interfaces/HO-Portal/src/assets/i18n/en.json
jannisvisser: Hmm you are technically maybe right, but in my interpretation PA stands for singular, and thus PAs for plural. Also, because there is otherwise no difference, which is confusing. 

On another note, you could actually argue we should minimize the short-forms completely, so we should say People Affected here then?

PR #2176, opened 2021-06-08
fix: update customData attribute issues AB#8076

jannisvisser:


jannisvisser:
@elwinschmitz there were so many issues with this code... 3 independent errors (and 1 unnecessary line). This clearly was not tested properly by either of us..

PS: Let's use your work-around for now on production, and have this go in the next release.

elwinschmitz:
The PR in which it happened was pretty big, so no real surprise it got overlooked there... ;)
https://github.com/global-121/121-platform/pull/2062/files

jannisvisser:
Sure, but then we need other checks and balances to catch something like this. That would be the feature test scenario's. 

BTW: In our new way-of-working we would have defined during refinement this scenario as one that needed testing. So there is a plan/progress.

PR #2227, opened 2021-06-30
test: update test scenarios import AB#8317

jannisvisser:


jannisvisser:
@elwinschmitz I made this update to Import test scenarios. Actually a good exercise to do this before building everything (this way you can see how I had in my mind roughly). Obviously, this needs updating with exact functionality/naming when done.

elwinschmitz:
@jannisvisser That is the exact definition of TDD ([Test Driven Development](https://en.wikipedia.org/wiki/Test-driven_development)) :D

PR #2247, opened 2021-07-07
feat: adapt metrics endpoint AB#8518

jannisvisser:


jannisvisser:
@elwinschmitz this is a first version of the new endpoint
- it counts the number of not-null-dates in the specific column (unlike filtering directly on status before)
- you can filter further by whether the date-column is in the right month and/or year and/or installment-period
- I did not change the attributes of PaMetrics for now, so that the front-end would keep working, and I wouldn't commit a broken state
- still the attributes do have to change of course, so not merging this for now. 
- But please continue with this, maybe by merging it into your own local branch and then continuing with it.

PR #2266, opened 2021-07-14
feat: 1st version UX copy for column + row headers metrics table AB#8652

jannisvisser:


elwinschmitz (approved):
Good starting point! 👍

PR #2319, opened 2021-08-24
AB#8669  Export metrics

arsforza:


interfaces/HO-Portal/src/app/program/metrics-states/metrics-states.component.html
```diff
@@ -1,10 +1,19 @@
 
 <table class="metrics-states--table">
   <caption>
-    <app-refresh-data
-      [lastUpdated]="lastUpdated"
-      (refresh)="update()"
-    ></app-refresh-data>
+    <ion-row>
+      <ion-col>
+        <ion-button float-left id="btn-export-csv" (click)="exportCSV()" size="small" >
```
elwinschmitz, line 6:
```suggestion
        <ion-button
          (click)="exportCSV()"
          size="small"
          class="ion-float-left"
        >
```

We try to use the 'regular' CSS class-names, instead of the (deprecated) attributes for alignment/layout.
Also, if an ID isn't necessary, we can leave it out.

And the HTML isn't being formatted automatically, but the 'rule of thumb' is "when there are more then 2 attributes, align them as 1-indent for every attribute on its own line".

(But there are places where this rule isn't applied. ;) )

elwinschmitz:
I think you have to take a look at you VSCode-Prettier extension config or something, to make sure you can resolve the linting-issues before pushing to GitHub etc.

interfaces/HO-Portal/src/app/program/metrics-states/metrics-states.component.ts
```diff
@@ -253,4 +254,24 @@ export class MetricsStatesComponent implements OnChanges {
...
+    const rows = [
+      ['', this.paStates.map(({ label }) => label)],
+      [rowHeader('payment'), this.paStates.map(({ forPayment }) => forPayment)],
+      [rowHeader('payment-from-start'), this.paStates.map(({ forPaymentFromStart }) => forPaymentFromStart)],
+      [rowHeader('calendar-month'), this.paStates.map(({ forMonth }) => forMonth)],
+      [rowHeader('month-from-start'), this.paStates.map(({ forMonthFromStart }) => forMonthFromStart)],
+      [rowHeader('to-date'), this.paStates.map(({ toDate }) => toDate)]
+    ]
+
+    const csvContents = rows.map((e) => e.join(",")).join("\n");
```
elwinschmitz, line 270:
```suggestion
    const csvContents = rows.map((e) => e.join(',')).join('\n');
```

interfaces/HO-Portal/src/app/program/metrics-states/metrics-states.component.ts
```diff
@@ -253,4 +254,24 @@ export class MetricsStatesComponent implements OnChanges {
...
+
+    saveAs(
+      new Blob([csvContents], { type: 'text/csv' }),
+      `program-${this.program.id}-metrics-${new Date().toISOString().substr(0, 10)}.csv`
+    );
+  }
+}
```
elwinschmitz, line 277:
```suggestion
}

```

elwinschmitz (approved):

PR #2321, opened 2021-08-25
fix: status callback performance AB#9202

jannisvisser:


jannisvisser:
@RubenGeo this is then supposed to be the fix, to merge in master. It is copied from [this commit](https://github.com/global-121/121-platform/commit/a93196e9790ae9506176a0ca23e303e376ce4203) on 'refactor-halfway', which also involves the extension of Twilio-mock.

If you agree that (1) this solution will work / not break anything, and (2) that this is for now the best we can do and hope that this solves it, then we should merge this and get it deployed to production before Tuesday.

PR #2322, opened 2021-08-25
Add documentation for recommended tools/extensions

elwinschmitz:


elwinschmitz:
@arsforza After this is merged to master you should be able to install the VS Code extensions we've talked about.

PR #2340, opened 2021-09-06
AB#9205 Adapt Interface

sowjanyaBysani:


interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.html
```diff
@@ -35,6 +38,16 @@ <h3>{{ 'page.program.program-payout.make-export'|translate }}</h3>
       [paymentInstallment]="exportInstallmentId"
       [disabled]="!exportInstallmentAvailable"
     ></app-export-list>
+    <ng-container>
```
elwinschmitz, line 41: The `ng-container` isn't really required here, I think, if there is no real condition-attribute on it.

(I'm not sure why it is used inside the `<option>`-elements above... :? )

interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.ts
```diff
@@ -232,6 +232,10 @@ export class ProgramPayoutComponent implements OnInit {
       this.exportInstallmentAvailable = false;
       return;
     }
+    if (isNaN(this.exportInstallmentId)) {
+      this.exportInstallmentAvailable = true;
+
```
elwinschmitz, line 237: I think we need a `return` here, so we don't do the request to the endpoint with an 'invalid' `exportInstallmentId`.

interfaces/HO-Portal/src/app/program/program-payout/program-payout.component.html
```diff
@@ -35,6 +38,16 @@ <h3>{{ 'page.program.program-payout.make-export'|translate }}</h3>
...
+      <app-export-list *ngIf="programId"
+      [programId]="programId"
+      [exportType]="enumExportType.payment"
+      [minInstallment]=10
```
elwinschmitz, line 45: Is this number based on something?
I think hard-coding `[minInstallement]="0"` could be 'safe enough', but the `maxInstallment` should be related to "the most recent succesfull installment"(so `lastInstallmentId` or similar) or "the maximun installment possible for this program"(so `program.distributionDuration`)

elwinschmitz (changes requested):

interfaces/HO-Portal/src/app/program/export-list/export-list.component.ts
```diff
@@ -20,6 +20,10 @@ export class ExportListComponent implements OnChanges {
   public exportType: ExportType;
   @Input()
   public paymentInstallment: number;
+  @Input()
+    public minInstallment: number;
```
elwinschmitz, line 24: This indentation should be auto-fixed by Prettier, didn't it run?

jannisvisser:
@sowjanyaBysani I've committed a changed endpoint to this endpoint
- please pull this code locally
- I've left it for you to integrate the code further. I did already change the calling of the endpoint in HO-portal's programs-service-api-service.ts
- but it's up to you to call the exportList function in there correctly from the places where it is used
- the gist is that you have to pass a minInstallmentId and maxInstallmentId now instead of just one installmentId
- in the cases where you're still exporting one installment only, both values can be the same

PR #2349, opened 2021-09-07
AB#9267 Update readme

arsforza:


elwinschmitz:
@arsforza I don't know if you've come across the issue where you needed to rebuild `node-sass`?
But maybe you can also include that in the list of "known issues/fixes/workarounds":

So it could help to run `npm rebuild node-sass` (from the directory of an interface), to fix some `node-sass`-errors.
(But I haven't been able to reproduce the situation on my machine)

Also, I would recommend not to _repeat_ commands to do something in that README-section, but rather link to the related steps in the "Getting Started"-section. That way we only have to keep 1 place up-to-date.

PR #2363, opened 2021-09-13
AB#9353 Add info button/tooltip (i) to component

sowjanyaBysani:


interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -143,8 +143,9 @@
         "export-button": {
           "label": "Export as CSV"
         },
-        "totals": {
-          "total-helped-to-date": "Total number of individual PA helped to date"
+        "totals":{
+          "total-helped-to-date": "Total number of individual PA helped to date",
+          "pas-that-received-one-or-more-payments":"PA's that received one or more payments"
```
jannisvisser, line 148: @sowjanyaBysani I don't think it's very elegant to have the key the same as the value. I would name it something more generic like "tooltip". Please look at previous examples of how it's done, something like:

"total-helped-to-date": {
    "label": ....,
    "tooltip": ....,
}

PR #2364, opened 2021-09-13
AB#9399 Fix for double registrations in HO-portal & missing note

RubenGeo:


services/121-service/src/registration/registrations.service.ts
```diff
@@ -443,6 +443,8 @@ export class RegistrationsService {
     let q = await this.registrationRepository
       .createQueryBuilder('registration')
       .select('registration.id', 'id')
+      .distinctOn(['registration.id'])
```
jannisvisser, line 446: @RubenGeo I seem to remember that using DISTINCT in sql-queries is a lot heavier then using GROUP BY, while receiving the same result. But if it works fine then ok. (Did you test this with the big database?)

RubenGeo: With the mocked LVV dataset loading still takes <1 second

PR #2385, opened 2021-09-20
AB#8473 payment information does not refresh when entering screen

sowjanyaBysani:


interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -213,7 +213,10 @@ export class ProgramPeopleAffectedComponent implements OnInit {
     this.locale = environment.defaultLocale;
     this.router.events.subscribe((event) => {
       if (event instanceof NavigationEnd) {
-        this.loadData();
+        // this.loadData();
```
jannisvisser, line 216: @sowjanyaBysani you forgot to remove the commented line

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -207,8 +208,17 @@ export class ProgramPeopleAffectedComponent implements OnInit {
...
+    private router: Router,
   ) {
     this.locale = environment.defaultLocale;
+    this.router.events.subscribe((event) => {
+      if (event instanceof NavigationEnd) {
+        // this.loadData();
+        this.pubSub.subscribe(PubSubEvent.dataRegistrationChanged, () => {
+          this.refreshData();
+        });
+      }
```
elwinschmitz, line 220:
```suggestion
      if (event instanceof NavigationEnd) {
        this.refreshData();
      }
```
This should be enough ⬆️ here.

Using the `PubSub`-service should work the other way around and not how you've done it here.

My intention with the sub-task was to use this:
https://github.com/global-121/121-platform/pull/2385/files#diff-d7ebb2f882957eb11ff0500b8aec39f0b5401397e9a047816ad6a795ccbdbdecR479-R482
Also in the other components. (new-payment etc.)

That way, when that event is published on the PubSub-service, they could refresh/reload themselves.

PR #2389, opened 2021-09-21
Feat.cancel vouchers

jannisvisser:


interfaces/HO-Portal/src/app/shared/input-prompt/input-prompt.component.html
```diff
@@ -17,10 +17,10 @@
   color="light"
   class="ion-padding"
 >
-  <h2 
+  <h5 
     *ngIf="subHeader" 
     [innerHTML]="subHeader">
-  </h2>
+  </h5>
```
jannisvisser, line 23: @elwinschmitz how do you feel about this commit? It's been a long time nuisance for me that the subheader of the inputPrompt (which is used often for extra explanation, as the 'message' is already reserved for the 'last executed timestamp') is so big.. It looks horrible. I know this component is used in more cases, but tbh I think it should be OK in all cases..

jannisvisser:
Closing this because we're not doing this for now. We want to keep the branch though.

PR #2390, opened 2021-09-21
AB#6669-sanitize-change-fsp-phone

RubenGeo:


services/121-service/src/registration/registrations.service.ts
```diff
@@ -864,10 +864,13 @@ export class RegistrationsService {
     const updatedRegistration = await this.addFsp(referenceId, newFsp.id);
 
     // Add new attributes
-    updatedRegistration.fsp.attributes.forEach(async attribute => {
-      updatedRegistration.customData[attribute.name] =
-        newFspAttributes[attribute.name];
-    });
+    for (const attribute of updatedRegistration.fsp.attributes) {
+      await this.addCustomData(
```
jannisvisser, line 868: @RubenGeo addCustomData() only does something if the key is not already present in customData. Does that go right always? I guess the simple solution could be to remove that weird rule from addCcustomData(), as it never matters?

PR #2391, opened 2021-09-22
AB#6673 Change FSP of PA in HO Portal

arsforza:


arsforza:
@jannisvisser you can check the code now

jannisvisser:
@arsforza I've committed my changes, which bring it a bit further. Still quite some more needed though:
- as discussed, show the initial FSP when opening popup
- the attributes should actually NOT show, when you're selecting the same FSP as I already have (so when I first switch from whatsapp to no-whatsapp > and then switch back > the whatsappnumber field should not show)
- now I used the name of the attribute as label, but that's not a readable string, and there's not currently a better label-property in the fsp-attribute itself
- and then of course the stuff you didn't get to yet, like that the 'update' button only enables once all the required fsp-attributes are filled in

interfaces/HO-Portal/src/app/components/update-property-item/update-property-item.component.html
```diff
@@ -34,6 +34,44 @@
           pattern="[0-9]*"
         ></ion-input>
       </ng-container>
+      <ng-container *ngSwitchCase="'fsp'">
```
elwinschmitz, line 37: I would suggest actually to not include this whole FSP-specific block in this generic "update-property-item"-component.

But instead create a new component for it, or directly in: `edit-person-affected-popup.component.html`
https://github.com/global-121/121-platform/pull/2391/files#diff-7f2096eb325da4786303a7ecf9a059f8e3a5b49f65fcdca62aed7f32251356d4R61-R67

interfaces/HO-Portal/src/app/models/q-and-a.models.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class Answer {
+  code: string;
+  value: string;
+  label: string;
+}
+
+export class AnswerSet {
+  [code: string]: Answer;
+}
```
elwinschmitz, line 33: If all these aren't used, I'd not include them.
As they will get 'out of date' and only add confusion later.

interfaces/HO-Portal/src/app/program/edit-person-affected-popup/edit-person-affected-popup.component.html
```diff
@@ -58,6 +58,13 @@
         [inProgress]="inProgress?.whatsappPhoneNumber || false"
         class="ion-margin-vertical"
       ></app-update-property-item>
+      <app-update-property-item
```
elwinschmitz, line 61: So I'd suggest to put the whole FSP-specific block here.
That also reduces the handling of the `fspList`-object a bit.

interfaces/HO-Portal/src/app/program/edit-person-affected-popup/edit-person-affected-popup.component.ts
```diff
@@ -14,11 +15,18 @@ export class EditPersonAffectedPopupComponent implements OnInit {
...
 
+  @Input()
+  public programId: number;
+
   public inProgress: any = {};
 
   public noteModel: string;
   public noteLastUpdate: string;
 
+  public fspList: any[];
```
elwinschmitz, line 26:
```suggestion
  public fspList: Fsp[];
```

interfaces/HO-Portal/src/app/models/q-and-a.models.ts
```diff
@@ -0,0 +1,33 @@
+export enum AnswerType {
+  // Translate the types used in the API to internal, proper types:
+  Number = 'numeric',
+  Text = 'text',
+  Date = 'date',
+  Enum = 'dropdown',
+  phoneNumber = 'tel',
+  email = 'email',
+}
```
elwinschmitz, line 9: You could also include this definition in the `fsp.model.ts`-file, because it is only related to that.

The whole "Q-and-A"-concept doesn't really exist in the HO-Portal.

interfaces/HO-Portal/src/app/components/update-property-item/update-property-item.component.html
```diff
@@ -34,6 +34,44 @@
...
+        >
+          <ion-select-option
+            *ngFor="let fspItem of fspList"
+            value="{{ fspItem.fsp }}">
+              {{ fspItem.fsp }}
+          </ion-select-option>
+        </ion-select>
+        <ion-item *ngIf="attributeDifference.length > 0">
+          <ion-text color="warning">
+            <p>Selecting this FSP will delete the following attributes related to the old FSP:</p>
```
elwinschmitz, line 55: Make sure to make these texts translatable...

elwinschmitz (changes requested):

interfaces/HO-Portal/src/app/program/edit-person-affected-popup/edit-person-affected-popup.component.ts
```diff
@@ -136,4 +145,22 @@ export class EditPersonAffectedPopupComponent implements OnInit {
   public closeModal() {
     this.modalController.dismiss();
   }
+
+  private getFspList() {
+    this.fspList = [];
+
+    if (this.programId) {
```
elwinschmitz, line 152: If you turn this around, into an "early return", it would save 1 level of indentation for the rest of the logic and that would help for readability/understandability etc..

It will also highlight that "nothing will happen without a ProgramId".

So:
```
if (!this.programId) {
  return;
}
```

Or you can move the check outside of the `getFspList`-method and do a check at the very start in the `ngOnInit`-method..

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -304,7 +304,8 @@
             "whatsappPhoneNumber": {
               "label": "WhatsApp number",
               "explanation": "This value is required for a 'WhatsApp-enabled' FSP only."
-            }
+            },
+            "fspChangeWarning": "Selecting this FSP will delete the following attributes related to the old FSP:"
```
elwinschmitz, line 308: It's not super important, but my initial intention with this structure was that all child-elements of the `...edit-person-affected-popup.properties`-object are only "field/column names".
So this message is outside of that scope, so it could get "its own branch" below `edit-person-affected-popup`, maybe.

interfaces/HO-Portal/src/app/components/update-fsp/update-fsp.component.html
```diff
@@ -0,0 +1,74 @@
...
+            {{ attribute.name }}
+          </ion-label>
+          <ion-input
+          ngDefaultControl
+          [name]="attribute.name"
+          [value]="attribute.value"
+          [placeholder]="attribute.placeholder?.en"
+          (ionChange)="onAttributeChange($event)"
+          ></ion-input>
+        </ion-item>
```
elwinschmitz, line 47: Wouldn't it be possible to use the 'generic' "`<app-update-property-item>`"-component here?

That way all the same rules apply to all the fields...

elwinschmitz (approved):
Some more improvements/tweaks can be done in a follow-up-PR, so this can be merged...

interfaces/HO-Portal/src/app/components/update-fsp/update-fsp.component.ts
```diff
@@ -0,0 +1,160 @@
...
+    const alert = await this.alertController.create({
+      backdropDismiss: false,
+      message: resultMessage,
+      buttons: [
+        {
+          text: this.translate.instant('common.ok'),
+          handler: () => {
+            alert.dismiss(true);
+            if (refresh) {
+              window.location.reload();
```
jannisvisser, line 150: @elwinschmitz we could not come up with a better way to refresh. The pubsub-thing didn't seem to work (and we didn't know how it sould work).

The ideal situation is: that the popup stays open but gets refreshed, but that might be hard, so we went for the easy / second-best of refreshing the whole page.

interfaces/HO-Portal/src/app/components/update-fsp/update-fsp.component.ts
```diff
@@ -0,0 +1,160 @@
...
+          this.actionResult(
+            this.translate.instant('common.update-success'),
+            true,
+          );
+        },
+        (error) => {
+          this.inProgress = false;
+          console.log('error: ', error);
+          if (error && error.error) {
+            this.actionResult(error.error.message);
```
jannisvisser, line 86: @elwinschmitz it was more difficult here to use the same formatErrors() function you've used elsewhere, because it's potentially about multiple attributes, etc. It could be done of course, but this was GEFN we judged.

interfaces/HO-Portal/src/app/components/update-fsp/update-fsp.component.ts
elwinschmitz: The idea is to _publish_ an event at this point, telling everybody else that "something has changed!"

So all the other components that _want_ to respond to that event can do/decide that themselves.

And in this case the "registrationDataChanged"-event is also applicable. (So very similar to a note being updated)

interfaces/HO-Portal/src/app/components/update-fsp/update-fsp.component.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('UpdateFspComponent', () => {
+  let component: UpdateFspComponent;
+  let fixture: ComponentFixture<UpdateFspComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [UpdateFspComponent],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
+      imports: [FormsModule, TranslateModule.forRoot()],
+    }).compileComponents();
```
elwinschmitz, line 16: Above this line, the `providers`-array should be provided to reflect the contents of the `constructor`, from: https://github.com/global-121/121-platform/pull/2391/files#diff-8ba64fef935c2f040240510612e0a92f9626c6614317e660f5a4b54d6bec5fbdR45-R49

elwinschmitz: So you shouldn't only try to include things the error-message asks for, but you should include what the **intent** of the component is.

It could be that the component requires a specific service only (and that service will 'under water' use the HttpClient), so for this test-scenario only that 1st level service is important, not every specific part down the 'dependency tree'.

PR #2392, opened 2021-09-22
AB#9767 Only return last 3 vouchers on incoming whatsapp

RubenGeo:


RubenGeo:
see comment i left in task @jannisvisser

PR #2394, opened 2021-09-24
AB#9449 Fixed as per comments

sowjanyaBysani:


elwinschmitz:
@sowjanyaBysani  Is this change replacing the changes from #2385 ?
If so, you can close that PR.

interfaces/HO-Portal/src/app/program/make-payment/make-payment.component.ts
```diff
@@ -38,7 +39,14 @@ export class MakePaymentComponent implements OnInit {
     private pastPaymentsService: PastPaymentsService,
     private translate: TranslateService,
     private alertController: AlertController,
-  ) {}
+    private router: Router,
+  ) {
+    this.router.events.subscribe((event) => {
+      if (event instanceof NavigationEnd) {
+        this.refresh();
```
elwinschmitz, line 46: This refresh-method is only there for some specific error-state, I think.
It is not a generic "refresh the component"-method.

I think it could be enough to re-run `this.ngOninit()` (but I haven't tested/tried that.. so maybe it can break something...)

elwinschmitz (approved):

PR #2433, opened 2021-10-13
Changed webhook to work for minor version AB#1075

RubenGeo:


tools/webhook.js
```diff
@@ -36,9 +36,9 @@ function deploy(target) {
  * Check whether to deploy/ignore a release
  * @param {string} target
  */
-function isPatchUpgrade(target) {
+function isMinorUpgrade(target) {
   const currentVersion = fs.readFileSync(`${process.env.GLOBAL_121_WEB_ROOT}/VERSION.txt`, { encoding: 'utf-8' });
-  const currentMinorVersion = currentVersion.replace(/v(\d+)\.(\d+)\.([\S\s]*)/, 'v$1.$2.');
+  const currentMinorVersion = currentVersion.replace(/v(\d+)\.(\d+)\.([\S\s]*)/, 'v$1.');
 
   return target.includes(currentMinorVersion);
```
elwinschmitz, line 43: @RubenGeo 

The name of this method and what is actually does isn't really in sync anymore, it seems. 

The logic was "if I want to deploy a new patch-version, at least the minor-version should match between existing-version and target-version.

I think we should not change this, but only ADD the option to also upgrade the minor-version, by verifying that the major-version matches and maybe that it's only a "+1-upgrade". (But that is maybe a bit too cautious)

PR #2484, opened 2021-11-02
fix: spanish template AB#10182

jannisvisser:


jannisvisser:
just for documentation: these changes have been applied via API on nrlc-staging and nlrc-pv, and successfully tested on staging.

PR #2521, opened 2021-11-16
Feat.export question labels

jannisvisser:


jannisvisser:
@RubenGeo I went in the end only for labels for dropdown answers, not for questions. Because the questions can be really long and contain HTML. And for the questions it's better doable to keep the short name understandable I thought.

jannisvisser:
integrated in #2519

PR #2538, opened 2021-11-18
Update seed script to final version

diderikvw:
DevOps #10371: update ZOA seed script with latest info from ZOA.


RubenGeo:
These changes are now in this PR #2544

PR #2556, opened 2021-11-24
Fix.dorcas question1

jannisvisser:


jannisvisser:
@elwinschmitz FYI: I've added a 'sort by id' step in all buildQuestions methods, as this was needed. Otherwise it sorts them by the order they happen to be in the table in the database (which can for example change if you update one question).

PR #2585, opened 2021-12-01
import fsp reconciliation AB#10289

jannisvisser:


services/121-service/src/registration/services/bulk-import.service.ts
```diff
@@ -380,4 +389,100 @@ export class BulkImportService {
...
+
+      importResponseRecord.importStatus = ImportStatus.imported;
+      importResponseRecords.push(importResponseRecord);
+      countImported += 1;
+    }
+
+    this.actionService.saveAction(
+      userId,
+      programId,
+      AdditionalActionType.importPeopleAffected,
```
RubenGeo, line 441: `AdditionalActionType.importPeopleAffected` should be a new value?

services/121-service/src/registration/services/bulk-import.service.ts
```diff
@@ -380,4 +389,100 @@ export class BulkImportService {
...
+        relations: ['fsp'],
+      });
+      if (!registration) {
+        importResponseRecord.importStatus = ImportStatus.unmatched;
+        importResponseRecords.push(importResponseRecord);
+        countNotFound += 1;
+        continue;
+      }
+
+      const paTransactionResult = new PaTransactionResultDto();
```
RubenGeo, line 420: What happens if you import the same file twice? You get multiple of the same transactions with later timestamps in the database?

jannisvisser: Yes, but that's no problem right? That's how our transactions system is set up, you can send as many status updates as you want, and always the latest one is taken. Right?

services/121-service/src/registration/services/bulk-import.service.ts
jannisvisser: yes.

jannisvisser: done

services/121-service/src/registration/services/bulk-import.service.ts
RubenGeo: But then the transaction date changes right. Do we not use that somewhere? Don't we want to differentiate between PAs that received their payment during an earlier import?

jannisvisser: Sure. I've made task to not update transaction if status stays the same.

jannisvisser:
Closing this: see https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%2064?workitem=10289

PR #2605, opened 2021-12-13
AB#10292 Improve look&feel of "change password" page

arsforza:


interfaces/HO-Portal/src/app/user/user.page.html
```diff
@@ -39,21 +43,50 @@ <h2>{{ 'page.user.change-password.title'|translate }}</h2>
...
-                  [minlength]="8"
-                  name="new-password"
-                  [(ngModel)]="newPassword"
-                  ngDefaultControl
-                  class="ion-margin-vertical">
-                </ion-input>
+                <div class="input-item--content">
+                  <ion-input
+                    type="password"
+                    autocomlete="off"
```
elwinschmitz, line 49: Why change this?
It can trigger users' password-managers to insert/prompt for the new password.

interfaces/HO-Portal/src/app/user/user.page.html
```diff
@@ -39,21 +43,50 @@ <h2>{{ 'page.user.change-password.title'|translate }}</h2>
...
+                      <ion-note color="danger" *ngIf="showInvalidPasswordMessage()">{{ 'page.user.error-messages.password-characters'|translate }}</ion-note>
+                  </div>
+                </div>
+              </div>
+              <div class="input-item ion-margin-bottom ion-padding-bottom">
+                <label color="medium" class="input-item--label auth-label">{{ 'page.user.change-password.confirm-password'|translate }}</label>
+                <div class="input-item--content">
+                  <ion-input
+                    type="password"
+                    autocomlete="off"
```
elwinschmitz, line 72: Here it makes sense, because we don't want that users' browsers/extensions will suggest/fill-in the old password here or something..

interfaces/HO-Portal/src/app/user/user.page.html
```diff
@@ -39,21 +43,50 @@ <h2>{{ 'page.user.change-password.title'|translate }}</h2>
...
+                    autocapitalize="off"
+                    clearInput="true"
+                    clearOnEdit="false"
+                    required
+                    [minlength]="8"
```
elwinschmitz, line 77: As we use this 'magic' value in at least 2 places (but also maybe inside the validation-messages?), we could use a single constant/variable here. So we only have to change it in 1 place and can't overlook anything that should be connected to it.

interfaces/HO-Portal/src/app/user/user.page.ts
```diff
@@ -31,4 +32,29 @@ export class UserPage implements OnInit {
...
   }
+
+  public checkConfirmPasswords() {
+    this.confirmPassword !== this.newPassword
+      ? this.newPasswordForm.form.setErrors({ differentPasswords: true })
+      : this.newPasswordForm.form.setErrors(null);
+  }
+
+  private shortPassword(): boolean {
+    return this.newPassword.length < 8;
```
elwinschmitz, line 43: Here the same variable/constant should be used. Otherwise things will 'get out of sync' when only 1 of them is changed.

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -457,7 +457,13 @@
       "change-password": {
         "title": "Change password",
         "new-password": "New password",
+        "confirm-password": "Confirm password",
         "success": "Password changed"
+      },
+      "error-messages": {
+        "password-length": "Password should be more than 8 charachters.",
```
elwinschmitz, line 464: We should use a variable inside the translations as well.
I think we do that in the PA-app already.

interfaces/HO-Portal/src/global.scss
```diff
@@ -97,6 +97,15 @@ ion-button {
       --padding-end: 8px;
     }
   }
+
+  .validation-message {
+    min-height: 36px;
```
elwinschmitz, line 102: To avoid using a 'magical' number/height here, we could make the text 'invisible' instead.
So the text will always be rendered AND occupy space/height, regardless of its size/length(taking translations into account).

So the default-styles will be: 
```css
.validation-message {
  // By default 'invisible':
  visibility: hidden;

  &.is-triggered {
    visibility: visible;
  }
}
```

interfaces/HO-Portal/src/global.scss
```diff
@@ -97,6 +97,15 @@ ion-button {
...
+
+    ion-note {
+      font-weight: bold;
+      font-size: 12px;
```
elwinschmitz, line 106: Would using a `<small>`+`<strong>`-element also work visually?

So we don't need the specific CSS here.

interfaces/HO-Portal/src/global.scss
```diff
@@ -107,3 +116,39 @@ ion-button {
...
+    padding: 0;
+    width: 48px;
+  }
+
+  h1 {
+    margin-top: 10px;
+    margin-left: 4px;
+    letter-spacing: 3px;
+  }
+}
```
elwinschmitz, line 138: We could make separate component out of this even maybe... (but maybe not now per se.. ;) )

elwinschmitz (changes requested):
Looks good, only some small 'future proofing' needed.

interfaces/HO-Portal/src/app/user/user.page.html
arsforza: It does it anyway

image

But I can change it back

interfaces/HO-Portal/src/app/user/user.page.ts
arsforza: Variable added

interfaces/HO-Portal/src/app/user/user.page.html
arsforza: Variable added

interfaces/HO-Portal/src/assets/i18n/en.json
arsforza: Done

elwinschmitz (approved):

PR #2632, opened 2021-12-22
AB#9967 Update login validation behavior

arsforza:


arsforza:
@elwinschmitz
To check with SPD team before merging:
- added failed login validation message between title and login form
- removed green from login fields border
- on login fail, both fields have red border
- validation message for invalid format under both fields

PR #2640, opened 2021-12-28
AB#11547 Update Look & Feel of 121 Portal

arsforza:


interfaces/HO-Portal/src/app/components/header/header.component.html
```diff
@@ -1,62 +1,37 @@
...
+      <ion-col size="1">
+        <ion-button
+          [title]="startButtonRouterLink === '/home' ? ('header.home'|translate) : ('header.help'|translate)"
+          [attr.aria-label]="startButtonRouterLink === '/home' ? ('header.home'|translate) : ('header.help'|translate)"
+          [routerLink]="startButtonRouterLink"
+          fill="clear"
+          size="small"
+          color="medium"
+        >
+          <ion-icon [name]="startButtonIcon" slot="icon-only"></ion-icon>
```
elwinschmitz, line 13: We should not use conditional values for these properties, but use basically 2 buttons and show only 1 based on some condition.

So 2 chunks of HTML to show the "Home"-button and the "Help"-button.
And then some logic elsewhere to set `showHome` and/or `showHelp`.

interfaces/HO-Portal/src/app/components/header/header.component.html
```diff
@@ -1,62 +1,37 @@
...
+        </ion-button>
+      </ion-col>
+      <ion-col size="4">
+        <h1 class="ion-no-margin">
+          {{ title }}
+        </h1>
+        <ion-note *ngIf="programPage">
```
elwinschmitz, line 20: I would not put 'implicit logic' in these templates, just only 'display logic'.

So only check whether `programTitlePortal` is set/has any value.
Then the whole 'connection' or knowledge about "what is a program-page?" can be left out of this component.

interfaces/HO-Portal/src/app/program/disable-registration/disable-registration.component.html
```diff
@@ -1,10 +1,6 @@
...
-  class="ion-align-items-center ion-justify-content-end"
->
-
-<ion-item *ngIf="program">
+<ion-row *ngIf="program" class="ion-align-items-center">
   <ion-label>{{ 'page.program.phases.registrationValidation.toggleText-enable-registrations'|translate }} </ion-label>
-  <ion-toggle  [ngModel]=publishedStatus (click)="updateRegistrationStatus()"></ion-toggle>
-</ion-item>
-</div>
+  <ion-toggle  [ngModel]=publishedStatus (click)="updateRegistrationStatus()" color="secondary"></ion-toggle>
```
elwinschmitz, line 3:
```suggestion
  <ion-toggle
    [(ngModel)]="publishedStatus"
    (click)="updateRegistrationStatus()"
    color="secondary"
  ></ion-toggle>
```

I'm not sure if the `ngModel`-property would need the 2-way-data-binding, but I think it almost always is for these type of properties..

interfaces/HO-Portal/src/app/program/phase-navigation/phase-navigation.component.html
```diff
@@ -1,21 +1,21 @@
 <span
   *ngFor="let phase of programPhases"
-  class="phase-navigation"
+  class="active-page-button"
```
elwinschmitz, line 3: The name of this (CSS-only)components is a bit confusing/misleading.

It is a "page-button" that _can_ have the state "active", but using the term "active" in its name is a bit confusing.

interfaces/HO-Portal/src/global.scss
```diff
@@ -29,15 +29,30 @@
...
 
 .horizontal-scroll-container {
   overflow-x: auto;
   -webkit-overflow-scrolling: touch;
 }
 
+.top-header {
+  border-bottom: 1px solid #c2c2c2;
+  &::after {
+    background-image: none;
```
elwinschmitz, line 52: Is this to override some Ionic-default styling?

Maybe better to put it on the same selector as the Ionic-styling then.
So `ion-header::after`??

interfaces/HO-Portal/src/global.scss
```diff
@@ -152,3 +167,18 @@ ion-icon {
   margin-left: auto;
   margin-right: auto;
 }
+.active-page-button {
+  --phase-navigation--indicator-size: 2px;
```
elwinschmitz, line 171: If moving this CSS-component into the global scope, it is also better to rename its variables/custom-properties, I think.

But why would it be necessary to move away from the specific component anyway?

elwinschmitz (changes requested):

PR #2660, opened 2022-01-04
Feat.stop canceling

jannisvisser:


jannisvisser:
Already merged this - after short review with Diderik - because we wanted to include this asap in deploy to production. However, this still warrants a more detailed review by a developer.

PR #2665, opened 2022-01-05
feat: show amount in PA-table AB#11799

jannisvisser:


interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.html
```diff
@@ -260,23 +255,30 @@ <h2>{{ 'page.program.program-people-affected.section-title'|translate }}</h2>
            }"
           [attr.title]="'page.program.program-people-affected.transaction.show-status'|translate"
           [attr.aria-label]="'page.program.program-people-affected.transaction.show-status'|translate"
+          [class.payment-cell-no-popup]="!hasVoucherSupport(row.fsp) && !hasError(row, column.paymentIndex)"
```
elwinschmitz, line 258: This will also work, but there is already an `[ngClass]`-property on this element, so better to add it there.

The pattern there is:  
`'class-name': conditionToAddClassNameWhenTrue`

So, add this: `'payment-cell-no-popup': !hasVoucherSupport(row.fsp) && !hasError(row, column.paymentIndex),`
to line 254 above.

@jannisvisser

elwinschmitz (changes requested):

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.html
jannisvisser: @elwinschmitz feel free to take this item over and process these changes yourself, if you want?

elwinschmitz (approved):
We, from WC-duck... ;)

PR #2685, opened 2022-01-12
Feat roles permissions

jannisvisser:


services/121-service/src/user/user-role.entity.ts
```diff
@@ -1,13 +1,12 @@
...
 import { ProgramAidworkerAssignmentEntity } from '../programs/program-aidworker.entity';
 import { Base121Entity } from '../base.entity';
+import { PermissionEntity } from './permissions.entity';
 
 @Entity('user_role')
 export class UserRoleEntity extends Base121Entity {
-  @Column()
-  public role: UserRole;
+  @Column({ unique: true })
+  public role: string;
```
elwinschmitz, line 9: Shouldn't this become `DefaultUserRole` instead of accepting any string? @RubenGeo

RubenGeo: No, it should also allow custom user defined roles

elwinschmitz (approved):

PR #2713, opened 2022-01-26
Do not merge roles but replace them

RubenGeo:


RubenGeo:
I think we need this to be able to change the Roles on out production servers @jannisvisser. Can you double check if this does not break something

PR #2747, opened 2022-02-08
Feat: Token in cookies in interfaces

PeterSmallenbroek:


interfaces/HO-Portal/src/app/models/user.model.ts
```diff
@@ -1,11 +1,9 @@
...
-
 export class UserModel {
   token: string;
 }
 
 export class User {
   token?: string;
   username: string;
-  permissions: Permission[];
+  permissions: string[];
```
elwinschmitz, line 8: The type `Permission` in the interface(s) is not the same as the `Permission`-object in the back-end;
So we should keep this as it was.

interfaces/HO-Portal/src/app/services/jwt.service.ts
```diff
@@ -5,19 +5,21 @@ import { JwtHelperService } from '@auth0/angular-jwt';
   providedIn: 'root',
 })
 export class JwtService {
-  private tokenKey = 'jwt-HO';
+  private tokenKey = 'logged-in-user-HO';
```
elwinschmitz, line 8: We might not even need this anymore...
As "Users" and their cookies would be valid for both interfaces at the same time...

But maybe this would need some extra testing...

PR #2781, opened 2022-02-25
Feat: Make custom attributes clickable

PeterSmallenbroek:


features/HO-Portal/Edit_Info_Person_Affected.feature
```diff
@@ -22,7 +22,7 @@ Feature: Edit information on Person Affected
     Then a popup opens
     And in the title the ID-number of the Person Affected is mentioned
     And an input-field for the "paymentAmountMultiplier" and "phoneNumber" is shown
-    And an input-field for each Custom Attribute (such as "partnerOrganization") is shown
+    And an input-field for each Custom Attribute with type 'string' (such as "namePartnerOrganization") is shown
```
arsforza, line 25:
```suggestion
    And an input-field for each Custom Attribute with type other than 'boolean' (such as "namePartnerOrganization") is shown
```

@PeterSmallenbroek I'm assuming custom attributes can be "string", "number", or "phoneNumber"

PR #2818, opened 2022-03-23
fix: apply also to AW-app login AB#13081

jannisvisser:


jannisvisser:
@RubenGeo the exact same problem also occurred in AW-app local sw-mode. There logging in worked fine, but afterwards when using e.g. 'download data' you would get a 401 error because of the missing cookie.

PR #2799, opened 2022-03-10
feat: PA app batch registrations AB#12149

arsforza:

PR #2863, opened 2022-04-06
Revert "feat: PA app batch registrations AB#12149"

jannisvisser:
Reverts global-121/121-platform#2799


jannisvisser:
@elwinschmitz this seems to be allright?

PR #2983, opened 2022-06-07
Feat.automatic multiplier

jannisvisser:


services/121-service/src/registration/registrations.service.ts
```diff
@@ -424,6 +428,11 @@ export class RegistrationsService {
       RegistrationStatusEnum.registered,
     );
     this.inclusionScoreService.calculateInclusionScore(referenceId);
+    this.inclusionScoreService.calculatePaymentAmountMultiplier(
```
RubenGeo, line 431: @jannisvisser I think this one is not needed. I commented this one out locally and it still worked. Are there scenarios where you need this?

jannisvisser: @RubenGeo You mean because at registration, this is already done via the setAttribute method?

PR #2990, opened 2022-06-08
Fix.refactor questions pa table

jannisvisser:


interfaces/HO-Portal/src/app/models/person.model.ts
```diff
@@ -26,7 +26,7 @@ export class Person {
   whatsappPhoneNumber?: string;
   paymentAmountMultiplier?: number;
   preferredLanguage?: LanguageEnum;
-  customAttributes?: object;
+  paTableAttributes?: object;
```
elwinschmitz, line 29: I'd say that "customAttributes" wasn't the best name, but "paTableAttribute" is at least worst. :P

But it depends on the porpose of this property what the "_best_" name would be.

My main comment/concern is that its name should describe the 'thing' in a slightly more generic way, because it will maybe be used in other ways as well, not only in a "table" for example..

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.html
```diff
@@ -337,6 +338,7 @@
       </ng-container>
     </ng-template>
   </ngx-datatable-column>
+  -->
```
elwinschmitz, line 341: This is left-over from something. :/ It should be removed, no?

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -309,7 +310,7 @@ export class ProgramPeopleAffectedComponent implements OnInit {
     const columnScoreWidth = 90;
     const columnPhoneNumberWidth = 130;
 
-    this.columnsAvailable = [
+    this.standardColumns = [
```
elwinschmitz, line 313: I would not use "standard". I think the better, more "programming"/code-related term would be "default" (if that is its purpose).

The term `columnsAvailable` was (I think) to create a collection of ALL possible columns, but that they could be 'turned off or on' based on some setting/value...
That logic hasn't changed right?

services/121-service/seed-data/program/program-pilot-nl-2.json
```diff
@@ -105,13 +105,7 @@
...
       },
       "name": "namePartnerOrganization",
-      "type": "string",
-      "export": [
-        "all-people-affected",
-        "included",
-        "selected-for-validation",
-        "payment"
-      ]
+      "type": "string"
```
elwinschmitz, line 108: This would become "`text`" instead of `string`, right?

See: https://github.com/global-121/121-platform/pull/2990/files#diff-04e9fd4a0e11700f1f08ed3c90e669d605859ef569543a9e95c93ecdc15dd59eR5

services/121-service/src/registration/enum/custom-data-attributes.ts
```diff
@@ -39,8 +39,9 @@ export enum GenericAttributes {
 }
 
 export class Attribute {
-  public attribute: string;
+  public name: string;
```
elwinschmitz, line 42: Doesn't this change require changes in the PA-app and/or AW-app as well?

Or where these only ever used in the Portal?

elwinschmitz (changes requested):

PR #3125, opened 2022-09-15
Seed multiple programs AB#15244

RubenGeo:


RubenGeo:
@elwinschmitz 

- I think this can already be merged in without breaking the current features. 
- So we can follow you incremental merging for large features. 
- This is sort of behind a switch ( seeding with demo-multiple)

elwinschmitz:
>     * I think this can already be merged in without breaking the current features.

As long as the existing seed-script(s) continue to work (only applicable for DRC though): 👌 
 
>     * So we can follow you incremental merging for large features.
>     * This is sort of behind a switch ( seeding with demo-multiple)

Indeed.

Only when some **existing** functionality is required to change, THEN we need to implement a feature-switch somewhere to 'fork' the code-path.

But this is just 'unused functionality', behind the 'secure' reset-token. To not 'accidentally' run this on a production instance, you could add in some `if (process.env.NODE_ENV === 'production') { return; }` in that `seed-program-demo-multiple.ts`-script.

RubenGeo:
"To not 'accidentally' run this on a production instance, you could add in some if (process.env.NODE_ENV === 'production') { return; } in that seed-program-demo-multiple.ts-script."

I think this is a bit overkill for this situation as we also do not have protection against seeding another 'wrong' program on a production server, which would have the same results. Shall I merge this? @elwinschmitz

PR #3141, opened 2022-09-21
Feat.multi select backend

jannisvisser:

PR #3140, opened 2022-09-20
Multi-select answertype

PeterSmallenbroek:

PR #3153, opened 2022-09-27
Multi select combined

jannisvisser:


jannisvisser:
This PR combines #3140 and #3141

PR #3156, opened 2022-09-28
Feat.multi select input validation

RubenGeo:


services/121-service/src/programs/program.entity.ts
```diff
@@ -154,4 +163,91 @@ export class ProgramEntity extends CascadeDeleteEntity {
...
+    const resultProgramCustomAttribute = await repo
+      .createQueryBuilder('program')
+      .leftJoin('program.programCustomAttributes', 'programCustomAttribute')
+      .where('program.id = :programId', { programId: this.id })
+      .andWhere('programCustomAttribute.name = :name', { name: name })
+      .select('"programCustomAttribute".type', 'type')
+      .getRawOne();
+
+    const repoInstance = getConnection().getRepository(InstanceEntity);
+    const monitoringQuestion = await repoInstance
```
jannisvisser, line 205: make relation with program @RubenGeo

services/121-service/src/registration/validator/registration-data-type.validator.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  private typeIsValid(
+    value: any,
+    type: string,
+    options: any[],
+    attribute: string,
+  ): boolean {
+    let isValid = false;
+    if (type === AnswerTypes.date) {
+      const parsedDate = Date.parse(value);
```
jannisvisser, line 77: @RubenGeo don't parse as date, but treat as string that needs to follow specified pattern

services/121-service/src/programs/program.entity.ts
RubenGeo: Instance is not related to program in the end

PR #3182, opened 2022-10-05
Prevent 500 error on data-storage endpoints AB#15150

RubenGeo:


elwinschmitz (approved):
This will result in error-status 403 instead of a 500 for "not logged in PAs" right?

PR #3200, opened 2022-10-11
AB#15590 Change permission to work per program

RubenGeo:


RubenGeo:
@elwinschmitz  Here you can get the new permission object

PR #3218, opened 2022-10-18
Feat.decimal transfer amount

jannisvisser:


services/121-service/src/payments/imagecode/image-code.service.ts
```diff
@@ -107,6 +107,25 @@ export class ImageCodeService {
...
+      } else {
+        return amount;
+      }
+    }
+    const amountParts = amount.split('.');
+    if (amountParts[1].length === 1) {
+      return `${amount}0`;
+    } else {
+      return `${amountParts[0]}.${amountParts[1].substring(0, 2)}`;
+    }
```
elwinschmitz, line 126: There is also a `padEnd`-method: https://devdocs.io/javascript/global_objects/string/padend

So, `amountParts[1].padEnd(2, '0')`.


But I have the feeling there must be a simpler way to get a "decimal currency amount string" from a (floating) number...

jannisvisser: @arsforza thanks. good suggestion. I implemented it with some small edits.

elwinschmitz (approved):

PR #3235, opened 2022-10-21
Fixes for portal after full test

PeterSmallenbroek:


interfaces/HO-Portal/src/app/services/error-handler.service.ts
```diff
@@ -66,12 +66,15 @@ export class ErrorHandlerService extends ErrorHandler {
...
 
   private replaceErrorMessages(errorMessage: string): string {
-    switch (errorMessage) {
-      case "The value '[]' given for the attribute 'skills' does not have the correct format for type 'multi-select'":
-        return this.translate.instant('common.answer-is-required');
-
-      default:
-        return errorMessage;
+    if (
+      errorMessage.includes(`The value '[]' given for the attribute '`) &&
```
jannisvisser, line 70: I tried this with a regular expression, but it didn't work and I couldn't understand why. So in the end I reverted to this super-ugly solution, which is GEFN.

PR #3275, opened 2022-11-01
Fix.api diagram

jannisvisser:


services/121-service/src/main.ts
```diff
@@ -17,6 +17,36 @@ import appInsights = require('applicationinsights');
...
+  // const mermaidEdges = edges
+  //   .filter(
+  //     ({ from, to }) =>
+  //       !genericModules.includes(from.module.name) &&
+  //       !genericModules.includes(to.module.name),
+  //   )
+  //   .map(({ from, to }) => `  ${from.module.name}-->${to.module.name}`);
+  // console.log('graph LR');
+  // console.log(mermaidEdges.join('\n'));
+
```
elwinschmitz, line 49: Can't we put this into a method/script that we can run manually? Like `npm run generate:api-graph`?
Or inside an if-`NODE_ENV===development`"?

That way the installed package is actually used and possible future issues with this piece of code are more easily spotted..

PR #3284, opened 2022-11-02
Endpoint to update monitoringQuestion

PeterSmallenbroek:


RubenGeo:
@PeterSmallenbroek Nice another PUT endpoint. I removed id from update-monitoring-question.dto as it did nothing.

PR #3313, opened 2022-11-15
feat: filter programs by query params AB#16273

arsforza:


arsforza:
https://github.com/global-121/121-platform/pull/3313/commits/a5dfc48d2e6dd0410c02a782744e50d1c10fb7f0

@elwinschmitz in this last commit I removed the redirect to /tabs/personal so we can use a shorter url with queryParams to link to clients.
Also, is Tab1 still in use. It feels like a relic from the past, where the app used tabs.

PR #3380, opened 2022-11-30
AB#16513 Create language column

RubenGeo:


services/121-service/seed-data/program/program-test.json
```diff
@@ -211,5 +211,6 @@
   "aboutProgram": {
     "en": "[about aid program]"
   },
-  "fullnameNamingConvention": ["name"]
+  "fullnameNamingConvention": ["name"],
+  "languages": ["en"]
```
elwinschmitz, line 215: I would suggest to have "at least more then 1" language available in the test-setup.

PR #3402, opened 2022-12-06
feat: remove logout btn iframe page AB#16651

jannisvisser:


jannisvisser:
already merging to test on test-vm also

elwinschmitz:
I think it is better to rephrase this to be a "show/hide log-out button"-feature.

That way it can be reused for other possible future applications/use-cases.
Now it requires extra/external knowledge about that there is an iframe-thing, that apparently alters the behavior of the user-state-component...

Thinking with starting with the smallest, most detailed part outwards.

PR #3404, opened 2022-12-06
feat: also disable link to change-pwd AB#16651

jannisvisser:


jannisvisser:
@elwinschmitz realized that also the 'link to change password' should be disabled..

PR #3410, opened 2022-12-06
fix: implement change PA-app to app AB#16564

jannisvisser:


tools/121-platform-https.conf
```diff
@@ -240,6 +240,7 @@ ServerSignature Off
   # Legacy support
   #
   Redirect permanent /HO-portal /portal
+  Redirect permanent /PA-app /app
```
elwinschmitz, line 243: Note to future-self: This needs to be added AFTER the "upgrade legacy urls"-period.

So that the server-specific config can catch anything more specific, otherwise the users will already have been redirected by this line before any server-specific rules would start to apply.

jannisvisser:
@elwinschmitz I reverted this (see last commit)

tools/121-platform-https.conf
jannisvisser: this change was reverted in the end. Instead put in the .env.conf file.

PR #3416, opened 2022-12-08
AB#16553 Script migrate pv to lvv

RubenGeo:


jannisvisser:
closing because the migration is done. this code is not needed in repository any more.

PR #3432, opened 2022-12-15
Chore: update translations

PeterSmallenbroek:


jannisvisser:
we'll keep this open until arabic is finished. As this is for documentation only anyway. The changes are already done via API on production.

PR #3450, opened 2022-12-21
fix: update unexisting reference

jannisvisser:


jannisvisser:
@elwinschmitz I updated this unexisting reference because it was breaking the 121-service on test-vm. I already updated it there manually to fix it. This is to bring the code back in line. I however wonder if we have to keep this seed-scripts for loading individual NLRC programs at all.. because now the nlrc-multiple script is the relevant one. But maybe just for documentation, just like we're keeping ethiopia, lebanon, etc.

elwinschmitz:
@jannisvisser Ow, yeah. I deleted the file via the GitHub web-interface, so I wasn't warned about any uses/references of the file. :/

We might need to make a (documented, weighted) decision on "keeping old stuff around"...
Sounds like a good "new years' resolution" :D

elwinschmitz (approved):

PR #3520, opened 2023-01-25
chore: changelog v1.78.0 AB#17533

jannisvisser:


jannisvisser:
this version was never deployed and is superceded by v1.79.0

PR #3547, opened 2023-01-31
feat: fix create program endpoint

jannisvisser:


services/121-service/src/programs/dto/create-program.dto.ts
```diff
@@ -57,11 +85,11 @@ export class CreateProgramDto {
   })
   public readonly currency: string;
 
-  @ApiProperty()
+  @ApiProperty({ example: 'week' })
```
elwinschmitz, line 88: We could possibly also use something like?

```suggestion
  @ApiProperty({
    example: 'week',
    enum: ['week', 'month'],
  })
```

jannisvisser (commented):
@RubenGeo @PeterSmallenbroek see here some draft options to add a bit more documentation to Swagger. You can find this under the 'Schema' tab (as opposed to 'Example value'). Not sure if we will use this more, but maybe.

PR #3576, opened 2023-02-07
Update interface dependencies

elwinschmitz:


jannisvisser:
@elwinschmitz technically speaking I feel like you should not be doing this right now. We said we would process dependabot PRs once per sprint, and that was already done in this sprint. That process is also meant so that we can spread out the work/understanding of this more over the team.

PR #3661, opened 2023-03-03
feat: PA profile page AB#12338

arsforza:


interfaces/HO-Portal/src/app/components/registration-activity-overview/registration-activity-overview.component.html
```diff
@@ -0,0 +1,86 @@
...
+      <ion-item
+        *ngFor="let item of getFilteredActivityOverview()"
+        style="font-size: 14px; --padding-start: 0"
+      >
+        <ion-grid>
+          <ion-row>
+            <div class="ion-padding-horizontal">
+              <div style="height: 100%; display: flex; align-items: center">
+                <ion-icon
+                  [name]="getIconName(item.type)"
```
elwinschmitz, line 59: This will break the production-build in that it won't detect the icons-in-use, so their files will not be included on the server...

We can work around that, I think...

elwinschmitz (changes requested):

PR #3813, opened 2023-04-17
Feat.Users List Endpoint

TatevAccenture:


agnesesaulite (commented):
Please add AB#18908 to the title with your task number

agnesesaulite (approved):

kaspars-grinbergs-accenture (approved):

services/121-service/src/user/user.controller.ts
```diff
@@ -45,6 +45,13 @@ export class UserController {
     return await this.userService.getUserRoles();
   }
 
+  @Admin()
```
elwinschmitz, line 48: I think the level of `Admin()`-only is a bit too strict for the use-case in question.

I think something like:
```
@UseGuards(PermissionsGuard, AdminAuthGuard)
```

And then with the correct specific `Permission`s defined.
Because the endpoint will be used in the Portal by specific Users only.

PeterSmallenbroek: But this is across programs right? And permissions are made with program assignments. So I'm not sure how that would work here.

jannisvisser: @TatevAccenture @kaspars-grinbergs-accenture I agree with Elwin's point, but I think it is not needed for this PR per se. I do think you should create another subtask within the same item to change the permissions indeed.

jannisvisser (approved):
PR approved, but see comment: please create a new subtask for this in the same item.

services/121-service/src/user/user.controller.ts
TatevAccenture: As written below comment we will create another subtask within the same item to change the permissions indeed, so I'm closing this conversation

jannisvisser:
@TatevAccenture feel free to merge it then

PR #3829, opened 2023-04-21
fix: Filter for FSP that are associated with program AB#19005

agnesesaulite:


TatevAccenture (approved):

jannisvisser:
@agnesesaulite @TatevAccenture @kaspars-grinbergs-accenture 

In another item I was working on, I stumbled upon this same bug, without knowing that this was on your backlog and you had an open PR about it. So this is already fixed now, in a slightly different way: see https://github.com/global-121/121-platform/commit/531a9c502625bd45294f79145dfb136030a76e3d. 

I will therefore close this PR.

PR #3850, opened 2023-04-26
Feat: Create PA from EspoCRM via API, Implement Validation for imported Registrations, and add Authentication AB#18173

TatevAccenture:


agnesesaulite (approved):

kaspars-grinbergs-accenture (approved):

agnesesaulite (dismissed):

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -200,25 +200,23 @@ export class RegistrationsController {
 
   @Permissions(PermissionEnum.RegistrationCREATE)
   @ApiOperation({
-    summary: 'Import set of registered PAs, from JSON only used in testing ATM',
+    summary: 'Import set of registered PAs from JSON ',
   })
   @ApiParam({ name: 'programId', required: true, type: 'integer' })
-  @Post('programs/:programId/registrations/import-registrations-cypress')
```
jannisvisser, line 206: 1. I don't think you can just change this existing endpoint. That is used with its old name in the Cypress tests.
2. Admittedly I'm not sure of the exact refinement here, but I would expect this endpoint to be in the espocrm module, and to be of a similar fashion as the updateRegistrations and deleteRegistrations endpoints already in there. So from the espocrm.controller it is fine to call the registration.service for the actual business logic.
3. Related to the above point, I would expect this item to also be tested by actually creating a PA in Espo and see if it creates a registration via the webhook. I can imagine you cannot do this yourselves, as you need access to Espo, so we should team up to do that. But it is necessary for this item to be completed, so please set up what you need.

@TatevAccenture @kaspars-grinbergs-accenture

kaspars-grinbergs-accenture: Thanks @jannisvisser, will look into this. Also saw Tijs comment regarding #2 on Teams.

TatevAccenture: I'm resolving this conversation , as we had after discussions and decisions for this

agnesesaulite (dismissed):

kaspars-grinbergs-accenture (dismissed):

agnesesaulite (dismissed):

kaspars-grinbergs-accenture (approved):

kaspars-grinbergs-accenture (approved):

agnesesaulite (dismissed):

RubenGeo:
@TatevAccenture 

I got a 500 error when calling the API

1. Seeded nlrc multiple
2. Called the endpoint with programId 1 and:
`{
  "namePartnerOrganization": "ABC",
  "preferredLanguage": "en",
  "paymentAmountMultiplier": 1,
  "nameFirst": "John",
  "nameLast": "Smith",
  "phoneNumber": "14155238886",
  "fspName": "Albert Heijn voucher WhatsApp",
  "whatsappPhoneNumber": "14155238886",
  "status": "registered"
}
`
4. Got back a 500 error:
image

I see in this item https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/18173/ in the comments that you were able to use it succesfully. Am I perhaps using a different setup or am I looking at a different version of the code?

RubenGeo:
I also pushed a commit with a description to that endpoint. Which we discussed internally in our team what we thought would be nice to have :smile:

TatevAccenture:
As I see you are passing object instead of array, need to pass body data [{}] as a array , after that as I see also your json data is not valid {   "namePartnerOrganization": "ABC",   "preferredLanguage": "en",   "paymentAmountMultiplier": 1,   "nameFirst": "John",   "nameLast": "Smith",   "phoneNumber": "14155238886",   "fspName": "Albert Heijn voucher WhatsApp",   "whatsappPhoneNumber": "14155238886",   "status": "registered" }  keys for nameFirst,  nameLast are not right need to be firstName, lastName, for example as a valid json I'm using this:
[ {
   "preferredLanguage": "en",
   "paymentAmountMultiplier": 1,
   "maxPayments": "",
   "firstName": "Test",
   "lastName": "Test",
   "phoneNumber": "141552388862",
   "fspName": "Intersolve-jumbo-physical",
   "whatsappPhoneNumber": "14155238886",
   "tokenCodeVisa": "",
   "isPhysicalCardVisa": "",
   "addressStreet": "Straat",
   "addressHouseNumber": 1,
   "addressHouseNumberAddition": "A",
   "addressPostalCode": "1234AB",
   "addressCity": "Den Haag",
   "referenceId": ""
 }]

RubenGeo:
@TatevAccenture 

- I added one commit to also throw 400 errors when you do not post an array
- The rest looks good :+1:  and merging it into master

Thanks for the work!

PR #3899, opened 2023-05-11
Test: added feature for manual test AB#19154

girtsfre:


TatevAccenture (dismissed):

PeterSmallenbroek:
@girtsfre 
A couple of remarks:
- The name of the file should always end in .feature
- The name of the file says 'for_espocrm', it should say 'from_espocrm'
- Not sure if its a preview bug but the indentation and new lines should be cleaned up
- On the contents of the test, the body properties depend on the chosen FSP. So I think it's not 100% correct that everything should always be provided.

PR #3909, opened 2023-05-16
Remove VM-specific code/config/docs AB#19346

elwinschmitz:


jannisvisser (dismissed):
@elwinschmitz the changes are fine. But do you already want to merge it, or wait a bit more until all the stuff around these items is finished?

PR #3918, opened 2023-05-18
FEAT: Change FSP from EspoCRM via API  AB#18557

TatevAccenture:


agnesesaulite (dismissed):

kaspars-grinbergs-accenture (dismissed):

services/121-service/src/registration/dto/set-fsp.dto.ts
```diff
@@ -14,8 +14,6 @@ export class SetFspDto {
 const fspArray = Object.values(FspName).map((item) => String(item));
 
 export class UpdateChosenFspDto {
-  @ApiProperty({ example: '910c50be-f131-4b53-b06b-6506a40a2734' })
-  @Length(5, 200)
   public readonly referenceId: string;
```
PeterSmallenbroek, line 17: @TatevAccenture I think this line also needs to go. As you now only removed the decorators and not the property itself.

TatevAccenture: Thanks for the comment @PeterSmallenbroek  I did the changes you can now review

kaspars-grinbergs-accenture (approved):

agnesesaulite (dismissed):

PR #3935, opened 2023-05-25
Fix: Add + to SMS and WhatsApp messages AB#19518

PeterSmallenbroek:


services/121-service/src/notifications/sms/sms.service.ts
```diff
@@ -23,7 +23,7 @@ export class SmsService {
         body: message,
         messagingServiceSid: process.env.TWILIO_MESSAGING_SID,
         statusCallback: EXTERNAL_API.smsStatus,
-        to: recipientPhoneNr,
+        to: `+${recipientPhoneNr}`,
```
RubenGeo, line 26: @PeterSmallenbroek please do this the same as in the whatsapp service

PR #3936, opened 2023-05-25
New feature added for manual testing AB#18557

girtsfre:
Created a feature file with test scenarios for testing in Swagger

### DIFF
--- a/features/Other/Edit_Info_Person_Affected_From_EspoCRM.feature
+++ b/features/Other/Edit_Info_Person_Affected_From_EspoCRM.feature
@@ -1,0 +1,35 @@
+@swagger-ui
+Feature: Edit info From ESPO CRM
+
+Background:
+Given a user in the Swagger UI
+
+Scenario: Successfully Update FSP from Whatsapp to Jumbo from EspoCRM
+
+    Given the "programId" is provided
+    Given the "refferenceId" is provided
+    Given the "NewFspAttributes" is provided
+    When 'newFspName' value is entered as 'Intersolve-jumbo-physical'
+    And the user has filled in values for "NewFspAttributes"
+    And calls the "registrations/update-chosen-fsp" endpoint
+    Then for PA FSP is changed 
+    And Code 201 is displayed
+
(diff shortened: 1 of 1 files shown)


TatevAccenture (dismissed):

features/Other/Edit_Info_Person_Affected_From_EspoCRM.feature
```diff
@@ -0,0 +1,35 @@
...
+Background:
+Given a user in the Swagger UI
+
+Scenario: Successfully Update FSP from Whatsapp to Jumbo from EspoCRM
+
+    Given the "programId" is provided
+    Given the "refferenceId" is provided
+    Given the "NewFspAttributes" is provided
+    When 'newFspName' value is entered as 'Intersolve-jumbo-physical'
+    And the user has filled in values for "NewFspAttributes"
```
jannisvisser, line 13: I believe the idea here is that there can only 1 'When' statement (and thus that there cannot be 'And's after a 'When'), as the 'When' should really pinpoint 1 specific action that is done by the user in a scenario. So all 'preparation' before this action should be under 'Given'. (Same for the other scenarios below.)

features/Other/Edit_Info_Person_Affected_From_EspoCRM.feature
```diff
@@ -0,0 +1,35 @@
+@swagger-ui
+Feature: Edit info From ESPO CRM
+
+Background:
+Given a user in the Swagger UI
+
+Scenario: Successfully Update FSP from Whatsapp to Jumbo from EspoCRM
+
+    Given the "programId" is provided
+    Given the "refferenceId" is provided
```
jannisvisser, line 10:
```suggestion
    Given the "referenceId" is provided
```

jannisvisser: same change for other scenarios below

jannisvisser (changes requested):
these are just some minor changes. if they're changed, i'll approve it.

jannisvisser (approved):

PR #3953, opened 2023-05-31
fix: ion-input label warnings AB#19636

arsforza:


elwinschmitz:
@arsforza There are still some mentions/logs about `type="email" name="emailAidworker"` (for the Portal)

I'm not sure if the other interfaces also had these messages? (You could check the build-runs)

elwinschmitz (approved):

PR #3994, opened 2023-06-15
Some irritating comments

diderikvw:


diderikvw:
Hi @jannisvisser , I am learning more about the 121 Platform by going through the code and trying to understand what is happening. When I have an opinion or question about something, I add a TODO or TODO REFACTOR comment. Can you see if the ones I made until now would be valuable for "other developers/later"? Or what I can/should do to make them better? Perhaps more optimal if we take 15' together and go through some of them.

jannisvisser:
Hi @diderikvw good effort. But I'm not convinced this is the optimal way of doing this. Because you can argue about some of these things, while this way it sounds very binary. But on the other hand, the "arguing" can be done during reviewing such PRs, so then it would maybe be all right. Anyway no detailed time today, so we'll talk further next week.

diderikvw:
Yes, good points @jannisvisser . Placing the comments are a way for me to understand the code, as a "reminder" to have a conversation at some point. Indeed during review of PR, or even beyond if we accept to leave them in the code "for now" (with perhaps some changes). Yes, I made them quite "binary", to keep them short. Otherwise there would be long sentences with a lot of "perhaps", "maybe", "could be" with lists of arguments.

PR #3995, opened 2023-06-15
Feat.retry payment

RubenGeo:


services/121-service/src/payments/payments.service.ts
```diff
@@ -69,12 +69,10 @@ export class PaymentsService {
     const payments = await this.transactionRepository
       .createQueryBuilder('transaction')
       .select('payment')
-      .addSelect('MIN(transaction.created)', 'paymentDate')
-      // TEMP FIX: this code will become obsolete once PBI #19365 is implemented. In the meantime this change will do, as the amount is only used for 'retry all', which is only possible with at least 1 failing transaction, which already contains the unmultiplied amount
-      // .addSelect(
-      //   `MIN(CASE WHEN transaction.status = 'error' THEN transaction.amount ELSE transaction.amount / coalesce(r.paymentAmountMultiplier, 1) END )`,
-      //   'amount',
-      // )
+      .addSelect(
```
jannisvisser, line 72: @RubenGeo @PeterSmallenbroek I believe 'amount' is not needed at all any more in this endpoint right? And either way this hacky if-error-then-unmultiplied-else-multiplied-amount code should belong to the past now right?

PR #4044, opened 2023-07-04
fix: use title attribute to show full datetime & fsp in payment histo…

jannisvisser:
…ry [AB#20060](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/20060)

### DIFF
--- a/services/121-service/src/payments/payments.service.ts
+++ b/services/121-service/src/payments/payments.service.ts
@@ -391,7 +391,6 @@ export class PaymentsService {
         .groupBy('data.id')
         .limit(1);
     }, 'paymentAddress');
-    console.log(q.getSql());
     return q;
   }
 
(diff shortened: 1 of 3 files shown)


services/121-service/src/payments/payments.service.ts
```diff
@@ -391,6 +391,7 @@ export class PaymentsService {
         .groupBy('data.id')
         .limit(1);
     }, 'paymentAddress');
+    console.log(q.getSql());
```
elwinschmitz, line 394: I don't think this needs to be here...

```suggestion
```

jannisvisser:
@elwinschmitz can I interpret your (resolved) comment also as "the rest was fine and approved"?

PR #4074, opened 2023-07-11
Feat.payment history in profilepage #AB19799

agnesesaulite:


package-lock.json
jannisvisser, line 1: @agnesesaulite what is this package-lock.json in the root? Is it a mistake?

interfaces/HO-Portal/src/app/pages/registration-details/registration-details.page.html
```diff
@@ -56,6 +56,7 @@
             [person]="person"
             [paId]="paId"
             [programId]="programId"
+            [program]="program"
```
jannisvisser, line 59: @agnesesaulite it seems unnecessary to pass both 'program' and 'programId'. So if you need the full 'program' object now, you canremove 'programId' here.

interfaces/HO-Portal/package.json
```diff
@@ -48,6 +48,7 @@
     "@swimlane/ngx-datatable": "^20.1.0",
     "core-js": "2.6.8",
     "file-saver": "^2.0.2",
+    "moment": "^2.29.4",
```
jannisvisser, line 51: What are you using 'moment' library for? This is for date/time operations right? I believe we should be consistent in the 121 platform on how we handle dates, and not use it some places and not at others.

jannisvisser (changes requested):
Hi Agnese
- See some inline comments/questions
- Furthermore: for a 'not yet sent' payment, the details after opening the accordion are empty, unlike the design where some fields are filled in.
- When you click the 'details'/'open voucher' btn this opens the popup and simultaneously opens the accordion in the background. I personally feel this is a bit confusing, and would opt to only open the popup when clikcing the button, and only open the accordion when clicking anywhere else.

agnesesaulite:
> Hi Agnese
> 
>     * See some inline comments/questions
A: Thanks!
> 
>     * Furthermore: for a 'not yet sent' payment, the details after opening the accordion are empty, unlike the design where some fields are filled in.
A: Yes, this was an ask from Tijs side, because, if the payment hasn't been done, then there shouldn't be any info there. 

>     * When you click the 'details'/'open voucher' btn this opens the popup and simultaneously opens the accordion in the background. I personally feel this is a bit confusing, and would opt to only open the popup when clikcing the button, and only open the accordion when clicking anywhere else.
A: I agree, this is confusing ionic accordion feature. I researched it but didn't find solution for it.

package-lock.json
agnesesaulite: it is being created automatically when there are changes in package.json
Removed it.

interfaces/HO-Portal/package.json
agnesesaulite: We are using moment library for date/time also in other interfaces, but were not yet using in HO, I think it is best approach for date formatting and takes less space and less complexity that's why I implemented it also in HO interface.

interfaces/HO-Portal/src/app/pages/registration-details/registration-details.page.html
agnesesaulite: I need both of them, because Program comes from Program class (Porgram model) but ProgramId comes from Person class (person model)

jannisvisser: I don't get this. I just did a commit to explain exactly what I mean. (Which we can revert of course if it turns out I'm wrong.)

interfaces/HO-Portal/package.json
jannisvisser: I don't see the moment library in the other interfaces?

PeterSmallenbroek: Also, moment is not actively supported anymore: "Moment.js is a legacy project, now in maintenance mode. In most cases, you should choose a different library."
If we really need this we should use something like date-fns or luxon

jannisvisser: OK, so please revert this then Agnese.

jannisvisser: @agnesesaulite where are you even using it? I don't see it anywhere?

interfaces/HO-Portal/src/app/pages/registration-details/registration-details.page.html
jannisvisser: I added one more commit to this, as programId was still used in the html as well. But I found out that all those parameters that you pass to the accordion component are not actually used there.

jannisvisser (approved):

kaspars-grinbergs-accenture (approved):

PR #4125, opened 2023-07-26
feat: added originalConversationId to payment report

Vladimir-Petkov:


services/121-service/src/export-metrics/export-metrics.service.ts
```diff
@@ -767,6 +768,11 @@ export class ExportMetricsService {
       )
       .setParameters(latestTransactionPerPa.getParameters())
       .leftJoin('transaction.registration', 'registration')
+      .leftJoin(
+        'SafaricomRequestEntity',
+        'safaricom_request',
+        'transaction.id = safaricom_request.transactionId',
+      )
```
jannisvisser, line 775: @Vladimir-Petkov this works, but to be honest this is exactly what I hoped to make clear in the call on Monday. This is not a generic solution, but a very hard-coded/fsp-specific solution within a generic function.

What I would be thinking of instead:
- get the originatorConversationId from transaction.customData, not from safaricom_request. (It's also in there right?) That already makes it more generic and also easier, because then you don't have to join with safaricom. 
- But then still you need a little bit of hardcoding to specify when to take something from customData and what, and when not. I'm thinking of a method like the following in export-metrics.service:
- `private getFspExportOptions(fsp: FspName) {
    if (fsp === FspName.safaricom) {
      return ['request.originatorConversationID'];
    } else {
      return [];
    }
  }`

jannisvisser (changes requested):
@Vladimir-Petkov see requested change in comment. Let me know if you want to discuss!

jannisvisser:
@Vladimir-Petkov FYI: this is the more generic solution we came up with. We also load tested it for performance.

PR #4173, opened 2023-08-10
feat: match old password before update with new one AB#20074

Vladimir-Petkov:
Matching the old password and if is incorrect return error message: "Your password was incorrect." and if is correct then update with new password.


services/121-service/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16-slim as base
+FROM node:18-slim as base
```
elwinschmitz, line 1: I don't think we want to mix in this big change in with this PR, or is it a blocking issue for the back-end changes?

elwinschmitz (changes requested):

services/121-service/Dockerfile
agnesesaulite: This is just for testing, will remove it when everything is ready. 
And this PR is not ready for review yet, because started to add front end side to it as requested to do it together with backend. Will ping you when it is ready  :)

PR #4223, opened 2023-08-21
Pagination

PeterSmallenbroek:


interfaces/HO-Portal/src/assets/styles/ngx-datatable/_ngx-datatable.component.scss
```diff
@@ -1,7 +1,6 @@
 // Source, see: https://github.com/swimlane/ngx-datatable/blob/16.1.1/projects/swimlane/ngx-datatable/src/lib/components/datatable.component.scss
 .ngx-datatable {
   display: block;
-  overflow: hidden;
```
elwinschmitz, line 4: Why this specific change?
I think we should update this whole file from v16.1.1 (as stated in the top comment, to v20.1.0.(the version we actually use)

And apply any overrides that are necessary _on top of that_, instead of changing the "out of the box behaviour" of the package.

services/121-service/src/shared/const.ts
```diff
@@ -0,0 +1,43 @@
+const nameConstraintQuestionsArray = [
```
elwinschmitz, line 1: This constant should be used in the test, instead of copy-pasting the values only.

services/121-service/test/registrations/get-registration.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    addressHouseNumberAddition: '',
+    addressPostalCode: '1234AB',
+    addressCity: 'Stad',
+  };
+
+  const attribute1 = 'whatsappPhoneNumber';
+  const attribute2 = 'addressCity';
+  const attribute3 = 'referenceId'
+
+  const exepectedAttributes = [
```
elwinschmitz, line 34: The "shared constant" `nameConstraintQuestionsArray` should be used here, instead of copy-pasting its values...

And the typo should be fixed. ;)

```suggestion
  const expectedAttributes = [
```

interfaces/HO-Portal/src/app/pages/registration-details/registration-details.page.ts
```diff
@@ -117,11 +118,11 @@ export class RegistrationDetailsPage implements OnInit, OnDestroy {
     const person = (
       await this.programsService.getPeopleAffected(
         this.programId,
-        this.canViewPersonalData,
-        this.canViewPaymentData,
+        1,
+        1,
```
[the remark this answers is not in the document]
PeterSmallenbroek: No, those are the pagesize and which page

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -341,6 +342,8 @@ export class ProgramPeopleAffectedComponent implements OnDestroy {
 
   private messageColumnStatus = MessageStatusMapping;
 
+  page = new PaginationMetadata();
```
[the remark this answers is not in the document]
PeterSmallenbroek: - The default is public, that's why it works already. If you want me to make it more explicit by adding 'public', sure.
- I'll rename it to 'pageMetaData'

interfaces/HO-Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -1317,15 +1314,13 @@ export class ProgramPeopleAffectedComponent implements OnDestroy {
   ) {
     let registrationsWithPayment;
     if (payment) {
```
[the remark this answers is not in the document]
PeterSmallenbroek: I've now commented out the if block. I've chosen to do this so this is still on our radar when we're going to do the 'Do payment' bulk action PBI

interfaces/HO-Portal/src/assets/styles/ngx-datatable/_ngx-datatable.component.scss
PeterSmallenbroek: I've updated the file

PR #4237, opened 2023-08-22
Upgrade 121-service to Node.js v18 AB#23354

elwinschmitz:


jannisvisser:
@elwinschmitz is this something you want to process? Because I was about to merge the nestjs-upgrade branch, and then saw this.

elwinschmitz:
@jannisvisser 
No I only 'parked' it here, to review/test some more later.

PR #4239, opened 2023-08-22
feat: Current password  AB#20074

agnesesaulite:


features/121-Portal/Change_password.feature
jannisvisser, line 1: @girtsfre please reference this file also in the menu of feature files in features/README.md (there is already a bullet for 'Change password' which had no file associated with it yet, so you can replace that).

features/121-Portal/Change_password.feature
```diff
@@ -0,0 +1,34 @@
...
+    And existing password is not at least 8 characters one capital letter, one lowercase letter and one number
+    And existing password field turns red
+    And new password is not at least 8 characters one capital letter, one lowercase letter and one number
+    And user enters different password in "Confirm password" field
+    And confirm password field turns red
+    And update button turns black\
+    And "Passwords do not match" is dispalyed under "Confirm password" field
+    Then when user clicks on "Update" button password is changed
+    And user is navigated to different page
+    And "Password change failed" is displayed
```
jannisvisser, line 34: @girtsfre it doesn't say this. It says 'Failed attempt'. But in fact, that feedback is not the most informative. Shouldn't it say something more like 'Failed attempt: current password incorrect'? @agnesesaulite

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -597,9 +597,11 @@
       "pageTitle": "User",
       "change-password": {
         "title": "Change password",
+        "existing-password": "Existing password",
```
jannisvisser, line 600: I would call this 'Current password' unless already otherwise agreed with someone else from our team.

interfaces/HO-Portal/src/assets/i18n/en.json
```diff
@@ -597,9 +597,11 @@
...
         "new-password": "New password",
         "confirm-password": "Confirm password",
-        "success": "Password changed"
+        "success": "Password changed",
+        "failed": "Failed attempt"
```
jannisvisser, line 604: I would call this: 'Failed attempt: current password incorrect'

interfaces/HO-Portal/src/app/user/user.page.html
```diff
@@ -133,19 +186,34 @@ <h1 class="text-uppercase">
...
-          <ion-icon
-            name="checkmark-circle"
-            color="success"
-            size="large"
-            aria-hidden="true"
-          ></ion-icon>
-          <br />
-          <strong>{{ 'page.user.change-password.success'|translate }}</strong>
-        </div>
+        <ng-container *ngIf="passwordChanged">
```
jannisvisser, line 189: Referring to discussion in the chat: was this design already discussed/aligned with Tijs or Esra?
- How it looks exactly
- That you see a different screen for a few seconds and then automatically go back (why not an alert/popup/etc?)

features/121-Portal/Change_password.feature
```diff
@@ -0,0 +1,34 @@
...
+    When user enters new password
+    And new password is at least 8 characters one capital letter, one lowercase letter and one number
+    And new password field turns green when avtive
+    And user enters same password in "Confirm password" field
+    And confirm password field turns green when avtive
+    And in both fields password match
+    And update button turns black
+    Then when user clicks on "Update" button password is changed
+    And user is navigated to different page
+    And "Password changed" is displayed
```
jannisvisser, line 21: It doesn't describe the observed behaviour that after a few seconds you automatically go back.

features/121-Portal/Change_password.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Background:
+    Given a logged-in user
+    And "Change password" page is open
+
+  Scenario: Change password succesfully
+    Given user has entered current password correctly
+    Given Existing password field apperas green when active
+    When user enters new password
+    And new password is at least 8 characters one capital letter, one lowercase letter and one number
```
jannisvisser, line 13: I think the rule is here you cannot have 'And's after a When. It should be immediately followed by a Then. So I think you should transform most of these to Given's and then take as When the action of clicking 'Update'.

features/121-Portal/Change_password.feature
```diff
@@ -0,0 +1,34 @@
...
+    And confirm password field turns green when avtive
+    And in both fields password match
+    And update button turns black
+    Then when user clicks on "Update" button password is changed
+    And user is navigated to different page
+    And "Password changed" is displayed
+
+  Scenario: Change password unsuccesfully
+    When user has entered current password incorrectly
+    And existing password is not at least 8 characters one capital letter, one lowercase letter and one number
```
jannisvisser, line 25: I don't think it's very insightful to repeat all the exact same things as above here. I get that you need to fill in the new password correctly to be able to hit 'Update', but still it's exactly the same as above, and it's just noise then. I would simply refer to the above scenario with something like 'Given for the rest everything the same as in previous scenario'

jannisvisser (dismissed):
- Functionally everything works as expected
- Main question is about the design of what happens after clicking 'Update'
- and some other minor points in this PR

PeterSmallenbroek (approved):

PeterSmallenbroek (approved):

PR #4265, opened 2023-08-29
feat: Remove CORS use (moved to Azure App Services configuration) AB#19333

elwinschmitz:
## To Do before release:
- [x] Set CORS-settings in Azure Portal for "121-NLRC" App-Service (like 121-krcs, etc..)


jannisvisser:
@elwinschmitz what is the deal with this?

elwinschmitz:
@jannisvisser Its a small refactor/simplification, see: AB#19333

But it also needs some proper testing, I think.

The test/staging-environments already use this, so no real blocking issues have popped up; But I wasn't so sure to already merge it.

elwinschmitz (commented):
TODO: Update .env.example

PR #4279, opened 2023-08-31
AB23420 Deploy "feat.paginate"-branch to test-environment

RubenGeo:


RubenGeo:
@elwinschmitz We want to keep deploying the paginate branch to test-vm so diderik/tijs can test on it. Are these the required changes to do that?

elwinschmitz (approved):
This would indeed 'change the default';

So everything that is merged/pushed to the branch `feat.paginate` will trigger a deployment to the test-App-Service+interfaces

We can still manually deploy the `master`-branch after this.

But we also could manually deploy the `feat.paginate`-branch every time we want to test something?

RubenGeo:
Yes, but we think we will be testing on the paginate branch more often than the master branch in the coming month. Therefore the switch makes sense. If we want to test often on both we may think about spinning up an extra slot + set of interfaces. But let's only go through this hassle if we find this approach to annoying. @elwinschmitz

PR #4301, opened 2023-09-08
Feat.See users AB#18857

agnesesaulite:


kaspars-grinbergs-accenture (approved):

services/121-service/src/user/user.service.ts
```diff
@@ -469,4 +471,19 @@ export class UserService {
...
+    return await this.userRepository.find({
+      select: {
+        id: true,
+        username: true,
+        admin: true,
+        active: true,
+        lastLogin: true,
+      },
+      where: {
+        userType: UserType.aidWorker,
```
RubenGeo, line 485: @Vladimir-Petkov If people register with the Person Affected app they would also appear in the query. That's why I added the where statement to only get aidworker users. (This is hard to know)

interfaces/HO-Portal/src/app/users/users-table/users-table.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+enum StatusName {
+  active = 'Active',
+  inactive = 'Inactive',
+}
+
+enum UserType {
+  admin = 'Admin',
+  regular = 'Regular',
+}
```
elwinschmitz, line 18: These string-values are not "data", right? They are to display only, so they need to be translatable.

So they need to be contained in the `en.json`-file.

To 'signal' that a special string is for "data-purposes" only, I would make them all lowercase otherwise.
Or just use an `enum` without any string-values. (TypeScript will use just numbers instead, but that won't matter if it doesn't have to match any database value.)

interfaces/HO-Portal/src/app/users/users.page.html
```diff
@@ -1,6 +1,121 @@
...
+          "
+            [attr.aria-label]="
+            'page.users.search'
+              | translate
+          "
+            (ionInput)="onInput($event.detail.value)"
+            [debounce]="1000"
+            [clearInput]="true"
+            class="filter-input"
+            data-cy="table-text-filter"
```
elwinschmitz, line 89:
```suggestion
```

We don't need hooks for Cypress-tests anymore.

interfaces/HO-Portal/src/app/users/users-table/users-table.component.html
```diff
@@ -0,0 +1,95 @@
...
+  data-target-width-element-selector="datatable-scroller"
+>
+  <div class="proxy-scrollbar--content"></div>
+</div>
+<ngx-datatable
+  class="bootstrap data-table--max-height"
+  [rows]="filteredRows"
+  [columnMode]="'force'"
+  [rowHeight]="50"
+  [scrollbarV]="true"
```
elwinschmitz, line 14:
```suggestion
  [rowHeight]="'auto'"
```

Quickly discussed with PO to tweak the design a little.

elwinschmitz (changes requested):

interfaces/HO-Portal/src/app/users/users.page.html
```diff
@@ -1,6 +1,121 @@
...
+        <app-users-table [filterValue]="filterInput"></app-users-table>
+      </div>
+    </div>
+
+    <!--tab 2 roles-->
+    <div
+      *ngIf="segment === 'roles'"
+      class="ion-padding"
+    >
+      <ion-text>Empty</ion-text>
```
elwinschmitz, line 118: This text needs to be translatable.

interfaces/HO-Portal/src/app/users/users.page.scss
```diff
@@ -0,0 +1,40 @@
...
+  margin-top: 30px;
+  padding-left: 32px;
+  border-bottom: 1px solid var(--fiveten-neutral-600);
+}
+
+.tab {
+  --background: #f4f7fa;
+  margin-right: 16px;
+  --color-checked: #4f22d7;
+}
```
elwinschmitz, line 14: These "tab"-styles are already defined somewhere else right?

Or we need to make them reusable on other pages.

interfaces/HO-Portal/src/app/users/users.module.ts
```diff
@@ -1,16 +1,24 @@
...
 import { SharedModule } from '../shared/shared.module';
+import { UsersTableComponent } from './users-table/users-table.component';
 import { UsersPage } from './users.page';
 
 @NgModule({
   imports: [
     CommonModule,
+    FormsModule,
     IonicModule,
+    HttpClientModule,
```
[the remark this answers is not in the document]
agnesesaulite: this is needed for tests to pass :)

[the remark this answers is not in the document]
elwinschmitz: Hmm.... it looks weird. :)

interfaces/HO-Portal/src/app/users/users.page.html
agnesesaulite: 'roles' is a segment value not text 
'Empty' will be replaced with roles and permissions cards, it is ready, but need to merge this first, because of the tabs.

services/121-service/src/user/user.service.ts
Vladimir-Petkov: @RubenGeo Understand, thank you for this feedback. I think only aidWorkers will be in Users table.

interfaces/HO-Portal/src/app/users/users.page.scss
agnesesaulite: hmmm not really because this is from the new design. But ok, will put it in global.scss.

interfaces/HO-Portal/src/app/users/users.page.html
elwinschmitz: Ah, yes. So its just a placeholder-during-development? 👍

services/121-service/src/user/user.service.ts
RubenGeo: Unfortunately not .You can try this locally if you want:
- Seed nlrc-multiple
- Go to program 1 and open it for registrations (by going to registration phase)
- Start the pa app (npm run start:pa)
- Open the PA app and register 1 PA
- Check you users tables> select * from users

RubenGeo: I happens through this endpoint user/person-affected

RubenGeo: It's very confusing. But it's legacy code from the PA app where people affected used to be able to login and logout

elwinschmitz (approved):

PR #4314, opened 2023-09-12
View Registration-profile in Redline-iframe AB#18640 AB#23442

elwinschmitz:


interfaces/HO-Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -129,9 +129,10 @@ export class ProgramsServiceApiService {
     programId: number | string,
     phase: ProgramPhase,
```
elwinschmitz, line 130: Would this work:

```
phase?: ProgramPhase = '',
```

To set it as "when undefined, use as empty-string".

PR #4326, opened 2023-09-13
Feat.subtitle login

tijsziere:


interfaces/HO-Portal/src/app/login/login.page.html
```diff
@@ -27,6 +27,16 @@ <h1 class="text-uppercase">
             {{ 'page.login.portalTitle'|translate }}
           </h1>
         </div>
+
+        <div
+        id="portal-title"
+        class="ion-margin-vertical"
+      >
+          <span style="font-size: 1.2em; font-style: italic; color: #888;">
+            Easy, Safe and Fast CVA
```
elwinschmitz: Text needs to be translatable; so should be contained in the `en.json`.

elwinschmitz (changes requested):

PR #4328, opened 2023-09-15
fix: export duplicates of empty values AB#19850

Vladimir-Petkov:


RubenGeo:
@Vladimir-Petkov 
- When I was reviewing your code I wanted to remove the hardcoded whatsappPhoneNumber because else there would a risk in the future if we have another 2 fsps with the same attribute. (For example if we have 2 FSPs with 'bankAccountNumber' as attribute you would have the same problem
- At first I thought this would be an easy change, but in the end I had to change more and more to get it to work.
- So in the end I changed quite a bit in your PR, leaving little of your original work. Sorry for that!
- Could you please review my code to check if I made any mistakes?

PR #4362, opened 2023-09-26
Feat.add isAdmin field in login response AB#23742

agnesesaulite:
Users table scroll added.
Is User an Admin check added.


jannisvisser (approved):
@agnesesaulite Given the assumption that the issues were
- that you could see the users in the menu-options when non-admin
- and that you couldn't scroll the table

.. I can confirm that both work as expected now, and therefore the PR is approved. I will merge it.

PR #4410, opened 2023-10-04
Feat.pagination multiple filter

PeterSmallenbroek:


interfaces/HO-Portal/src/app/program/table-filter-row/table-filter-row.component.ts
```diff
@@ -1,5 +1,8 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { Observable } from 'rxjs';
+import { RegistrationStatusEnum } from '../../../../../../services/121-service/src/registration/enum/registration-status.enum';
```
elwinschmitz, line 3: There is a 'local' enum that should be imported instead.

PR #4473, opened 2023-10-24
test: rm all empty spec files AB#24209

jannisvisser:


elwinschmitz:
Some of these files are also/already removed in the `paginate`-branch.. but it will be very easy to resolve those merge-conflicts. :D

elwinschmitz (approved):

PR #4485, opened 2023-10-24
refactor.api endpoints

Vladimir-Petkov:


services/121-service/test/helpers/utility.helper.ts
```diff
@@ -39,7 +39,7 @@ export async function getAccessToken(): Promise<string> {
 }
 
 export function loginAsProgramManager(): Promise<request.Response> {
-  return getServer().post(`/user/login`).send({
```
RubenGeo, line 42: I think this change will break our external api integrations (e.g. kobo connect/espocrm) @jannisvisser

PR #4486, opened 2023-10-24
Docs: add responses Swagger AB#24019

jannisvisser:


services/121-service/src/user/user.controller.ts
```diff
@@ -88,6 +107,15 @@ export class UserController {
   @ApiOperation({ summary: 'Delete existing user role' })
   @ApiParam({ name: 'userRoleId', required: true, type: 'integer' })
   // TODO: REFACTOR: rename to /users/roles/
+  @ApiResponse({
+    status: 200,
+    description: 'Returns the deleted role',
```
jannisvisser, line 112: @PeterSmallenbroek is this actually logical / good practice to do in case of a DELETE endpoint?

PR #4491, opened 2023-10-25
fix: log all requests and responses including body AB#24300

Vladimir-Petkov:


jannisvisser:
integrated in release/v1.103.3 branch

PR #4494, opened 2023-10-25
Fix.prohibit in backend new payment if still in progress

Vladimir-Petkov:


jannisvisser:
integrated in release/v1.103.3 branch

PR #4575, opened 2023-11-07
AB#24482 Docker test redit

RubenGeo:


.github/workflows/test_service_api.yml
```diff
@@ -47,6 +47,8 @@ jobs:
           /bin/sed -i '/^GENERIC_THROTTLING_LIMIT=/s/=.*/=1200/' ./services/.env.example
           /bin/sed -i '/^HIGH_THROTTLING_TTL=/s/=.*/=60/' ./services/.env.example
           /bin/sed -i '/^HIGH_THROTTLING_LIMIT=/s/=.*/=100/' ./services/.env.example
+          /bin/sed -i '/^REDIS_HOST=/s/=.*/=services-121-redis-1/' ./services/.env.example
+          /bin/sed -i '/^REDIS_PORT=/s/=.*/=6379/' ./services/.env.example
```
elwinschmitz, line 51: @RubenGeo You need to make sure these 2 `ENV`-variables are already present in the `.env.example`-file.

These commands here only "find-and-replace" something, they don't automatically "add" things.

And when you add them, add them in a relevant section (near the database ones) with relevant commands above them...

PR #4591, opened 2023-11-13
Add REDIS_PREFIX AB#24649

RubenGeo:


services/121-service/src/main.ts
```diff
@@ -71,6 +71,15 @@ function generateModuleDependencyGraph(app: INestApplication): void {
 async function bootstrap(): Promise<void> {
   const app = await NestFactory.create(ApplicationModule);
 
+  if (process.env.NODE_ENV !== 'development' && !process.env.ENV_NAME) {
+    throw new Error('ENV_NAME not set when not in development mode');
+  }
+  if (process.env.ENV_NAME.includes(':')) {
```
elwinschmitz, line 77: Lets also include these specifics in `.env.example`;
So that everybody is aware of this 'dual use' of the `ENV_NAME`-string.

You could also just 'sanitize' the string if there is one (removing any special characters, but using the rest as-is).

services/121-service/src/main.ts
```diff
@@ -71,6 +71,15 @@ function generateModuleDependencyGraph(app: INestApplication): void {
 async function bootstrap(): Promise<void> {
   const app = await NestFactory.create(ApplicationModule);
 
+  if (process.env.NODE_ENV !== 'development' && !process.env.ENV_NAME) {
```
elwinschmitz, line 74: There is already the const `DEBUG` that can be used here, instead of a special custom string-check.

elwinschmitz (approved):

PR #4640, opened 2023-11-24
Changed "121 development team" text according to requirement AB#25029

diderikvw:


diderikvw:
There are more occurrences of "121 development team", which however do relate to technical errors. To align with PO on what to do with these, so please do not merge yet.

PR #4642, opened 2023-11-24
Renamed HO-Portal to Portal and ho-portal to portal and HO-portal to …

diderikvw:
…Portal


diderikvw:
I renamed the front-end code occurrences and the folder with the interface. Restarted locally and the PA App, AW App and Portal are running. Not sure what else to test. Is the term "HO-Portal" anywhere in data in the database? I do not think it is in any .env variable, so no need to change it there.

diderikvw:
@arsforza  or @PeterSmallenbroek , can you have a look at this PR?

diderikvw:
Or @elwinschmitz , do you have any insight on extra tests we should do, or things to change/be aware of for our instances?

elwinschmitz (changes requested):
@diderikvw Make sure to add a note in the `CHANGELOG` under the "Unreleased / Changed"-section about this.

elwinschmitz (dismissed):

interfaces/Portal/src/app/auth/auth.service.ts
```diff
@@ -12,7 +12,7 @@ import Permission from './permission.enum';
 export class AuthService {
   public redirectUrl: string;
   public isIframe: boolean;
-  private userKey = 'logged-in-user-HO';
+  private userKey = 'logged-in-user-portal';
```
elwinschmitz, line 15: @jannisvisser This will log-out all currently logged-in users. 

Not the end of the world, but still might be good to warn/notify users.

jannisvisser: @elwinschmitz made a note in release item

PR #4644, opened 2023-11-24
Feature.do import of 10 k p as in programs

Vladimir-Petkov:


RubenGeo:
We decided ot to return early. Created this PR instead to limit to 1000 registration: https://github.com/global-121/121-platform/pull/4659

PR #4645, opened 2023-11-27
First go, copy of bob finance

diderikvw:


jannisvisser:
@diderikvw I think this work is outdated now and this PR can be removed right?

PR #4684, opened 2023-12-05
fix: use scope in patch registration endpoint AB#25146

jannisvisser:


jannisvisser:
closed because already merged in other branch locally

PR #4760, opened 2023-12-27
feat: add language switcher AB#25345

arsforza:


elwinschmitz (dismissed):

interfaces/Portal/src/app/services/language.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  constructor(private translate: TranslateService) {
+    if (!this.translate) {
+      return;
+    }
+    this.translate.setDefaultLang(this.DEFAULT_LANGUAGE_KEY);
+
+    if (!this.selectedLanguage && !this.getStoredLanguage()) {
+      this.selectedLanguage = this.DEFAULT_LANGUAGE_KEY;
+      this.storeLanguage();
```
elwinschmitz, line 27: We don't need to specifically store the Default value, actually, right?

interfaces/Portal/src/app/services/language.service.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    this.selectedLanguage = this.getStoredLanguage();
+
+    this.translate.use(this.selectedLanguage);
```
elwinschmitz, line 32: I think that setting the document Language + direction needs to happen right after here... (and not in the current place)

```suggestion
    this.translate.use(this.selectedLanguage);
    document.documentElement.lang = this.selectedLanguage;
    document.documentElement.dir = this.translate.instant('_dir');
```

Or is the `lang`-attribute already set by the ngx-translate-package?

interfaces/Portal/src/app/services/language.service.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  public changeLanguage(languageKey: string) {
+    if (!this.translate) {
+      return;
+    }
+    this.translate.use(languageKey);
+    this.selectedLanguage = languageKey;
+    this.storeLanguage();
+    window.location.reload();
+    document.documentElement.dir = this.translate.instant('_dir');
```
elwinschmitz, line 71: This will not have any real effect... as the page will be reloading, right?

interfaces/Portal/src/app/services/language.service.ts
```diff
@@ -0,0 +1,96 @@
...
+  private DEFAULT_LANGUAGE_KEY = 'en';
+  private DEFAULT_LANGUAGE_LABEL = 'ENGLISH';
+
+  private selectedLanguage: string;
+
+  constructor(private translate: TranslateService) {
+    if (!this.translate) {
+      return;
+    }
+    this.translate.setDefaultLang(this.DEFAULT_LANGUAGE_KEY);
```
elwinschmitz, line 23: Isn't this already done by the App-component? Or in the App-Module?

interfaces/Portal/src/app/services/language.service.ts
```diff
@@ -0,0 +1,96 @@
...
+  key: string;
+  name: string;
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LanguageService {
+  private DEFAULT_LANGUAGE_KEY = 'en';
+  private DEFAULT_LANGUAGE_LABEL = 'ENGLISH';
```
elwinschmitz, line 15: Is it in capitals on purpose?

```suggestion
  private DEFAULT_LANGUAGE_LABEL = 'English';
```

PR #4782, opened 2024-01-08
Make  hard-coded  strings translatable

elwinschmitz:


interfaces/Portal/src/app/components/program-card/program-card.component.html
```diff
@@ -32,35 +32,25 @@ <h1>
...
-    </ion-row>
+    <ion-card-title color="primary">
+      <strong class="program-title">{{ program.titlePortal }}</strong>
+    </ion-card-title>
+    <ion-card-subtitle color="medium">
+      <small>
+        <strong>
+          {{ 'page.programs-list.card.updated' | translate }}
+          {{ program.updated | date: DateFormat.dayAndTime : locale }}
+        </strong>
```
elwinschmitz, line 43: Could maybe be combined in 1 translatable-string with a "{{date}}"-placeholder.

```suggestion
        <strong>
          {{
            'page.program-list.card.updated' |
              translate: {
                date: (program.updated | date: DateFormat.dayAndTime : locale)
              }
          }}  
        </strong>
```

That way the position of the date + its surrounding syntax (":") can be adjusted per language.

interfaces/Portal/src/app/components/physical-card-popup/physical-card-popup.component.html
```diff
@@ -68,67 +69,77 @@
...
       </ion-label>
-      <ion-label class="card-key-value ion-text-right"
-        ><strong
-          >{{ card?.balance / 100 | number: '1.2-2' }} {{ currency }}</strong
-        >
+      <ion-label class="card-key-value ion-text-end">
+        <strong>
+          {{
+            card?.balance / 100 | currency: currency : 'code' : '1.2-2' : locale
+          }}
```
elwinschmitz, line 82: The position of the Currency-symbol in relation to the number is not always the same in every language.

This way will fix that for us.

interfaces/Portal/src/app/components/physical-card-popup/physical-card-popup.component.html
```diff
@@ -68,67 +69,77 @@
...
+      <ion-label class="card-key-value">
+        {{
           'registration-details.physical-cards-overview.spent-this-month'
             | translate
               : {
-                  maxAmount: card?.maxToSpendPerMonth / 100,
-                  currency: currency
+                  maxAmount:
+                    card?.maxToSpendPerMonth / 100
+                    | currency: currency : 'code' : '1.2-2' : locale
```
elwinschmitz, line 96: Including the amount into the sentence is already a good idea, the positioning of the currency-symbol should not be a manual concern of the translator anymore...

interfaces/Portal/src/app/components/physical-card-popup/physical-card-popup.component.html
```diff
@@ -68,67 +69,77 @@
...
                 }
         }}:
```
elwinschmitz, line 98: Actually, these types of syntax should be included in the translatable-text.

So translators can make their own choices for whitespace around it...

(But probably not worth it to "fix every occurence" right now...)

PR #4798, opened 2024-01-10
Feat.template labels

jannisvisser:


services/121-service/migration/1701426130489-CreateAndUpdateMessageTemplates.ts
```diff
@@ -8,7 +8,7 @@ export class CreateAndUpdateMessageTemplates1701426130489
   public async up(queryRunner: QueryRunner): Promise<void> {
     // Commit transaction because the tables are needed before the insert
     await queryRunner.commitTransaction();
-    await this.migrateData(queryRunner);
+    // await this.migrateData(queryRunner);
```
elwinschmitz, line 11: This would  still need to happen the first time this code is deployed/run on any instance, right?

So we can only remove/comment-out this code AFTER it was part of a deployed release, right?

jannisvisser: @elwinschmitz yes, but this was already released right? (only relevant for NLRC)

services/121-service/src/scripts/seed-helper.ts
```diff
@@ -354,13 +356,17 @@ export class SeedHelper {
     language: string,
     message: string,
     isWhatsappTemplate: boolean,
+    isSendMessageTemplate: boolean,
+    label: JSON,
```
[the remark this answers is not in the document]
jannisvisser: good point, will change it.

PR #4803, opened 2024-01-12
Feat.supports PAs with empty phone number

jannisvisser:


elwinschmitz (changes requested):
I don't agree with the changes around the `formatPhoneNumber`-helper.

interfaces/Portal/src/app/program/export-select/export-select.component.ts
```diff
@@ -7,13 +7,16 @@ import { ProgramPhase } from '../../models/program.model';
...
   @Input()
   public programId: number;
 
   @Input()
   public thisPhase: ProgramPhase;
 
+  @Input()
+  public showValidation: boolean;
+
   public isPopoverOpen = false;
```
elwinschmitz, line 20:
```suggestion
```

This property isn't used anymore, right?

services/121-service/src/registration/services/registrations-import.service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  const importRegistrationsCsvInput = [
+    {
+      namePartnerOrganization: 'ABC',
+      preferredLanguage: language,
+      paymentAmountMultiplier: '',
+      maxPayments: '5',
+      nameFirst: 'Test',
+      nameLast: 'Test',
+      phoneNumber: '31600000000',
+      fspName: 'Intersolve-voucher-paper',
```
elwinschmitz, line 22: There's am Enum for this value as well, right?

services/121-service/src/registration/services/registrations-import.service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      await expect(
+        registrationsImportService.validateRegistrationsInput(
+          importRegistrationsCsvInput,
+          programId,
+          userId,
+        ),
+      ).rejects.toHaveProperty('response', [
+        {
+          lineNumber: 1,
+          column: 'phoneNumber',
```
elwinschmitz, line 78: There is `GenericAttributes.phoneNumber` for this, looking at the actual Service.

services/121-service/src/utils/phone-number.helpers.ts
```diff
@@ -5,18 +5,12 @@
...
-  }
-
   // Strip all (possibly existing) prefixes
   const onlyNumbers = phoneNumber.replace(/\D/g, '');
-
-  // What is left, could be invalid already
+  // If result is empty, then do not add a "+"-prefix
   if (!onlyNumbers) {
-    throw new Error('No valid phone number provided');
+    return onlyNumbers;
```
elwinschmitz, line 12: This is not the promised outcome of this helper-method.

"only the digits" is not "a formatted phone-number in the shape of: `+0000000000`".

If you need the phone-number as as formatted string, you should use this helper. If you don't want that, you should _not_ use this helper.

services/121-service/src/registration/services/registrations-import.service.spec.ts
jannisvisser: done

services/121-service/src/registration/services/registrations-import.service.spec.ts
jannisvisser: done

services/121-service/src/registration/registrations.service.ts
```diff
@@ -415,7 +416,14 @@ export class RegistrationsService {
...
     customDataKey: string,
     customDataValue: string | number | string[],
+    programId: number,
   ): Promise<string | number | string[]> {
+    const allowEmptyPhoneNumber = (
+      await this.programRepository.findOneBy({
+        id: programId,
+      })
+    )?.allowEmptyPhoneNumber;
+
```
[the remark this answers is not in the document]
jannisvisser: then we have to execute this query at all occurences before this function is called.

interfaces/Portal/src/app/program/export-select/export-select.component.ts
jannisvisser: it is right? When I search all for it, it clearly seems to be used.

services/121-service/src/utils/phone-number.helpers.ts
jannisvisser: You're right that it doesn't align with what's promised. So I changed the return description now.

I do stand by the dropping of the error throwing given that we need to allow empty phoneNumber sometimes now given this new feature.

services/121-service/src/registration/registrations.service.ts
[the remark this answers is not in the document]
elwinschmitz: It could still be done once somewhere else, but then re-used and "handed-down" like the Program-ID is currently.

elwinschmitz (approved):

PR #4878, opened 2024-02-02
Show message statuses in PA Profile Activity Overview

diderikvw:


interfaces/Portal/src/app/components/registration-activity-overview/registration-activity-overview.component.ts
```diff
@@ -83,6 +85,7 @@ export class RegistrationActivityOverviewComponent implements OnInit {
     ActivityOverviewType.status,
   ];
   public canUpdatePersonalData: boolean;
+  public chipStatus = MessageStatusMapping;
```
jannisvisser, line 88: I don't think 'chipStatus' is the right name here, because this thing is the full enum, and not the status of the chip. At its most literal, messageStatusMappingEnum could work alternatively?

jannisvisser (changes requested):

PR #4882, opened 2024-02-05
fix: Use icon in correct (and smaller) file-format

elwinschmitz:


jannisvisser:
@elwinschmitz this is not related to a task I think. Should it be reviewed? Or can it be merged?

elwinschmitz:
There's not much to be reviewed :)

But I thought that merging this would trigger a deploy to that new instance...  But that is of course not the case. 🫠 ;)

PR #4956, opened 2024-02-26
feat: Visa in PV

jannisvisser:


services/.env.example
```diff
@@ -168,7 +168,6 @@ INTERSOLVE_VISA_OIDC_ISSUER=
 INTERSOLVE_VISA_API_URL=
 INTERSOLVE_VISA_CLIENT_ID=
 INTERSOLVE_VISA_CLIENT_SECRET=
-INTERSOLVE_VISA_BRAND_CODE=
```
elwinschmitz, line 171: This ENV-variable is still required, right?

Not in local development, but "while setting up an instance", right?

RubenGeo: It's mostly needed in local development

PR #4976, opened 2024-02-27
AB#26732 Fix: Throw 401 if token is expired and refacture guards

RubenGeo:


jannisvisser:
This is outdated and works fine in the SSO branch.

PR #4984, opened 2024-02-29
Update portal translations

smeber:


interfaces/Portal/src/assets/i18n/nl.json
```diff
@@ -419,7 +419,7 @@
           "custom-attribute-true": "Ja",
           "dob": "Geboortedatum",
           "failedPayment": "Mislukt bij betaling #",
-          "fspDisplayNamePortal": "Financiële dienstverlener",
+          "fspDisplayNamePortal": "Type hulp",
```
RubenGeo, line 422: This translation is only specific for schoolmaaltijd. @tijsziere  Is that okay?

PR #4989, opened 2024-03-01
fix: add all tables to readonlyuser AB#26767

jannisvisser:


services/121-service/migration/1709300672260-Readonlyuser.ts
```diff
@@ -0,0 +1,18 @@
...
+      `SELECT * FROM pg_roles WHERE rolname='readonlyuser'`,
+    );
+    if (doesReadonlyuserExist.length) {
+      await queryRunner.query(
+        `GRANT SELECT ON all tables in schema "121-service" TO readonlyuser;`,
+      );
+    }
+  }
+
+  public async down(_queryRunner: QueryRunner): Promise<void> {
```
ashishkumarbasak-accenture, line 15: Why we are not using the revert query here?

```
public async down(queryRunner: QueryRunner): Promise<void> {
    const doesReadonlyuserExist = await queryRunner.query(
      `SELECT * FROM pg_roles WHERE rolname='readonlyuser'`,
    );
    if (doesReadonlyuserExist.length) {
      await queryRunner.query(
        `REVOKE SELECT ON ALL TABLES IN SCHEMA "121-service" FROM readonlyuser;`,
      );
    }
}
```

jannisvisser: as there is no use case for this. This is meant as an infrequently used way of making sure this user has access also to all new tables since last time it was executed. If you know of a better way to do this (whereby this user automatically gets select access also to new tables, then please go ahead).

ashishkumarbasak-accenture: OK got your point. Its OK from my side to keep the `down` method empty.
But normally I didn't see like this empty method instead the opposite of `up` method, thats why asked about it. Thanks for clarification.

ashishkumarbasak-accenture (approved):
Looks great to me, Approving!

PR #5019, opened 2024-03-11
Use `displayName` instead `fspDisplayNamePortal` and `fspDisplayNamePaApp` for financial_service_provider

ashishkumarbasak-accenture:


services/121-service/migration/1710079688508-add-display-name-to-financial-service-provider.ts
RubenGeo, line 1: If you now use the reset endpoint http://localhost:3000/docs/#/instance/ScriptsController_resetDb the fspDisplayName column is empty in the database. This is because in the fsp json files the old naming (fspDisplayNamePortal &  fspDisplayNamePaApp) is still used. These files should also use displayName instead of  seed-data/fsp/fsp-intersolve-voucher-whatsapp.json.

services/121-service/migration/1710080807910-migrate-fsp-display-name-portal-data-to-display-name.ts
RubenGeo, line 1: The migration works perfect thanks!

services/121-service/migration/1710079688508-add-display-name-to-financial-service-provider.ts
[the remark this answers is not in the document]
ashishkumarbasak-accenture: I just wanted to do the changes step by step just for clear visibility about the changes that we are doing rather then doing in a single complicated step, nothing else. BTW as you told, it doesn't really matter if we do it together or one by one.

interfaces/PA-App/src/app/personal-components/select-fsp/select-fsp.component.ts
```diff
@@ -62,13 +62,10 @@ export class SelectFspComponent extends PersonalDirective implements OnInit {
     this.conversationService.startLoading();
     // NOTE: we use this attribute being empty to avoid a new fsp-attribute, while this is a short-term feature only
     this.fsps = this.program.financialServiceProviders.filter(
-      (fsp) => fsp.fspDisplayNamePaApp,
+      (fsp) => fsp.displayName,
```
elwinschmitz, line 65: Also check on one of the 2 Intersolve FSP-names that we _do_ want to show...

```suggestion
      (fsp) => fsp.displayName && ['Intersolve-whatsapp', 'Intersolve-paper-voucher'].includes(fsp.fsp),
```

(or similar)

PR #5092, opened 2024-03-28
Feat. Free text search

Vladimir-Petkov:


interfaces/Portal/src/app/program/table-filter-row/table-filter-row.component.html
```diff
@@ -12,13 +12,17 @@
       class="util--flex-row ion-align-items-center"
       style="flex-basis: 80%"
     >
-      <span>
+      <span
+        class="ion-align-items-center ion-text-right"
+        style="width: 9rem"
```
elwinschmitz, line 17: I think this should be a `min-width` or a `max-width` only.. Or a percentage.

Because otherwise the layout isn't as flexible as it should be.

interfaces/Portal/src/app/program/table-filter-row/table-filter-row.component.html
```diff
@@ -12,13 +12,17 @@
...
+      >
         <ion-icon
           name="funnel"
           color="black"
           aria-hidden="true"
-          class="ion-margin-start"
         ></ion-icon>
+
+        {{ 'page.program.program-people-affected.filter-label' | translate }}
```
elwinschmitz, line 25: I think this text-labels makes more sense if it sits 'next to': `page.program.table-filter-row.select-column-to-filter`, right?

So inside the `table-filter-row`-group.

But this kind of issues is also a good candidate for a more helpful/specific guideline in the wiki :)

interfaces/Portal/src/app/services/filter.service.ts
```diff
@@ -35,6 +36,20 @@ export enum FilterOperatorEnum {
   providedIn: 'root',
 })
 export class FilterService {
+  public DEFAULT_FILTER_OPTION: Filter = {
+    name: 'quickSearch',
```
elwinschmitz, line 40: Will this string-value be also used as a query-parameter in the request? Then we can make it 'the same' by using an enum-value somewhere..

Are other query-parameters already somewhere defined in an enum?
Somewhere near `FilterOperatorEnum`?

interfaces/Portal/src/app/services/filter.service.ts
```diff
@@ -35,6 +36,20 @@ export enum FilterOperatorEnum {
...
+    label: this.translate.instant(
+      'page.program.program-people-affected.filter-quick-search',
```
elwinschmitz, line 42: Also move this text-label into the `table-filter-row`-group.

interfaces/Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -600,6 +604,13 @@ export class ProgramsServiceApiService {
     sort?: PaginationSort,
     // TODO: Fix the 'any' for the 'links' parameter
   ): Promise<{ data: Person[]; meta: PaginationMetadata; links: any }> {
+    const quickSearch = filters
+      ? filters.find(
+          (filter) =>
+            filter.name === this.filterService.DEFAULT_FILTER_OPTION.name,
```
elwinschmitz, line 610: On second thought @arsforza ... We should be sharing this name-string via an Enum here and not through the 'whole FilterService`, I think.

To prevent a really complicated dependency-diagram etc.. But enum-values are 'just strings', so there is less complexity necessary.

interfaces/Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -616,6 +627,10 @@ export class ProgramsServiceApiService {
     if (statuses) {
       params = params.append('filter.status', `$in:${statuses.join(',')}`);
```
elwinschmitz, line 628: All these query-parameters should maybe be defined in an enum somwehere...
And the `$in:`-part should then also be used as the enum-value, instead of hard-coded like this...

These parameter-values are  "something of the API that the Portal isn't in charge of"...

interfaces/Portal/src/assets/i18n/en.json
```diff
@@ -454,7 +454,9 @@
           },
           "multiple-choice-hidden-options": "Above list only shows a status if assigned to at least one Person Affected"
         },
+        "filter-label": "Select filter:",
         "filter-placeholder": "Type value...",
```
elwinschmitz, line 458: This `filter-placeholder` can also be moved to the `filter-row`-related group, I think.

services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -133,8 +133,9 @@ export class RegistrationsPaginationService {
     }
 
     if (hasPersonalReadPermission) {
-      paginateConfigCopy.relations = ['data'];
-      paginateConfigCopy.searchableColumns = ['data.(value)'];
+      paginateConfigCopy.relations = ['data', 'dataSearchBy'];
```
elwinschmitz, line 136: Can these magic strings also be used from enum-values throughout the back-end?
To keep track of all use-cases automatically? And to prevent typo's, etc.

elwinschmitz (changes requested):

elwinschmitz (dismissed):

PR #5095, opened 2024-03-29
fix: fspDisplayName does not work in message placeholder AB#27260

Vladimir-Petkov:


elwinschmitz (changes requested):
The "TranslatableStringService" should be used to get the right language-version 'out' of an object/string.

PR #5104, opened 2024-04-03
Feat: Change password without current password

PeterSmallenbroek:


services/121-service/src/user/user.controller.ts
```diff
@@ -490,4 +492,21 @@ export class UserController {
   ): Promise<GetUserReponseDto[]> {
     return await this.userService.getUsersInProgram(Number(params.programId));
   }
+
+  @Admin()
+  @ApiTags('users')
+  @ApiOperation({ summary: 'Reset user password without current password' })
+  @ApiResponse({
+    status: 204,
```
elwinschmitz, line 500: You could also use: `status: HttpStatus.NO_CONTENT,` here..

Or even `ApiNoContentResponse({ description: '...' })`...

But I think picking 1 of these 3 methods (and then use it consistently for all endpoints) is better. ;)

PR #5108, opened 2024-04-04
fix wrong description

tijsziere:


elwinschmitz (approved):
Looks more accurate indeed.

but… I think this small ‘bug’ was already resolved in the SSO-branch. And I don’t think it is useful to give @PeterSmallenbroek  any more new merge-conflicts to solve on that branch. ;)


So I think we can close this unmerged. 😘

PR #5110, opened 2024-04-05
fix: display translatable string for FSP name

ashishkumarbasak-accenture:


interfaces/Portal/src/app/services/past-payments.service.ts
```diff
@@ -110,6 +112,7 @@ export class PastPaymentsService {
         person.referenceId,
         transactions,
       );
+      transaction.fspName = this.translatableString.get(transaction.fspName);
```
elwinschmitz, line 115: Looking at lines 120-125+ it seems that `transaction` can also be `null`/`undefined`, so this line would mess with that..

So I think moving this one line to inside the `else`, would be safer...

elwinschmitz: We could also look into the 'components that display this' layer, to apply the `translatable-string-magic`....

But for now this Service seems ok.

elwinschmitz (changes requested):

interfaces/Portal/src/app/services/past-payments.service.ts
ashishkumarbasak-accenture: Just move that line to else block, OK, Thanks!

elwinschmitz (approved):

PR #5111, opened 2024-04-05
AB#27359 remove PA-App folder and related scripts

aberonni:
In a couple of places I wasn't sure about my choices - you can find these by searching for `XXX:` in the Files Changed tab. Would love some feedback on those choices (maybe worth going through them together with someone knowledgeable on the files I've changed so I can also gain some more context).


aberonni:
In a couple of places I wasn't sure about my choices - you can find these by searching for `XXX:` in the Files Changed tab. Would love some feedback on those choices (maybe worth going through them together with someone knowledgeable on the files I've changed so I can also gain some more context).

tools/git-hooks/pre-commit
```diff
@@ -22,7 +22,6 @@ check_if_need_to_lint() {
 }
 
 # Include all packages/applications that support an `npm run lint` task:
-check_if_need_to_lint "interfaces/PA-App/"
```
elwinschmitz, line 25: The bits about “keeping things in sync” in this file can also be removed. There’s not much similar-enough anymore to keep it.

aberonni: Done!

aberonni:
@elwinschmitz meanwhile I've also reverted the changes to the service folder because I realised that they are out of scope for this task/PR. The only `XXX:`s remaining now are in the `features` - I can't tell if they are used anywhere for automation (it doesn't seem so) but either way worth a look.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 
 - ENV-variable `CREATE_PROGRAM_ENDPOINT` for Kobo-Connect API URL is removed from the Portal.
+- PA-App and all related references (AB#27359)
```
aberonni, line 17: Check whether this is in the right place after release on `10/04/24`

elwinschmitz (approved):

PR #5118, opened 2024-04-09
Feat: remove registration statuses

PeterSmallenbroek:


ashishkumarbasak-accenture:
@Vladimir-Petkov The PR looks nice and working according to description in the task. 
One question, I have seen that there are some [seed data file](https://github.com/global-121/121-platform/blob/main/services/121-service/seed-data/message-template/message-template-nlrc-pv.json#L46) where we have message template with `inclusionEnded`, `rejected` etc. Should we keep them or delete them?

Vladimir-Petkov:
@ashishkumarbasak-accenture I think for now we can keep them, because we using them in Twilio.

services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
```diff
@@ -1,18 +1,31 @@
...
+    await queryRunner.query(`DELETE FROM "121-service"."event" WHERE "registrationId" IN
+      (SELECT "id" FROM "121-service"."registration" WHERE "registrationStatus" = '${registrationStatus}')`);
+
+    await queryRunner.query(`DELETE FROM "121-service".registration_data WHERE "registrationId" IN
+      (SELECT "id" FROM "121-service"."registration" WHERE "registrationStatus" = '${registrationStatus}')`);
 
     await queryRunner.query(
-      `DELETE FROM "121-service"."registration"
-             WHERE "registrationStatus" = 'startedRegistration'`,
+      `DELETE FROM "121-service"."registration" WHERE "registrationStatus" = '${registrationStatus}'`,
```
ashishkumarbasak-accenture, line 25: @aberonni I had to do these changes for this migration.

aberonni: Thanks for sharing. Should we make these changes on a separate PR to get them merged? I wouldn't want to wait for this PR to be merged before fixing a potentially production-disruptive issue.

services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
```diff
@@ -1,18 +1,31 @@
...
+const registrationStatus = 'startedRegistration';
+
 export class RemoveStartedRegistrationState1713363871246
   implements MigrationInterface
 {
   name = 'RemoveStartedRegistrationState1713363871246';
 
   public async up(queryRunner: QueryRunner): Promise<void> {
     console.time('RemoveStartedRegistrationState1713363871246');
+    await queryRunner.commitTransaction();
```
aberonni, line 12: Are `commitTransaction` and `startTransaction` part of the fix?

services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
ashishkumarbasak-accenture: Yes, Sure. Let me create a MR to main then.

services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
ashishkumarbasak-accenture: Since there are multiple queries, so better to do them in a transaction.

services/121-service/migration/1713422095087-migrate-statuses-to-declined.ts
```diff
@@ -0,0 +1,82 @@
...
+        );
+
+        await queryRunner.query(
+          `INSERT INTO "121-service".event_attribute(created, updated, "eventId", key, value) VALUES ($1, $2, $3, $4, $5)`,
+          [
+            'NOW()',
+            'NOW()',
+            insertedId[0].id,
+            'reason',
+            'Status changed to declined by 121 team to simplify registration states',
```
RubenGeo, line 74: I checked the portal and we do not support showing reason for status changes in the portal. (only for data changes).  So this insert is not needed. 
image

services/121-service/migration/1713422095087-migrate-statuses-to-declined.ts
```diff
@@ -0,0 +1,82 @@
...
+          ],
+        );
+      }),
+    );
+
+    console.timeEnd('migrateStatusChanges');
```
RubenGeo, line 80: `console.timeEnd` This is good for testing purposes, if a migration takes longer than a few minutes the App service will not start up. But when we release it, it is better not to add this. I think we accidentally also left this behind in some of the older migrations which is probably why you added it here.

services/121-service/migration/1714467998004-make-fspId-not-null-at-registration.ts
```diff
@@ -0,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class MakeFspIdNotNullAtRegistration1714467998004
+  implements MigrationInterface
+{
```
RubenGeo, line 5: Very nice that you also added fspId is not nullable :)

interfaces/Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -995,8 +998,6 @@ export class ProgramPeopleAffectedComponent implements OnDestroy {
   ) {
     const statusRelatedBulkActions = [
       BulkActionId.include,
```
RubenGeo, line 1000: I think we need to add somewhere to the bulk actions that you can `Decline` registrations. Probably in the payments phase. I created this task for it AB#27875

RubenGeo: I mean this menu: 
image

RubenGeo (changes requested):
Great work!

The 2 things that need to change:
- are that the reason is not needed in the migration
- the `decline` bulk actions need to be added

services/121-service/migration/1713422095087-migrate-statuses-to-declined.ts
ashishkumarbasak-accenture: Removed!

services/121-service/migration/1713422095087-migrate-statuses-to-declined.ts
ashishkumarbasak-accenture: Yes, just copy paste things 😄 
Removed!

interfaces/Portal/src/app/program/program-people-affected/program-people-affected.component.ts
ashishkumarbasak-accenture: Updated the PR!
it was one flag `showIfNoValidation` which was `false` and due to this it was not showing the menu option `Mark as Declined`. I made it `true`. Could you please confirm that its the right change. Thanks!

RubenGeo (approved):

PR #5126, opened 2024-04-09
refactor: Syntax issues in Portal only  AB#27415

elwinschmitz:
These changes are I think save-to-merge already.
The additional changes from #5125 need testing.


interfaces/Portal/src/app/auth/auth.service.ts
```diff
@@ -103,10 +103,17 @@ export class AuthService {
...
     }
+
     if (!user || !user.username || !user.permissions) {
       console.warn('AuthService: No valid user');
       return null;
     }
+
+    if (!user.expires || Date.parse(user.expires) < Date.now()) {
+      console.warn('AuthService: Expired token');
+      return null;
```
PeterSmallenbroek, line 114: This will cause an issue for SSO because then the user doesn't have an 'expires' property
image

PR #5136, opened 2024-04-11
Refactor: remove unnecessary module dependencies from app module

diderikvw:


diderikvw:
All API tests that fail, do succeed on my local dev env. Except for the message time-outs which seem to always fail everywhere.

services/121-service/test/helpers/program.helper.ts
```diff
@@ -353,7 +353,7 @@ export async function waitForMessagesToComplete(
 
     // If not all PAs received a message, wait for a short interval before checking again
     if (!allMessageUpdatesSuccessful) {
-      await waitFor(1000); // Wait for 1 second (adjust as needed)
+      await waitFor(3000); // Wait for 1 second (adjust as needed)
```
elwinschmitz, line 356:
```suggestion
      await waitFor(3_000);
```

elwinschmitz:
To generate the chart, some modules are 'removed', it seems:
https://github.com/global-121/121-platform/blob/refactor.remove-app-module-useless-module-dependencies/services/121-service/src/main.ts#L40-L52

Should that list be updated as well? Or would it give a better picture with these removed?

diderikvw:
@elwinschmitz  AFAIU the reasoning was that removing these from the diagram would make for a better readible diagram. So, I suggest to leave it like it is.

PR #5167, opened 2024-04-16
AB#27576 Get program custom attributes in seperate query

RubenGeo:


services/121-service/src/programs/programs.service.ts
```diff
@@ -82,6 +81,13 @@ export class ProgramService {
       throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
     }
 
+    // Program attributes are queries seperately because the performance is bad when using relations
```
elwinschmitz, line 84:
```suggestion
    // Program attributes are queried separately because the performance is bad when using relations
```

PR #5168, opened 2024-04-17
Remove startedRegistration state

aberonni:
[AB#27475](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27475)

### DIFF
--- a/services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
+++ b/services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
@@ -1,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RemoveStartedRegistrationState1713363871246
+  implements MigrationInterface
+{
+  name = 'RemoveStartedRegistrationState1713363871246';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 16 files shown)


services/121-service/migration/1713363871246-remove-startedRegistration-state.ts
```diff
@@ -0,0 +1,22 @@
...
+{
+  name = 'RemoveStartedRegistrationState1713363871246';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    console.time('RemoveStartedRegistrationState1713363871246');
+
+    await queryRunner.query(
+      `UPDATE "121-service"."registration"
+             SET "registrationStatus" = 'registered'
+             WHERE "registrationStatus" = 'startedRegistration'`,
```
aberonni, line 14: @tijsziere it seems that this change as-is might not be smart, because users in state `startedRegistration` are probably not compatible with `registered` state. Is it safe to change this so that we delete the PAs with`"registrationStatus" = 'startedRegistration'` instead?

aberonni: Confirmed that this is ok to change via DevOps

arsforza (approved):

PR #5178, opened 2024-04-19
AB#27636 add missing tests

Piotrk39:


aberonni (approved):
I didn't thoroughly review this - my thinking is that "some tests" is better than "no tests" and we can iterate on the specifics (like better selectors and things like that) as we go along.

What I did look at is "architecture" (setting up TS, eslint, prettier etc.) and I am happier with this than what we have on main currently.

PR #5179, opened 2024-04-19
Clean TS setup for new `e2e` folder + add eslint/prettier

aberonni:
- Reset tsconfig to avoid weirdness that leads to having to use "ts-no-check"
- Add eslint to e2e folder
- Add prettier to e2e folder
- Import eslint & prettier config from other folders
- Fix existing linting issues in files
- Add linting step for e2e folder in CI


aberonni:
FYI @ashishkumarbasak-accenture I built on your work here

PR #5183, opened 2024-04-22
feat: Remove unused "meetingDocuments"-property from Program(s) AB#27476

elwinschmitz:
This will probably require a proper migration (to get rid of all production-data)...


aberonni:
Continuing this in #5205

PR #5184, opened 2024-04-22
AB#27664 Fix editing paginate query out of function unit

RubenGeo:


services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -63,6 +63,9 @@ export class RegistrationsPaginationService {
     noLimit: boolean,
     queryBuilder?: ScopedQueryBuilder<RegistrationViewEntity>,
   ): Promise<Paginated<RegistrationViewEntity>> {
+    // Deep clone query here to prevent mutation out of this function
+    query = JSON.parse(JSON.stringify(query));
```
elwinschmitz, line 67: Is this the safest/fastest way?

https://medium.com/@pmzubar/why-json-parse-json-stringify-is-a-bad-practice-to-clone-an-object-in-javascript-b28ac5e36521

We already have Lodash available, so we could use `cloneDeep` from that. (see: https://devdocs.io/lodash~4/index#cloneDeep )

PR #5191, opened 2024-04-23
Add husky & simplify eslint tasks

aberonni:
I hear of lint fixes being made on PRs. This seems silly to me! Hopefully, this change will avoid that happening in the future. Let's let our machines do this work for us before we even push the code 🔥 

This PR proposes a few changes:

- Remove explicit `prettier` scripts where `lint` already takes care of it through the `eslint-prettier-plugin` (AKA everywhere we check js & ts files)
- Add `husky` and move `pre-commit` hook to the husky folder. This will run folder-specific lint tasks, and a general prettier task for everything that isn't ts or js
- Remove git hooks folder (and what I suspect to be an unused pre-push commit)
- Update documentation in the root README file to make sure `npm install` is run in the root folder (to install husky)
- Update workflows to ignore husky altogether
- Add a consistent `lint:fix` script which will be run by husky
- Ran `npm run test:doc -- --write` from the root folder to make sure everything that we will test with `lint-staged` is formatted correctly by prettier


.github/workflows/test_docs.yml
```diff
@@ -8,7 +8,10 @@ on:
     branches:
       - main
     paths:
-      - '**/*.md'
+      - '**/*.{md,json,yml,scss,html}'
```
elwinschmitz, line 11: Change the (file-)name of this workflow to be "all the rest of the code" ;)

interfaces/Portal/.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "**/*.{ts,js}": ["eslint --fix"]
```
elwinschmitz, line 2: There are also lint-rules set up for `*.html`-files; So these will be missed

interfaces/Portal/.eslintrc.json
```diff
@@ -13,7 +13,7 @@
         "plugin:@typescript-eslint/recommended",
         "plugin:@angular-eslint/recommended",
         "plugin:@angular-eslint/template/process-inline-templates",
-        "prettier"
+        "plugin:prettier/recommended"
```
elwinschmitz, line 16: Also add in L46 (for HTML)

services/121-service/package.json
```diff
@@ -10,18 +10,11 @@
...
   "main": "index.js",
   "scripts": {
-    "lint": "npm run lint:seed-data && npm run lint:code && npm run lint:syntax",
-    "lint:code": "eslint \"src/**/*.{ts,js}\" \"migration/*.ts\" \"test/**/*.ts\"",
-    "fix:code": "npm run lint:code -- --fix",
-    "lint:seed-data": "prettier --check \"seed-data/**/*.json\"",
-    "fix:seed-data": "npm run lint:seed-data -- --write",
-    "lint:syntax": "prettier --check \"**/*.{md,ts,js,json,yml}\"",
-    "fix:syntax": "npm run fix:code && npm run lint:syntax -- --write",
+    "lint": "eslint \"src/**/*.{ts,js}\" \"migration/*.ts\" \"test/**/*.ts\"",
```
elwinschmitz, line 13: Try `eslint "**/*.{js,ts}"`

package.json
```diff
@@ -21,13 +21,18 @@
     "code:121-service": "code ./services/121-service/",
     "test": "echo 'See README.md#testing \\n\\n' && npm run test:all",
     "test:all": "npx npm-run-all --npm-path npm --print-label --parallel test:docs test:portal test:services",
-    "test:docs": "npx prettier --config=./.prettierrc.yml --check \"**/*.md\"",
+    "test:docs": "npx prettier --config=./.prettierrc.yml --check \"**/*.{md,json,yml,scss,html}\"",
```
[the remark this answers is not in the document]
elwinschmitz: Remove "html" from the list (it should be included already in eslint's scope.

services/mock-service/package.json
```diff
@@ -10,10 +10,7 @@
   },
   "main": "index.js",
   "scripts": {
-    "lint": "npm run lint:code && npm run lint:syntax",
-    "lint:code": "eslint \"src/**/*.{ts,js}\"",
-    "lint:syntax": "prettier --check \"**/*.{md,ts,js,json,yml}\"",
-    "fix": "npm run lint:code -- --fix && npm run lint:syntax -- --write",
+    "lint": "eslint \"src/**/*.{ts,js}\"",
```
elwinschmitz, line 13: Add "fix" (same as 121-service).

services/121-service/package.json
elwinschmitz: Add "fix" -> "npm run lint -- --fix"

package.json
```diff
@@ -21,13 +21,18 @@
...
   },
-  "optionalDependencies": {
-    "prettier": "3.2.5"
+  "devDependencies": {
+    "husky": "^9.0.11",
+    "lint-staged": "^15.2.2",
+    "prettier": "^3.2.5"
+  },
+  "lint-staged": {
+    "*.{md,json,yml,scss,html}": "prettier --config=./.prettierrc.yml --write"
```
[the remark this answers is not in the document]
elwinschmitz: Also remove "html" from here.

interfaces/Portal/package.json
```diff
@@ -83,6 +82,7 @@
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-jasmine": "^4.1.3",
     "eslint-plugin-jsdoc": "^48.2.3",
+    "eslint-plugin-prettier": "^5.1.3",
```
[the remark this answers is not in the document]
aberonni: Looks like they're both necessary after all: https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#installation

package.json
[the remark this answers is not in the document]
aberonni: > npm run test:prettier -- --write?

I tried this but it doesn't work because `lint-staged` needs to pass in its own arguments which get ignored with the extra `--`

.github/workflows/test_prettier.yml
```diff
@@ -1,14 +1,17 @@
...
   workflow_dispatch:
   pull_request:
     branches:
       - main
     paths:
-      - '**/*.md'
+      - '**/*.{md,json,yml,scss}'
+
+env:
+  HUSKY: 0
```
elwinschmitz, line 14: This is the only workflow that really needs this, right? Because this runs `npm install` in the root?

.prettierignore
```diff
@@ -0,0 +1,12 @@
+# Make sure to keep contents aligned with all other interfaces!
+#
```
elwinschmitz, line 2:
```suggestion
```

This warning is no longer relevant when we have only 1 interface. ;)

interfaces/Portal/src/app/pages/dashboard/dashboard.page.html
```diff
@@ -1,4 +1,4 @@
-<app-header [title]="'page.dashboard.pageTitle'|translate"></app-header>
+<app-header [title]="'page.dashboard.pageTitle' | translate"> </app-header>
```
elwinschmitz, line 1: Weird that these auto-format issues wheren't there before...

The Portal already had Prettier v3.2.5... so previous runs of the linting should have highlighted these issues... :?

.prettierignore
```diff
@@ -0,0 +1,12 @@
...
+# See .gitignore
+coverage/
```
elwinschmitz, line 4: Will these ignore-rules now be taken into account when you open the Portal-only in VSCode?

So if you run "`npm test`" in the Portal-folder (at least once), the `coverage`-folder should be created... Subsequent runs of Prettier (in the Portal-folder) should not check these files then...

interfaces/Portal/src/app/pages/dashboard/dashboard.page.html
elwinschmitz: Its sort-of the opposite of what happend on the `index.html`... :?

.github/workflows/test_prettier.yml
aberonni: Exactamundo

.prettierignore
aberonni: Prettier won't run in the Portal-folder - eslint (with `npm run lint`) is the only thing being run on a sub-folder level (which already ignores the coverage folder).

The IDE might still "prettify" these files if you save them, but that isn't an issue IMHO because you aren't committing them anyway.

interfaces/Portal/.prettierignore
```diff
@@ -1,24 +1,5 @@
...
-# Exclude (generated) package files
-package*.json
-
-
 # Don't reorder imports for polyfills:
 src/polyfills.ts
 
-
 # Exception for inline code/styles
 src/index.html
```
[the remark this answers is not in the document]
aberonni: Indeed, this isn't being applied but it's because eslint is applying the prettier rules to js,ts,HTML files now, which doesn't look at `prettierignore`. 

I ended up just removing this file instead, because I don't think it was relevant anymore after this refactor.

interfaces/Portal/src/polyfills.ts
```diff
@@ -1,55 +1,55 @@
...
-/** *************************************************************************************************
- * Zone JS is required by default for Angular itself.
- */
-
-import 'zone.js';
-
-/** *************************************************************************************************
- * APPLICATION IMPORTS
- */
+/**
```
aberonni, line 1: I don't understand what changed here 😅

.prettierignore
elwinschmitz: Ah, yes.. makes sense. ESLint will be in charge of most things anyway indeed.

Prettier is just a "sub-contractor", so it won't have "the last say" anywhere...

interfaces/Portal/src/polyfills.ts
elwinschmitz: Probably the line-endings...
They where probably changed by a windows-user without the proper config for git (or something similar)..

And because this file was ignored before, it never got touched/fixed...

But I think the main reason this file was ignored.. (that the "order of imports" mattered)... isn't relevant anymore.. (as the current order is preserved).

interfaces/Portal/src/app/pages/dashboard/dashboard.page.html
elwinschmitz: Probably a side-effect of the mismatch in config-for-eslint by the missing "eslint-plugin-config-prettier" (or the other one).

elwinschmitz (approved):
💥

PR #5192, opened 2024-04-23
Refactor short label

aberonni:
[AB#27477](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27477)


services/121-service/seed-data/fsp/fsp-africas-talking.json
```diff
@@ -8,21 +8,19 @@
     {
```
aberonni, line 8: All of the changes within the `services/121-service/seed-data` folder were applied through a custom script (so no manual work was involved).

RubenGeo: Nice! There are also a lot of formatting changes in those json files. For me these are automatically edited on save if I open a vscode window from the `121-service` folder.

You can also run `npm run test:prettier -- --write` from the root folder. Maybe we should have something that `npm run fix` also runs that command

RubenGeo (changes requested):
Needs more prettier linter magic, to prevent unneeded changes

services/121-service/seed-data/fsp/fsp-africas-talking.json
aberonni: Good that you mention it! These should be picked up automatically, and CI should complain about them if they aren't correct. I thought I address this in https://github.com/global-121/121-platform/pull/5191 but evidently something has slipped through the cracks. Will take a look and report back ✌️

aberonni:
@RubenGeo I pushed the linting fixes. Looks like the CI job is not triggering when it is supposed to. Will address that in a separate PR.

RubenGeo (approved):
Amazing!

PR #5195, opened 2024-04-24
fix: allow registration with empty phoneNumber AB#27722

Vladimir-Petkov:


services/121-service/src/registration/validators/registrations-input-validator.ts
```diff
@@ -141,6 +141,10 @@ export class RegistrationsInputValidator {
       if (errorObj) {
         errors.push(errorObj);
       }
+
+      if (row.phoneNumber === undefined && program.allowEmptyPhoneNumber) {
+        row.phoneNumber = '';
+      }
```
RubenGeo, line 147: if (row.phoneNumber === undefined && program.allowEmptyPhoneNumber) {
        importRecord.phoneNumber = '';
      } else {
        importRecord.phoneNumber = row.phoneNumber;
      }

RubenGeo: I think this is a bit more readable. Otherwise good find :)

RubenGeo (dismissed):

services/121-service/src/registration/validators/registrations-input-validator.ts
Vladimir-Petkov: @RubenGeo  This will not work, because you have to use row value in 
image

RubenGeo:
I made an extra commit here: https://github.com/global-121/121-platform/pull/5209

PR #5197, opened 2024-04-25
Test.view activity overview

Piotrk39:


e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -22,6 +22,11 @@ class RegistrationDetails {
   financialServiceProviderDropdown = 'app-update-fsp #select-label';
   debitCardPaTable = 'ion-card-title';
   debitCardStatus = 'ion-label';
+  tabButton = 'ion-button';
+  historyTile = 'article';
+  historyTileText = 'span';
+  historyTileTimeStamp = 'time';
+  tileInformationPlaceHolder = 'strong';
```
aberonni, line 29: Still not a fan of the very generic selectors we are using :D

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
```diff
@@ -0,0 +1,119 @@
+import HomePage from '@121-e2e/pages/Home/HomePage';
+import LoginPage from '@121-e2e/pages/Login/LoginPage';
+import RegistrationDetails from '@121-e2e/pages/RegistrationDetails/RegistrationDetailsPage';
```
aberonni, line 3:
```suggestion
import RegistrationDetailsPage from '@121-e2e/pages/RegistrationDetails/RegistrationDetailsPage';
```

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+let accessToken: string;
+
+test.beforeEach(async ({ page }) => {
+  // Set WhatsApp-number for ALL tests in this suite only
+  const registrationVisa = {
+    ...registrationVisaDefault,
+    whatsappPhoneNumber: registrationVisaDefault.phoneNumber,
+  };
+
+  await resetDB(SeedScript.nlrcMultiple);
```
aberonni, line 42: Does it make sense to do this in a global setup? I imagine we do this for all test suites?

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+  accessToken = await getAccessToken();
+  await waitFor(2_000);
```
aberonni, line 44: Is the underscore a typo? Also, what are we waiting for? Maybe a comment would help

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+  const registration = new RegistrationDetails(page);
+  const homePage = new HomePage(page);
+
+  await test.step('Should navigate to PA profile page in Payment table', async () => {
+    await homePage.navigateToProgramme('NLRC OCW program');
+    await table.selectTable('Payment');
+    await table.clickOnPaNumber(1);
+  });
+
+  await test.step('Validate Status histor tab on PA Activity Overview table', async () => {
```
aberonni, line 102:
```suggestion
  await test.step('Validate the "Status history" tab on the PA Activity Overview table', async () => {
```

aberonni (approved):
RE: selectors, I think it would be wise to address this sooner rather than later, in a systemic way.

My suggested approach would be to
- reach an agreement on how we (can) do selectors
- clean up the existing test cases 
- make sure new test cases follow the agreement

At the very least, we should follow the best practice of [using locators](https://playwright.dev/docs/best-practices#use-locators).

Ideally, we would do this before we go too crazy adding loads more selectors because every selector added is a line of tech debt we eventually have to pay back. I don't want to block your work for this, but I'd like to shed some light on this debt so we can prioritise paying it back.

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
Piotrk39: I'll definitely drop this ones as they make no sense anymore and probably will cause more harm in the future

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
Piotrk39: I thought about it. But then we have some differences in the setup for each test case and basically just resetDB is kind of a global setup. But we should look into it.

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverview.spec.ts
Piotrk39: This one is taken from API test suite and because of asynchronous methods there are this hardcoded waits. Honestly I don't like them and tried to remove them in one test but it causes some errors.

aberonni (approved):

PR #5198, opened 2024-04-25
fix: registrations not allowed without csv import if program is not published yet AB#27479

ashishkumarbasak-accenture:
fix: registrations not allowed without csv import if program is not published yet [AB#27479](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27479)

Task: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/27479

### DIFF
--- a/interfaces/Portal/src/app/mocks/api.programs.mock.ts
+++ b/interfaces/Portal/src/app/mocks/api.programs.mock.ts
@@ -22,7 +22,7 @@ const programsArray: Program[] = [
     phase: ProgramPhase.design,
     validation: false,
     author: {},
-    published: false,
+    published: true,
     notifications: { en: 'Notification text' },
     programCustomAttributes: [
       {
--- a/services/121-service/seed-data/program/program-demo.json
+++ b/services/121-service/seed-data/program/program-demo.json
@@ -1,5 +1,5 @@
 {
-  "published": false,
+  "published": true,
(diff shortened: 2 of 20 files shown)


Vladimir-Petkov:
@ashishkumarbasak-accenture I think you don`t have to change seed-data programs.

Vladimir-Petkov (dismissed):

RubenGeo (changes requested):
- Functionally works as expected :)
- I discussed with Tijs that it is no problem if we set all program json files to published = true (than you revert most of the api test changes)
- I think it makes sense to also cover your error in an API test or a unit test:
  - An api test could be added for example to test/registrations/import-registration.test.ts. Where you unpublish a program and that see if you get the 404 error

ashishkumarbasak-accenture:
> * Functionally works as expected :)
> * I discussed with Tijs that it is no problem if we set all program json files to published = true (than you have to revert most of the api test changes unfortunately)
> * I think it makes sense to also cover your error in an API test or a unit test:
>   
>   * An api test could be added for example to test/registrations/import-registration.test.ts. Where you unpublish a program and that see if you get the 404 error

Updated the MR accordingly, Thanks for your review! 😊

RubenGeo (approved):
Thanks!

PR #5204, opened 2024-04-26
Disable spam comments from Azure on PRs

aberonni:
These comments have been polluting the PRs without (IMHO) adding much value, given that the deployment status information is already visible elsewhere on the GitHub UI.

Sadly, the Azure team is aware of this issue but isn't prioritising a fix for it: https://github.com/Azure/static-web-apps/issues/1135

The proposed solution seems hacky, but, if it works, it works.

### DIFF
--- a/.github/workflows/deploy_test_portal.yml
+++ b/.github/workflows/deploy_test_portal.yml
@@ -7,7 +7,9 @@ env:
 
 permissions:
   actions: read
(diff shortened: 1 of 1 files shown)


elwinschmitz (approved):
I like these comments sometimes... (so that's why I'd leave them on.)

But I hate them more often. ;)

PR #5206, opened 2024-04-26
Remove unnecessary `waitFor`s in API tests

aberonni:
## Describe your changes

I attempted to speed up the tests by getting rid of useless `waitFor`s wherever possible. In some places, I couldn't get rid of them easily so I just left them, but in most places the tests seem to be working fine without the pauses.

While working on this, I ended up also adding a fancy CI jest reporter that shows test results with pretty colours and collapsed test results for Passed tests:

image

## Speed comparison

Before:
image

After:
image

gif


services/121-service/package.json
```diff
@@ -18,9 +18,11 @@
     "seed:dev": "ts-node src/scripts seed-dev",
     "test": "echo 'No automated tests. Run `npm run test:e2e:all` to run API-tests locally. Run `npm run test:unit:all` to run unit-tests locally.' && exit 0",
     "test:unit:all": "jest --runInBand --config=./jest.unit.config.js",
+    "test:unit:ci": "FORCE_COLOR=1 npm run test:unit:all -- --reporters='@matteoh2o1999/github-actions-jest-reporter'",
     "test:unit:watch": "npm run test:unit:all -- --watchAll",
     "test:unit:coverage": "npm run test:unit:all -- --coverage --coverageDirectory=coverage/unit",
     "test:e2e:all": "node --expose-gc node_modules/.bin/jest --config=jest.e2e.config.js --runInBand --detectOpenHandles --logHeapUsage",
+    "test:e2e:ci": "FORCE_COLOR=1 npm run test:e2e:all -- --reporters='@matteoh2o1999/github-actions-jest-reporter'",
```
elwinschmitz, line 25: Apparently, judging by the screenshot, the "heap size" (i.e. memory use) isn't shown anymore for each test/scenario.
Is there a way to configure this reporter more?

And also, would it be possible to have it also working when running locally? (So to not have to many differences in the 'dev v.s. test' workflow/environments...

aberonni: > Apparently, judging by the screenshot, the "heap size" (i.e. memory use) isn't shown anymore for each test/scenario.
Is there a way to configure this reporter more?

Indeed, we seem to have lost that. I didn't notice. The reporter doesn't have any configuration options. Is this something we need/use in CI?

> And also, would it be possible to have it also working when running locally? (So to not have to many differences in the 'dev v.s. test' workflow/environments...

The reporter? We could, but it wouldn't make much sense given that it was built for CI (and the setup I've included is what is recommended by the author).

aberonni:
I'm actually closing this because I've noticed some additional flakiness, and it probably isn't worth the tradeoff.

Longer term I'd like to look into the use of something like [testcontainers](https://testcontainers.com/) as a more stable solution.

services/121-service/package.json
elwinschmitz: Well, the heap-size isn't used "on a daily basis", but I think it is a good indicator of "badly written tests" ;)
So sometimes you can debug an issue based on "where the memory went".

> ... dev v.s CI

Some improvements to " make automated processes as smooth as possible" will improve both the 'development experience' AND the CI-flows...
As any developer will be running these tests locally mostly in an automated way (very similar to on GitHub)...

elwinschmitz:
> ...testcontainers

That looks promising; our "mock-service" is in a sense just that, right? ;)

There is also a thing in GitHub Actions flow called "service containers"(https://docs.github.com/en/actions/using-containerized-services/about-service-containers);
This is something I still want to look into.

It would work similar to our current set-up (as that the DB/Queue is a separate Docker-service); but hopefully with a more streamlined set of resources and better performance/speed...

aberonni:
> That looks promising; our "mock-service" is in a sense just that, right? ;)

I'm not full aware of what the "mock-service" does but I don't think so. It essentially would replace our manual starting of services prior to running of tests, and instead spin up (a) service(s) for you ad-hoc at the beginning of a test. That would mean that each test would have its own instance of db etc. thus enabling parallel testing.

PR #5208, opened 2024-04-29
Release/v1.116.12

PeterSmallenbroek:


RubenGeo (dismissed):

RubenGeo:
Peter also approved this, but it was not good enough for some reason: 
image

RubenGeo:
I think we need to re-evaluate our pulls request strategy on hotfixes, since there is now not really a PR to review before it goes to production. Furthermore when you make PRs to main before you create the release/tag you risk that we accidentally update it with main

Maybe we can do something like:
1. Create a new release branch from the previous tag. (e.g from tag v1.116.11 create  branch release/v1.116.12)
2. Push this to github
3. Checkout locally to a new fix. branch. do any fixes you want in a fix. branch
4. Create a PR from you fix. branch to the release/.. branch 
5. Have someone review an merge it
6. Create tag and release using github actions
7. Create PR from release branch to main (this does not seem great because than someone has to review it again..)

This flow does seem a bit over-complicated tough, so maybe we can come up with something easier?

@aberonni Do you have suggestions?

aberonni:
@RubenGeo tbh, addressing the release process is high on my list of "things I wanna look at" right after "how we do git". Maybe your comments will push it higher up the list :)

If I broaden the scope a bit towards "how we do releases", I would like to challenge the current process and suggest we move to CI/CD rather than make our lives harder with hotfixes and per-instance versioning. It is not a good practice to code according to the idea that "we should release a hotfix because we can't trust what is on main", and this concept is at the core of how we currently treat hotfixes. A shift towards CI/CD would make any conversation on "how we do hotfixes according to the current process" redundant. But I'm sure there are intricacies I'm not aware of so we probably can't just make that change overnight.

It would probably be good for me to get a better grasp of how we currently operate and do things before making any drastic suggestions, though. Maybe it would be easier for me to jump on a call and let you explain that to me if you have the time for it :)

PS I'd be happy to approve this based on trust if that helps unblock you in the short term, but the conflict on `CHANGELOG.md` needs to be resolved first (because my review will be dismissed if you make significant changes to the branch).

RubenGeo:
Sure call me anytime ;)

RubenGeo (approved):

elwinschmitz:
The changes from this branch are already in #5212

PR #5210, opened 2024-04-29
Test.view activity overview messages tab

Piotrk39:
## Describe your changes

New automated E2E test for : View Activity overview “Messages tab" based on feature file and Test Case: https://dev.azure.com/redcrossnl/121%20Platform/_testPlans/define?planId=27408&suiteId=27410

- [ ] I have communicated above whether I need any deviation from our PR guidelines (eg. "I don't want the reviewer to merge on my behalf because...")


e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverviewMessageTab.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+import {
+  awaitChangePaStatus,
+  importRegistrations,
+} from '@121-service/test/helpers/registration.helper';
+import {
+  getAccessToken,
+  resetDB,
+} from '@121-service/test/helpers/utility.helper';
+import { test } from '@playwright/test';
+import data from '../../../../interfaces/Portal/src/assets/i18n/en.json';
```
RubenGeo, line 30: Great that you are using this import now. I do think it can benefit from a more descriptive name so maybe something like:

```suggestion
import englishTranslations from '../../../../interfaces/Portal/src/assets/i18n/en.json';
```
or just translations

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverviewMessageTab.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+  );
+
+  await waitForPaymentTransactionsToComplete(
+    programIdVisa,
+    paymentReferenceIds,
+    accessToken,
+    3001,
+    Object.values(StatusEnum),
+  );
+
```
RubenGeo, line 78: - I think I heard you mention during our call that this setup part needs to happen in a lot of test cases
- We also ran into this same code duplication problem in our API tests ;)
- Therefore we created some helper methods: you could replace this setup part with the seedPaidRegistrations or the seedIncludedRegistrations method from services/121-service/test/helpers/registration.helper.ts
- I think this can also be applied to some other playwright test cases

e2e/tests/121-Portal/ViewPaProfilePage/OpenProfilePage.spec.ts
```diff
@@ -47,9 +48,8 @@ test('[27411] Open PA profile page', async ({ page }) => {
       firstName: undefined,
       lastName: undefined,
       phoneNumber: undefined,
-      status: 'Registered',
+      status: undefined,
```
RubenGeo, line 51: Why is this undefined here? I think this could refer to the en.json on this line: https://github.com/global-121/121-platform/blob/0c957878576866bd5706817d7dc6c6850842226c/interfaces/Portal/src/assets/i18n/en.json#L66. Is that possible?

RubenGeo (changes requested):
@Piotrk39 I think these playwright test are really great! I hope I do not discourage you with some more comments, but I do think it is valuable to make some additional changes

e2e/tests/121-Portal/ViewPaProfilePage/OpenProfilePage.spec.ts
Piotrk39: I did it on purpose here to avoid conflicts since I wanted to check only the first cell of the row to save time as main objective of the test does not refer to check the table. When you put undefined then function does not search for the value. But maybe it can be done better.

Piotrk39: I'm open for suggestions ;)

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverviewMessageTab.spec.ts
Piotrk39: Thank you for your comment and suggestions the more of those the better for the quality of the project.

e2e/tests/121-Portal/ViewPaProfilePage/ViewActivityOverviewMessageTab.spec.ts
Piotrk39: I'll take a fix it is indeed very valuable to not repeat same thing over and over again.

Piotrk39: I have fixed this this and I'd like to address more improvements like this in a separate PR not to bloat this one with too many changes. Thanks a lot.

e2e/tests/121-Portal/ViewPaProfilePage/OpenProfilePage.spec.ts
```diff
@@ -38,7 +39,7 @@ test('[27411] Open PA profile page', async ({ page }) => {
 
   await test.step('Should display correct amount of runnig projects and open PAs for registration', async () => {
     await homePage.validateNumberOfActivePrograms(2);
-    await homePage.openPAsForRegistrationOcwProgram('NLRC OCW program');
+    await homePage.openPAsForRegistrationOcwProgram(NLRCProgram.titlePaApp.en);
   });
 
   await test.step('Should validate first row with uploaded PAs', async () => {
```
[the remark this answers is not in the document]
Piotrk39: I hope this table seizes to exist in the future in general... Kind of work in progress here with the table it unfortunately changes shape and position depending on the screen size used. I could change it but I chose the name to define and understand on which table I am verifying the information the left (where we have the PA number and name usually) or right (where in most circumstances there is a language ). I'll change the name tho.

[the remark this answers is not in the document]
Piotrk39: Renamed to 'verifiyProfilePersonalnformationTable' left and right but I'l address it in the future. Thanks tho correct naming is important.

e2e/tests/121-Portal/ViewPaProfilePage/OpenProfilePage.spec.ts
Piotrk39: But I also skipped this part because it is not relevant for the test. I mapped the table and added this validation for the future use and it most certainly can become much better.

PR #5222, opened 2024-04-30
fix: fixed the migration related to startedRegistration status change

ashishkumarbasak-accenture:
fix: fixed the migration related to startedRegistration status change which was failing against production data


aberonni (approved):
Thanks Ashish 🙏

PR #5224, opened 2024-04-30
Adding e2e tests in CI

aberonni:
## Describe your changes

Adding e2e tests in CI


.github/workflows/test_e2e.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Run Services with Docker
+        working-directory: ./services
+        run: |
+          docker --log-level 'warn' compose -f docker-compose.yml up -d --quiet-pull
+
+      - name: Start Portal
+        working-directory: ./interfaces/Portal
+        run: |
+          npm install
+          npm start -- --port 8888 > /dev/null 2>&1 &
```
elwinschmitz, line 37: I'd say that using a "production mode"-version of the Portal gives more realistic results;
So using:

```suggestion
          npm run start:debug-production > /dev/null 2>&1 &
```

This will then run on port `8088`, so this would need to be adjusted below in the health-check.

e2e/playwright.config.ts
```diff
@@ -48,12 +48,12 @@ export default defineConfig({
...
     video: 'on-first-retry',
     screenshot: 'only-on-failure',
-    headless: false,
+    headless: true,
     acceptDownloads: true,
     actionTimeout: 20000,
     launchOptions: {
       downloadsPath: 'resources/downloads',
-      args: ['--start-maximized'],
+      args: ['--window-size=1920,1024'],
```
aberonni, line 56: FYI this works in CI so this way we can be sure we are always testing with the same resolution

Piotrk39 (approved):
We did review during the call. All good.

PR #5240, opened 2024-05-03
Add retry to flaky API test

aberonni:
This test keeps failing the whole API integration test job in CI. And then it works fine on a re-run. I don't have enough context to actually fix the test, but this at least should avoid this routine annoyance and waste of time :)


Piotrk39:
It is better to rerun one test then the whole suite. But is there an option to set it globally for all the tests. In case of failure re-run?

Piotrk39 (approved):
As per comment. Maybe we should set it as a rule for all the tests. But if current suite does not provide this option I see this as sufficient solution. Good job.

aberonni:
@Piotrk39 we could do it for the whole suite, but I think it's better per-test (and used sparingly) because then it forces us to write more reliable tests.

PR #5247, opened 2024-05-06
AB#27848  Created pending migration and fix event entity

RubenGeo:
AB#27848

## Describe your changes
Create pending migrations and 

- [ ] I have communicated above whether I need any deviation from our PR guidelines (eg. "I don't want the reviewer to merge on my behalf because...")


services/121-service/src/base-audited.entity.ts
```diff
@@ -5,3 +5,8 @@ export class Base121AuditedEntity extends Base121Entity {
   @Column()
   public userId: number;
 }
+
+export class Base121OptionalAuditedEntity extends Base121Entity {
```
aberonni, line 9: I don't understand this change. Maybe it would be a nice idea to add a comment here explaining what is going on and/or why this is needed.

aberonni (approved):
Awesome job! I did a quick test and could see that no new migration files were generated. I'm not sure if you needed me to do any other testing, otherwise feel free to merge when you're happy with it. (The extra comment that I mention in the review would be nice, but I'm not sure it's actually necessary for those with context of what is going on).

aberonni (approved):

RubenGeo:
@aberonni  Thanks for the review. I do not think more additional testing is needed than what I described in the subtask in devops

PR #5264, opened 2024-05-08
AB#27976 Better validation for message sending

RubenGeo:
[AB#27976](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27976) 

## Describe your changes
Better validation for sending messages (on status change and custom message). This is to prevent that both a message and a messageTemplateKey are added to the body. 

### DIFF
--- a/services/121-service/src/registration/dto/registration-status-patch.dto.ts
+++ b/services/121-service/src/registration/dto/registration-status-patch.dto.ts
@@ -11,7 +11,7 @@ export class RegistrationStatusPatchDto {
   public readonly status: RegistrationStatusEnum;
 
   @ApiProperty({ example: 'Long enough rejection message', required: false })
-  @IsNotBothPresent('messageTemplateKey')
+  @IsNotBothPresent<RegistrationStatusPatchDto>('messageTemplateKey')
   public readonly message: string;
 
   @ApiProperty({
--- a/services/121-service/src/registration/dto/registration-status-patch.dto.ts
+++ b/services/121-service/src/registration/dto/registration-status-patch.dto.ts
@@ -1,6 +1,7 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsEnum, IsOptional, IsString, MinLength } from 'class-validator';
+import { IsEnum, IsOptional } from 'class-validator';
 import { RegistrationStatusEnum } from '../enum/registration-status.enum';
+import { IsNotBothPresent } from '../validators/is-not-both-present.class.validator';
 
 export class RegistrationStatusPatchDto {
   @ApiProperty({
@@ -10,9 +11,7 @@ export class RegistrationStatusPatchDto {
   public readonly status: RegistrationStatusEnum;
 
   @ApiProperty({ example: 'Long enough rejection message', required: false })
-  @MinLength(1)
-  @IsString()
-  @IsOptional()
+  @IsNotBothPresent('messageTemplateKey')
   public readonly message: string;
 
   @ApiProperty({
--- a/services/121-service/src/registration/dto/send-custom-text.dto.ts
+++ b/services/121-service/src/registration/dto/send-custom-text.dto.ts
@@ -7,8 +7,8 @@ export class SendCustomTextDto {
   @ApiProperty({
     example: 'Your voucher can be picked up at the location',
   })
-  @IsNotBothPresent('messageTemplateKey')
-  @IsNotBothEmpty('messageTemplateKey')
+  @IsNotBothPresent<SendCustomTextDto>('messageTemplateKey')
+  @IsNotBothEmpty<SendCustomTextDto>('messageTemplateKey')
   @ValidateIf((o) => !o.skipMessageValidation)
   public readonly message: string;
 
--- a/services/121-service/src/registration/dto/send-custom-text.dto.ts
+++ b/services/121-service/src/registration/dto/send-custom-text.dto.ts
@@ -1,26 +1,21 @@
 import { ApiProperty } from '@nestjs/swagger';
-import {
-  IsBoolean,
-  IsOptional,
-  IsString,
-  MinLength,
-  ValidateIf,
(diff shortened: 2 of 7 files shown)


services/121-service/src/registration/dto/registration-status-patch.dto.ts
```diff
@@ -10,9 +11,7 @@ export class RegistrationStatusPatchDto {
   public readonly status: RegistrationStatusEnum;
 
   @ApiProperty({ example: 'Long enough rejection message', required: false })
-  @MinLength(1)
-  @IsString()
-  @IsOptional()
+  @IsNotBothPresent('messageTemplateKey')
```
aberonni, line 14: I'm not sure about this, but I think it mostly has to do with the API of the new validator. Would it be possible to build something that explicitly includes both properties?

eg. `  @IsNotBothPresent('message', 'messageTemplateKey')`

I wonder whether this is possible on the DTO itself rather than on a single property, because for me it is weird/unexpected/unintuitive that you only specify a validator on `message`, and you leave `messageTemplateKey` as-is.

services/121-service/src/registration/dto/send-custom-text.dto.ts
```diff
@@ -1,26 +1,21 @@
...
 
 export class SendCustomTextDto {
   @ApiProperty({
     example: 'Your voucher can be picked up at the location',
   })
-  @IsString()
-  @MinLength(1)
-  @ValidateIf((o) => !o.skipMessageValidation && !o.messageTemplateKey)
+  @IsNotBothPresent('messageTemplateKey')
+  @IsNotBothEmpty('messageTemplateKey')
```
aberonni, line 11: Rather than using two validators, maybe it would be better to have something like `OnlyOneIsPresent` that combines the two. Ideally this somehow reuses `IsNotBothPresent`

services/121-service/src/registration/dto/registration-status-patch.dto.ts
```diff
@@ -1,6 +1,7 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsEnum, IsOptional, IsString, MinLength } from 'class-validator';
+import { IsEnum, IsOptional } from 'class-validator';
 import { RegistrationStatusEnum } from '../enum/registration-status.enum';
+import { IsNotBothPresent } from '../validators/is-not-both-present.class.validator';
 
 export class RegistrationStatusPatchDto {
```
aberonni, line 6: I think the addition of these validators to the DTOs should be covered by unit tests to make sure we don't regress.

Something like
- it should throw error if both are defined
- it should throw error if neither are defined
- it should work fine if only message is defined
- it should work fine if only messageTemplateKey is defined

aberonni (changes requested):
This is a great idea. What happens if an interface passes in both `message` and `messageTemplateKey`? I'm thinking we should fail gracefully but I don't know enough about validators to know what would happen in that scenario.

services/121-service/src/registration/dto/registration-status-patch.dto.ts
RubenGeo: That would be nicer I tried to find if there is a good option for it but I don't think there is yet: https://github.com/typestack/class-validator/issues/182

Another options would be to just add this validation in the controller: 

image

Doing it in the DTO is a bit nice because that is where we mostly define the validation of the DTO that comes in. But doing it in the controller is maybe a bit more easy for anyone to understand

services/121-service/src/registration/dto/send-custom-text.dto.ts
RubenGeo: Is see your point. But for decorators seperating them is a bit better because it fits more with the style of how the default decorators are setup.   

`@IsOptional()
 @IsBoolean()`

Could also be combined in `IsOptionalBoolean`

services/121-service/src/registration/dto/registration-status-patch.dto.ts
RubenGeo: Good point. I'll make it depending on this thread: https://github.com/global-121/121-platform/pull/5264#discussion_r1593784931

services/121-service/src/registration/dto/send-custom-text.dto.ts
aberonni: I see your point actually. Let's leave this as-is

services/121-service/src/registration/dto/registration-status-patch.dto.ts
aberonni: I think I'd prefer the cleaner solution even if it isn't as comprehensive as the one that you've built. The issue that you've referenced above is an example of what we're lacking here indeed. What would be the downside/risk of doing it in the controller?

RubenGeo: Validation of the dto object happens in 2 places

aberonni: Had a call about it - decided to go the controller route in this instance.

RubenGeo: I decided to go the dto route in the end ;)

aberonni (approved):
I like it!

services/121-service/src/registration/validators/is-not-both-empty.class.validator.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { validate } from 'class-validator';
+import { IsNotBothEmpty } from './is-not-both-empty.class.validator';
+
+class TestClass {
+  @IsNotBothEmpty<TestClass>('testOtherProperty')
```
mirceah99, line 5: This is not a good idea because you will want to mark optional properties with @IsOptional() anyway, so your validator will be skipped, which can end in both properties being null/undefined.

PR #5267, opened 2024-05-08
Test.successfully add note

Piotrk39:
A test case for adding a new note in Activity overview of PA

### DIFF
--- a/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
+++ b/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
@@ -25,19 +25,11 @@ class RegistrationDetails {
 
   constructor(page: Page) {
     this.page = page;
-<<<<<<< HEAD
     this.personalInformationTable = this.page.getByTestId(
       'registration-personal-information-data',
     );
     this.personAffectedName = this.page.getByTestId(
       'registration-details-name',
-=======
-    this.personalInformationTable = this.page.locator(
-      '[data-testid="personal-information-data"]',
-    );
-    this.personAffectedName = this.page.locator(
-      '[data-testid="registration-details-name"]',
->>>>>>> 759e0f73b (AB#27896 add test ids, refactor functions, add translations)
     );
     this.personAffectedStatus = this.page.locator('ion-label:text("Status")');
     this.personAffectedStatusDateUpdate = this.page.locator(
@@ -52,30 +44,15 @@ class RegistrationDetails {
     this.financialServiceProvider = this.page.locator(
       'ion-item:has(ion-label:has-text("Financial")) ion-label strong',
     );
-<<<<<<< HEAD
     this.showAllButton = this.page.getByTestId('show-all-button');
-=======
-    this.showAllButton = this.page.locator('[data-testid="show-all-button"]');
->>>>>>> 759e0f73b (AB#27896 add test ids, refactor functions, add translations)
     this.editPersonAffectedPopUp = this.page.locator(
       'app-edit-person-affected-popup',
     );
     this.financialServiceProviderDropdown = this.page.locator(
       'app-update-fsp #select-label',
     );
-<<<<<<< HEAD
     this.debitCardPaTable = this.page.getByTestId(
       'physical-cards-overview-title',
-=======
-    this.debitCardPaTable = this.page.locator(
-      '[data-testid="physical-cards-overview-title"]',
-    );
-    this.debitCardStatus = this.page.locator(
-      '[data-testid="card-status-chip"]',
-    );
-    this.tabButton = this.page.locator(
-      '[data-testid="activity-detail-tab-button"]',
->>>>>>> 759e0f73b (AB#27896 add test ids, refactor functions, add translations)
     );
     this.debitCardStatus = this.page.getByTestId('card-status-chip');
     this.tabButton = this.page.getByTestId(
--- a/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
+++ b/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
@@ -258,13 +258,14 @@ class RegistrationDetails {
     }
   }
 
-  async validateNewNoteTile(
+  async validateNoteTile(
+    noteIndex: number,
     changeTitle: string,
     userName: string,
     date: string,
     noteContent: string,
   ) {
-    const historyTile = this.historyTile.nth(0);
+    const historyTile = this.historyTile.nth(noteIndex);
 
     await expect(historyTile).toBeVisible();
     expect(
--- a/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
+++ b/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
@@ -247,15 +247,16 @@ class RegistrationDetails {
   async writeNote({
     placeholder,
     note,
-    buttonName
+    buttonName,
   }: {
-    placeholder: string,
-    note: string,
-    buttonName: string
+    placeholder: string;
+    note: string;
+    buttonName: string;
   }) {
     const okButton = this.page.getByRole('button', {
       name: buttonName,
     });
+    await expect(okButton).toBeDisabled();
     await this.page.getByPlaceholder(placeholder).fill(note);
     await this.page.waitForLoadState('networkidle');
     await expect(okButton).toBeEnabled();
--- a/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
+++ b/e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
@@ -244,7 +244,15 @@ class RegistrationDetails {
     await this.page.getByText(addNote).click();
   }
 
-  async writeNote(placeholder: string, note: string, buttonName: string) {
+  async writeNote({
+    placeholder,
+    note,
+    buttonName
+  }: {
+    placeholder: string,
+    note: string,
+    buttonName: string
+  }) {
     const okButton = this.page.getByRole('button', {
       name: buttonName,
     });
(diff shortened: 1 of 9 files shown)


e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,54 @@ class RegistrationDetails {
...
+    await this.page.getByText(addNote).click();
+  }
+
+  async writeNote(placeholder: string, note: string, buttonName: string) {
+    const okButton = this.page.getByRole('button', {
+      name: buttonName,
+    });
+    await this.page.getByPlaceholder(placeholder).fill(note);
+    await this.page.waitForLoadState('networkidle');
+    await expect(okButton).toBeEnabled();
```
aberonni, line 253: Would it be good to check `toBeDisabled` before you fill the note?

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,54 @@ class RegistrationDetails {
...
+
+    for (let i = 0; i < 2; i++) {
```
aberonni, line 255: I like the idea of the for loop to avoid repetition, but I think maybe there could be some value in manually repeating, and adding a comment with each repetition explaining what is happening. Because I look at this and I wonder "why are we doing the same thing twice?

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,54 @@ class RegistrationDetails {
...
+    }
+  }
+
+  async validateNewNoteTile(
+    changeTitle: string,
+    userName: string,
+    date: string,
+    noteContent: string,
+  ) {
+    const historyTile = this.historyTile.nth(0);
```
aberonni, line 267: Maybe this could be slightly more generic at little extra cost:

```suggestion
  async validateNoteTile(
    noteIndex: number,
    changeTitle: string,
    userName: string,
    date: string,
    noteContent: string,
  ) {
    const historyTile = this.historyTile.nth(noteIndex);
```

e2e/tests/121-Portal/ViewPaProfilePage/SuccessfullyAddNote.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+});
+
+test('[27499] Successfully add note', async ({ page }) => {
+  const table = new TableModule(page);
+  const registration = new RegistrationDetails(page);
+  const homePage = new HomePage(page);
+  const helpers = new Helpers();
+
+  await test.step('Should display correct amount of runnig projects and navigate to PA table', async () => {
```
aberonni, line 36:
```suggestion
  await test.step('Should display correct amount of running projects and navigate to PA table', async () => {
```

e2e/tests/121-Portal/ViewPaProfilePage/SuccessfullyAddNote.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    await homePage.navigateToProgramme(NLRCProgram.titlePortal.en);
+  });
+
+  await test.step('Should validate first row with uploaded PAs', async () => {
+    await table.clickOnPaNumber(1);
+  });
+
+  await test.step('Should validate PA profile opened succesfully, add a note and validate new note log tile', async () => {
+    const userName =
+      process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN ?? 'defaultUserName';
```
aberonni, line 47: Will this env variable ever not be defined? We rely on it in so many places. If this was to stop typescript from complaining, you can also tell typescript "I know this will always be defined" by adding an exclamation mark at the end:

```suggestion
      process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN!;
```

(Use sparingly :D )

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,54 @@ class RegistrationDetails {
     ).toContain(userName);
     expect(await timeStampLocator.textContent()).toContain(date);
   }
+
+  async addNote(actions: string, addNote: string) {
+    await this.page.getByText(actions).click();
+    await this.page.getByText(addNote).click();
+  }
+
+  async writeNote(placeholder: string, note: string, buttonName: string) {
```
aberonni, line 247: This is not only true here, but true in our whole codebase (and a personal gripe I have in many places in the 121-service). When using multiple arguments like this, it becomes hard to understand what is going on in the test itself. I look at where `writeNote` is being used, and I have no idea what is being passed into the function.

I would recommend using [an options object](https://stackoverflow.com/a/12827204) instead of multiple parameters, which besides readability has several other advantages (for example, order of arguments becomes irrelevant).

For example:

```suggestion
  async writeNote({
    placeholder,
    note,
    buttonName
  }: {
    placeholder: string,
    note: string,
    buttonName: string
  }) {
```

I will also add a suggestion below (where this is called) to demonstrate the increase in readability and I would recommend doing this for other new functions for consistency.

e2e/tests/121-Portal/ViewPaProfilePage/SuccessfullyAddNote.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    );
+    await registration.writeNote(
+      englishTranslations['registration-details']['activity-overview'][
+        'add-note-popup'
+      ]['note-textarea-placeholder'],
+      englishTranslations['registration-details']['activity-overview'][
+        'add-note-popup'
+      ]['note-visible'],
+      englishTranslations.common.ok,
+    );
```
aberonni, line 65: Following up on my suggestion above, you can see how difficult to read this is (what are we passing to the function?). While, with the change I suggest:
```suggestion
    await registration.writeNote({
      placeholder: englishTranslations['registration-details']['activity-overview'][
        'add-note-popup'
      ]['note-textarea-placeholder'],
      note: englishTranslations['registration-details']['activity-overview'][
        'add-note-popup'
      ]['note-visible'],
      button: englishTranslations.common.ok,
    });
```

aberonni (approved):
It could potentially be merged as-is, but there are a few quality of life improvements that I think would be worth addressing. Up to you!

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
Piotrk39: Could be actually. It is a fair point it was not mentioned in the "feature file" but I think it is required for it to be disabled before the note being typed.

Piotrk39: Added

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
Piotrk39: Yeah.. User adds note and after accepting note in the first screen then he is prompted with "saved successfully" pop-up that he has to click "ok" on it. so it is just double "Ok"button for adding one note. Would you say double command clicking Ok and waiting would be clearer in this case?

e2e/tests/121-Portal/ViewPaProfilePage/SuccessfullyAddNote.spec.ts
Piotrk39: Awesome thanks for the tip

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
Piotrk39: Okay this is gold in here

e2e/tests/121-Portal/ViewPaProfilePage/SuccessfullyAddNote.spec.ts
Piotrk39: I started realising it when the translation file was introduced to me and I just did not know how to do it in a better way. This now is cleaner easy to understand and anyone can come back to it many months from now and just understand what is going on. This rocks.

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,54 @@ class RegistrationDetails {
...
+      name: buttonName,
+    });
+    await this.page.getByPlaceholder(placeholder).fill(note);
+    await this.page.waitForLoadState('networkidle');
+    await expect(okButton).toBeEnabled();
+
+    for (let i = 0; i < 2; i++) {
+      await okButton.waitFor({ state: 'visible' });
+      await okButton.click();
+    }
```
aberonni, line 267:
```suggestion
    // accept note
    await okButton.waitFor({ state: 'visible' });
    await okButton.click();
    // dismiss "saved successfully" popup
    await okButton.waitFor({ state: 'visible' });
    await okButton.click();
```

Piotrk39: Yes this is how I also thought about it. It is a good solution. It was a pickle to decide here but it is more readable now and not so repetitive also it is almost the same amount of line like the loop.

e2e/pages/RegistrationDetails/RegistrationDetailsPage.ts
```diff
@@ -238,6 +238,66 @@ class RegistrationDetails {
...
+    await okButton.click();
+  }
+
+  async validateNoteTile(
+    noteIndex: number,
+    changeTitle: string,
+    userName: string,
+    date: string,
+    noteContent: string,
+  ) {
```
aberonni, line 278: Maybe also worth doing while you're at it

```suggestion
  async validateNoteTile({
    noteIndex = 0,
    changeTitle,
    userName,
    date,
    noteContent,
  }, {
    noteIndex?: number,
    changeTitle: string,
    userName: string,
    date: string,
    noteContent: string,
  }) {
```

aberonni (dismissed):

aberonni (approved):
LGTM!

PR #5271, opened 2024-05-09
chore: update TypeScript to v5

aberonni:
[AB#27996](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27996)

## Describe your changes

Updating TS to v5

### DIFF
--- a/services/121-service/package.json
+++ b/services/121-service/package.json
@@ -32,10 +32,10 @@
   "private": true,
   "dependencies": {
     "@nestjs/axios": "^3.0.2",
-    "@nestjs/passport": "^10.0.3",
     "@nestjs/bull": "^10.1.1",
     "@nestjs/common": "^10.3.8",
     "@nestjs/core": "^10.3.8",
+    "@nestjs/passport": "^10.0.3",
     "@nestjs/platform-express": "^10.3.8",
     "@nestjs/schedule": "^4.0.2",
     "@nestjs/swagger": "^7.3.1",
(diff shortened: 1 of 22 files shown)


services/121-service/package.json
```diff
@@ -32,10 +32,10 @@
   "private": true,
   "dependencies": {
     "@nestjs/axios": "^3.0.2",
-    "@nestjs/passport": "^10.0.3",
     "@nestjs/bull": "^10.1.1",
     "@nestjs/common": "^10.3.8",
     "@nestjs/core": "^10.3.8",
+    "@nestjs/passport": "^10.0.3",
```
elwinschmitz, line 38: This thing weirdly jumps around a few times...

I wonder if everyone is using the same Node version?
Or that there are differences in `npm` running on macOS or Windows...
Or that Dependabot has an influence on it?

services/121-service/package.json
```diff
@@ -100,6 +99,7 @@
     "supertest": "^7.0.0",
     "ts-jest": "^29.1.2",
     "tsc-watch": "^6.2.0",
+    "typescript": "^5.5.0-beta",
```
elwinschmitz, line 102: I'd be a bit squeamish about using a 'beta' version... but if it really unlocks features we need right now, or speeds things up...

services/121-service/package.json
aberonni: Indeed, I don't get it either. I do think it's dependabot moving it around manually, rather than using npm install. I'm not too fussed about it though tbh :D

services/121-service/package.json
aberonni: Normally, me too. I'm a bit reassured by the fact that:
- TS team says it can be used in production
- I'd assume that any weirdness stemming from a TS beta would be caught either in compilation or in starting the service
- This version is [set to release officially in about a month](https://github.com/microsoft/TypeScript/issues/57475) so we wouldn't be in beta for long

> but if it really unlocks features we need right now, or speeds things up...

Indeed, not necessarily that we "need" but it does avoid adding a lot of extra code that would be made redundant with the 5.5 version

elwinschmitz (approved):

PR #5274, opened 2024-05-09
chore: avoid using JSON types in entities and DTOs

aberonni:
[AB#28002](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28002)

This is foundational work that will contribute to making the "strictNullCheck" work easier.


ashishkumarbasak-accenture (approved):
@aberonni Looks fine to me. Tested most of the functionality in my local and working fine! Approving.

aberonni:
@global-121/510 FYI this is relevant for future entities/DTOs so it's worth checking what's changed. Mostly two things:

- Columns that are defined as type 'json' should not use `JSON` as their type in TypeScript, but instead define the shape of their contents
- Columns that are nullable should contain `| null` as part of their type definition

services/121-service/src/financial-service-providers/dto/update-financial-service-provider.dto.ts
```diff
@@ -1,16 +1,18 @@
...
+import { LocalizedString } from 'src/shared/enum/language.enums';
+import { QuestionOption } from 'src/shared/enum/question.enums';
 import { ExportType } from '../../metrics/dto/export-details.dto';
 import { ProgramPhase } from '../../shared/enum/program-phase.enum';
 
 export class UpdateFspAttributeDto {
   @ApiProperty({ example: { en: 'attribute label' } })
   @IsOptional()
-  public label: JSON;
+  public label: LocalizedString;
```
RubenGeo, line 11: Should this also be LocalizedString | null because of the is IsOptional? Or is LocalizedString already nullable?

Same question for the other `@IsOptional()` dto attributes?

services/121-service/src/financial-service-providers/fsp-question.entity.ts
```diff
@@ -24,25 +27,25 @@ export class FspQuestionEntity extends Base121Entity {
 
   @Column('json')
   @ApiProperty({ example: { en: 'label' } })
-  public label: JSON;
+  public label: LocalizedString;
```
RubenGeo, line 30: Great that you got rid of these JSON types! They were really annoying and I felt we always had to jump thourgh hopes to get them working..

services/121-service/src/financial-service-providers/dto/update-financial-service-provider.dto.ts
aberonni: Excellent question. Indeed they need to be addressed too. Will look into this.

PR #5275, opened 2024-05-09
chore: enable strict mode on 121-service

aberonni:
[AB#27995](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27995)

## Describe your changes

- Turn on `strict: true`
- Fix all compiler errors
- ???
- Profit

- [ ] ~I do not need any deviation from our PR guidelines~ - please do not merge on my behalf

### DIFF
--- a/services/121-service/src/notifications/message-incoming/message-incoming.service.ts
+++ b/services/121-service/src/notifications/message-incoming/message-incoming.service.ts
@@ -96,7 +96,7 @@ export class MessageIncomingService {
     return '';
   }
 
-  public async findOne(sid: string): Promise<TwilioMessageEntity> {
+  public async findOne(sid: string): Promise<TwilioMessageEntity | null> {
     const findOneOptions = {
       sid: sid,
     };
@@ -202,7 +202,7 @@ export class MessageIncomingService {
           select: ['transactionId'],
         },
       );
-      if (messageWithTransaction) {
+      if (messageWithTransaction?.transactionId) {
         await this.intersolveVoucherService.processStatus(
           callbackData,
           messageWithTransaction.transactionId,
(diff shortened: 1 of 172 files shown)


Vladimir-Petkov (approved):

services/121-service/src/registration/registrations.service.ts
```diff
@@ -287,6 +287,9 @@ export class RegistrationsService {
       customDataValueRaw,
       registration.programId,
     );
+    if (customDataValue == undefined) {
+      throw new Error('XXX: what to do here?');
```
ashishkumarbasak-accenture, line 291: I think its fine to throw `Error`. Could you add a proper valid message here?

services/121-service/src/programs/programs.service.ts
```diff
@@ -401,22 +421,23 @@ export class ProgramService {
   }
 
   // This function takes a filled ProgramEntity and returns a filled ProgramReturnDto
-  private fillProgramReturnDto(program: ProgramEntity): ProgramReturnDto {
+  private fillProgramReturnDto(program: FoundProgram): ProgramReturnDto {
+    // XXX: what to do with all these nullable values? For now, forcing them to defined, but this is hacky
```
ashishkumarbasak-accenture, line 425: IMO it makes more sense to fix the interface `FoundProgram` according to our data, But after checked it, I feel its more complicated 😄, may be for now its OK to keep it like this as hacky.

services/121-service/src/payments/fsp-integration/safaricom/safaricom.service.ts
```diff
@@ -91,6 +90,11 @@ export class SafaricomService
       (user) => user.referenceId == jobData.paPaymentData.referenceId,
     );
 
+    if (!resultUser) {
+      // XXX: should throw error?
```
ashishkumarbasak-accenture, line 94: I think, Yes, its better to throw an error here rather than just return.

services/121-service/src/notifications/message-incoming/message-incoming.service.ts
```diff
@@ -225,28 +225,47 @@ export class MessageIncomingService {
...
+      );
+    }
+
+    const registration = await this.registrationRepository.findOneOrFail({
       where: { id: message.registrationId },
     });
 
-    await this.queueMessageService.addMessageToQueue(
+    if (!message.processType) {
+      // XXX: does this make sense?
```
ashishkumarbasak-accenture, line 240: I think, its fine.

services/121-service/src/notifications/message-incoming/message-incoming.service.ts
```diff
@@ -225,28 +225,47 @@ export class MessageIncomingService {
...
-    );
+    });
   }
 
   private async handleTryWhatsappResult(
     callbackData: TwilioStatusCallbackDto,
     tryWhatsapp: TryWhatsappEntity,
   ): Promise<void> {
+    if (!tryWhatsapp.registration.phoneNumber) {
+      // XXX: should we fail silently instead?
```
ashishkumarbasak-accenture, line 264: To me, throws error makes more sense.

services/121-service/src/programs/programs.service.ts
aberonni: I decided to change `ProgramReturnDto` instead, so that the fields that can be undefined are set to `IsOptional` given that this reflects "reality" better.

ashishkumarbasak-accenture (approved):
From my side, it looks good to me. Approving!

ashishkumarbasak-accenture:
@aberonni Perhaps we can wait to merge this after tomorrow's deploy.

PR #5294, opened 2024-05-14
chore: add knip and remove dead code

aberonni:
[AB#28085](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28085)

## Describe your changes

This adds a tool called [knip](https://knip.dev/) to the 121-service, which makes it easier to stay on top of dead code in an automated way.


services/121-service/.knip.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
```
elwinschmitz, line 2:
```suggestion
  "$schema": "node_modules/knip/schema.json",
```

Would this work?
Then it would use the currently-installed-version's schema, right?

(I have used a similar pattern in another angular-app, so I guess it should work here as well..)

That way, we're not relying on a third-party-service AND it doesn't need to update for every version-upgrade...

RubenGeo (approved):
Very cool

RubenGeo (approved):

PR #5295, opened 2024-05-14
Fear.remove all related to phase from frontend

ashishkumarbasak-accenture:
[AB#27963](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27963)

## Describe your changes

Refactoring the frontend code to remove program phase and use program tabs enum instead.


interfaces/Portal/src/app/app-routing.module.ts
```diff
@@ -122,7 +114,13 @@ const routes: Routes = [
         // Fallback for change in url, from old to new syntax:
         path: 'registration-validation',
```
elwinschmitz, line 115:
```suggestion
        path: 'registrationValidation',
```

This would be the new "old" URL. :)
The move from "registration-validation" to "registrationValidation" is a _very_ long time ago now... So that can go, I think.

interfaces/Portal/src/app/components/status-table-filter/status-table-filter.component.ts
```diff
@@ -14,7 +14,7 @@ export class StatusTableFilterComponent {
   public programId: number;
 
   @Input()
-  public thisPhase: ProgramPhase;
+  public thisPhase: ProgramTab;
```
elwinschmitz, line 17: I wonder if it is worth it to "refactor the: 'in what phase are some things allowed'-logic totally away" versus "just rename it to something-NOT-called-phase..

interfaces/Portal/src/app/program/program-tabs-navigation/program-tabs-navigation.component.html
```diff
@@ -1,6 +1,6 @@
 <div>
   <span
-    *ngFor="let phase of programPhases"
+    *ngFor="let phase of ProgramTabs"
```
elwinschmitz, line 3: I'd also rename all "phase*"-things to "tab*"-things in the rest of the template below.

interfaces/Portal/src/app/services/program-phase.service.ts
```diff
@@ -14,8 +14,8 @@ export class Phase {
 @Injectable({
   providedIn: 'root',
 })
-export class ProgramPhaseService {
-  public activePhaseName: ProgramPhase;
+export class ProgramTabService {
```
elwinschmitz, line 17: Is it worth it to also rename all these internal "phase"-things to "tab"-things?

Or would it be similar effort to "refactor this thing away"?

interfaces/Portal/src/app/pages/payment/payment.page.ts
```diff
@@ -14,7 +14,7 @@ export class PaymentPage {
   public Permission = Permission;
 
   public programId = this.route.snapshot.params.id;
-  public thisPhase = ProgramPhase.payment;
+  public thisPhase = ProgramTab.payment;
   public isReady: boolean;
```
elwinschmitz, line 18: I think this whole "ready"-state thing per phase is irrelevant by now...
We might be able to "just delete it all" already...

interfaces/Portal/src/app/pages/peoples-affected/peoples-affected.page.html
```diff
@@ -39,12 +20,15 @@
...
+</div>
 
+<ion-content class="ion-padding">
   <app-program-people-affected
     #table
     *appIfPermissions="[Permission.RegistrationREAD]"
     [programId]="programId"
     [thisPhase]="thisPhase"
     (isCompleted)="onReady($event)"
+    displayImportRegistration="true"
```
elwinschmitz, line 32:
```suggestion
    [displayImportRegistration]="true"
```

I think the type-checking isn't enabled yet on these templates, but this would make it checkable in the future...

interfaces/Portal/src/app/program/program-people-affected/program-people-affected.component.ts
```diff
@@ -84,15 +84,16 @@ export class ProgramPeopleAffectedComponent implements OnDestroy {
   @Input()
   public programId: number;
   @Input()
-  public thisPhase: ProgramPhase;
+  public displayImportRegistration: boolean = false;
+  @Input()
```
elwinschmitz, line 88:
```suggestion
  public displayImportRegistration = false;

  @Input()
```

Hmm.. I thought ESLint would complain about this... :?
Not a real issues, but its not necessary to define the type as a `string`|`number`|`boolean` when you already set the default value (if that has one of these types)...

interfaces/Portal/src/app/services/program-phase.service.ts
elwinschmitz: When we work on the new Payment-flow (and remove the current Payment-page), this whole service will not be necessary anymore.
So I think we can live with it for a little bit longer and not rename-all-the-things for now...

interfaces/Portal/src/app/components/status-table-filter/status-table-filter.component.ts
ashishkumarbasak-accenture: I think we could live with this term `thisPhase` for now, leave it for later :)

elwinschmitz (approved):
There will probably be still some random "phase"-related terms here and there... but we can weed them out whenever we touch any of these files again, probably...

PR #5297, opened 2024-05-15
Chore.visa mock service

RubenGeo:
## Describe your changes


RubenGeo (commented):
@aberonni  

- You put all these import on src/shared is it a problem if I change it to ../shared?\
- The reason I am asking is because if I don't change it get problem importing DTOs from the mock-service
- If it is not a problem to change it to ../shared. Do you think we should look at a eslint rule to enforce this?

PR #5300, opened 2024-05-16
Test.view information icon in table

Piotrk39:
Refactors TableModule aligning it with the suite's pattern and adds e2e test: [View information icon in table](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28035)

### DIFF
--- a/e2e/tests/121-Portal/EditInfoPersonAffected/ViewInformationIconInTable.spec.ts
+++ b/e2e/tests/121-Portal/EditInfoPersonAffected/ViewInformationIconInTable.spec.ts
@@ -1,12 +1,12 @@
 import HomePage from '@121-e2e/pages/Home/HomePage';
 import LoginPage from '@121-e2e/pages/Login/LoginPage';
+import TableModule from '@121-e2e/pages/Table/TableModule';
 import NLRCProgram from '@121-service/seed-data/program/program-nlrc-ocw.json';
 import { SeedScript } from '@121-service/src/scripts/seed-script.enum';
 import { seedPaidRegistrations } from '@121-service/test/helpers/registration.helper';
 import { resetDB } from '@121-service/test/helpers/utility.helper';
 import { registrationsOCW } from '@121-service/test/registrations/pagination/pagination-data';
 import { test } from '@playwright/test';
-// import englishTranslations from '../../../../interfaces/Portal/src/assets/i18n/en.json';
 
 test.beforeEach(async ({ page }) => {
   await resetDB(SeedScript.nlrcMultiple);
@@ -25,7 +25,7 @@ test.beforeEach(async ({ page }) => {
 });
 
 test('[28035] View information icon in table', async ({ page }) => {
-  // const table = new TableModule(page);
+  const table = new TableModule(page);
   const homePage = new HomePage(page);
 
   await test.step('Navigate to PA table', async () => {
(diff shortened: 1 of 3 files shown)


interfaces/Portal/src/app/program/program-people-affected/program-people-affected.component.html
```diff
@@ -173,6 +173,7 @@
         style="display: flex"
       >
         <button
+          data-testid="information-popup-button"
           *ngIf="canViewPersonalData"
           type="button"
```
elwinschmitz, line 178: I'd prefer to add these properties in "order of importance", so:

```suggestion
          *ngIf="canViewPersonalData"
          type="button"
          data-testid="information-popup-button"
```

Then you first see: "should this thing even be here or not" (based on the `if`)
Then; the most-important/default attributes; (could be more then only this `type`, in this case)
Then; Other specific attributes (like this `testid` one).
Then; Other, 'less critical' properties like styles/class/et.

In this example, the other properties below could also be reused according to all kinds of discussion, sot that is not valuable to change/discuss here/now;
But the thing you add here, should not be "jumping the line" so that it 'hides' the 'important `if`-condition.. in a way.

e2e/tests/121-Portal/EditInfoPersonAffected/ViewInformationIconInTable.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+test('[28035] View information icon in table', async ({ page }) => {
+  const table = new TableModule(page);
+  const homePage = new HomePage(page);
+
+  await test.step('Navigate to PA table', async () => {
+    await homePage.navigateToProgramme(NLRCProgram.titlePortal.en);
+  });
+
+  await test.step('Validate information icons are available in the table', async () => {
+    await table.validateInformationButtonsPresent({ expectedCount: 4 });
```
elwinschmitz, line 36: Would this number `4` be the `registrationsOCW.length`?

It could make sense to 'fix' the number to this hard-coded value, but sometimes I also like that the code is 'resilient to change'; so that you don't _have to_ change this value only because 1 other test-registration is added to the set (because of totally unrelated reasons)...

elwinschmitz (approved):
Besides the comments added previously; ✅

e2e/tests/121-Portal/EditInfoPersonAffected/ViewInformationIconInTable.spec.ts
Piotrk39: Kind of made sense to me to put it like that just in case seed function will change but if you believe it is better to keep it fixed I can change it.

PeterSmallenbroek (approved):

e2e/tests/121-Portal/EditInfoPersonAffected/ViewInformationIconInTable.spec.ts
elwinschmitz: My point was sort-of the other way around :)

I prefer to use "the _least_ amount of really specific numbers/strings in code"(so a number `4`, or some `string`) and **always** use "the _most_ descriptive/specific way to define something" (so using a variable/property/piece of working code; like this `.length`-property, or something like: `60 * 60 * 24 * 7` to describe 'the number of seconds in a week', instead of `604800`)

PR #5313, opened 2024-05-21
Refactor.home page pom

Piotrk39:
Slightly updates homePage POM with data-testIds and removes unnecessary okButton locator.

### DIFF
--- a/e2e/pages/Home/HomePage.ts
+++ b/e2e/pages/Home/HomePage.ts
@@ -6,18 +6,14 @@ class HomePage {
   readonly activeProgramsBanner: Locator;
   readonly programCard: Locator;
   readonly openPAsForRegistrationButton: Locator;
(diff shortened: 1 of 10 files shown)


aberonni (changes requested):
A lot of the test ids do not mention the component name in the id itself, which is a guideline I think we established in a previous PR, and which would make it easier to make sure that we don't have duplicate test ids. Could you make sure you check the new ids for this? The rest seems good :)

Piotrk39:
I made some changes in the names of testids. I hope they are a bit better now. Thanks for pointing this out @aberonni 

> A lot of the test ids do not mention the component name in the id itself, which is a guideline I think we established in a previous PR, and which would make it easier to make sure that we don't have duplicate test ids. Could you make sure you check the new ids for this? The rest seems good :)

PR #5344, opened 2024-05-28
chore: add 121-service alias to portal

aberonni:
[AB#28381](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28381)

### DIFF
--- a/interfaces/Portal/src/app/auth/permission.enum.ts
+++ b/interfaces/Portal/src/app/auth/permission.enum.ts
@@ -1,4 +1,4 @@
-import { PermissionEnum } from '../../../../../services/121-service/src/user/enum/permission.enum';
+import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
 /**
  * Note: Permissions-definition explicitly used from back-end, to prevent repetition.
  */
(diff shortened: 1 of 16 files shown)


interfaces/Portal/src/app/auth/permission.enum.ts
```diff
@@ -1,4 +1,4 @@
-import { PermissionEnum } from '../../../../../services/121-service/src/user/enum/permission.enum';
+import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
```
elwinschmitz, line 1: We could get rid of the "Portal-local"-file now, I think.

It was only here to prevent having all these 'deep' imports everywhere in the Portal-code.

With these aliases, it serves no purpose anymore.

But I still don't like the naming of (a lot of) these 121-service `enum`s.
- Some are (inconsistently) post-fixed with `*Enum`
- Some are in plural form
- etc..

elwinschmitz (approved):
I think the 'bigger' renaming/refactoring can wait some more; But we should look into it somehow

aberonni:
@elwinschmitz yeah, I figure a lot of this will be addressed when we get around to rebuilding the UI. The changes I made to the files were only auto-fixes from esilnt. I didn't bother making any further changes. Will merge as-is for the time being.

interfaces/Portal/src/app/auth/permission.enum.ts
aberonni: As mentioned in the other comment, I only made auto-fix lint changes to the files. I'd rather avoid making other changes on this PR to avoid scope creep and rabbit holes.

PR #5352, opened 2024-05-29
Improve speed of E2E-test runs

elwinschmitz:
- Fail as early as possible with the 'lightest' tests (linting)
- Add some caching-steps to reuse effort of previous test-runs

### DIFF
--- a/.github/workflows/test_e2e.yml
+++ b/.github/workflows/test_e2e.yml
@@ -18,7 +18,7 @@ on:
       - '!**.md'
 
 env:
-  e2eTestsPath: 'e2e/'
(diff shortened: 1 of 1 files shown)


aberonni (approved):
FTR the lint task was initially moved lower in the chain because the healthcecks for docker were done asynchronously - so it made sense to run the docker services ASAP and let them get ready "in the background". Since that has changed since (and the docker step does not pass until the services are healthy), then moving the lint task further up the chain makes a lot of sense ✌️

Piotrk39 (approved):
Looks good and comparably it runs faster!

aberonni:
LGTM? :)

PR #5351, opened 2024-05-28
Build with swc AB#28260

RubenGeo:
[Item](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20129?workitem=28260) 

## Describe your changes

**Build faster with swc**

Main changes:
- Moved all code and seed jsons to the src folder. I did not find a way to make swc work without it. These are the majority of the changes files. Stuff I moved
  - ormconfig
  - seed-data
  - migrations
- Config files for swc
  - Changes to the package.json npm script to use nest-cli instead of node
  - nest-cli.json: Config file for nest clients, it mainly states to swc
  - .swcrc: Config file for swc. These I had to [duplicate](https://stackoverflow.com/a/76927890) the baseUrl and path of tsconfig. 
  - Dockerfile install procps to make swc work
- Code changes. 
  - I imlemented the [relation solution](https://docs.nestjs.com/recipes/swc#common-pitfalls) for our entities as described in the nestjs docs
  - For enums in our dtos I implemented the [WrapperType](https://docs.nestjs.com/recipes/swc#common-pitfalls). I do think this workaround suggested by the Nestjs docs is a bit annoying. But since the compile time is so much faster of swc I think it is worth it to use it. 
- Changes have also been applied to the mock-service 
- Changes to deploy scripts 
  - I added build to package.json
  - Added build to the workflow files
  - I think somebody should still take a critical look at these deploy scripts @aberonni or @elwinschmitz ?
  - It does work succesfully for test. See run 

TODO: 
- see if migrations still run on productions
  - I expect they do but I did not test this
- Check if there are no unintended changes are in this PR 

## Checklist before requesting a review

- [ ] I have performed a self-review of my code
- [ ] I have added tests wherever relevant
- [ ] I have made sure that all automated checks pass before requesting a review
- [ ] I do not need any deviation from our PR guidelines

PR #5355, opened 2024-05-30
chore: implement SWC

aberonni:
[AB#28405](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28405)

## Describe your changes

(Built on top of #5351)

**Build faster with swc**

Main changes:
- ~Moved all code and seed jsons to the src folder. I did not find a way to make swc work without it. These are the majority of the changes files. Stuff I moved~ This is now separated into #5357 
  - ormconfig
  - seed-data
  - migrations
- Config files for swc
  - Changes to the package.json npm script to use nest-cli instead of node
  - nest-cli.json: Config file for nest clients, it mainly states to swc
  - .swcrc: Config file for swc. These I had to [duplicate](https://stackoverflow.com/a/76927890) the baseUrl and path of tsconfig. 
  - Dockerfile install procps to make swc work
- Code changes. 
  - I imlemented the [relation solution](https://docs.nestjs.com/recipes/swc#common-pitfalls) for our entities as described in the nestjs docs
  - For enums in our dtos I implemented the [WrapperType](https://docs.nestjs.com/recipes/swc#common-pitfalls). I do think this workaround suggested by the Nestjs docs is a bit annoying. But since the compile time is so much faster of swc I think it is worth it to use it. 
- Changes have also been applied to the mock-service 
- Changes to deploy scripts 
  - I added build to package.json
  - Added build to the workflow files
  - It does work succesfully for test. See run 

## TODO (from Ruben): 
- [x] see if migrations still run on productions - tested in #5359
- [x] Check if there are no unintended changes are in this PR 
- [x] Take a critical look at these deploy scripts

- [ ] ~I do not need any deviation from our PR guidelines~ Please don't merge on my behalf


.github/actions/build-service/action.yml
```diff
@@ -30,6 +30,11 @@ runs:
       working-directory: ${{ inputs.servicePath }}
       run: 'npm ci --no-fund --no-audit'
 
+    - name: Build
+      shell: bash
+      working-directory: ${{ inputs.servicePath }}
+      run: 'npm run build'
```
elwinschmitz, line 36: We could add some caching-steps around this step (so that multiple test-runs can reuse the same data/code);
But lets not introduce that just-yet.

Lets first create something-to-improve-on ;)

aberonni: Even without cache it looks pretty fast :)

image

services/121-service/src/seed-data/program/program-nlrc-pv.json
```diff
@@ -4,7 +4,7 @@
   "location": "Nederland",
   "ngo": "NLRC",
   "titlePortal": {
-    "en": "NLRC Direct Digital Aid Program (PV)"
+    "en": "NLRC Direct Digital Aid Program"
```
aberonni, line 7: Not sure about this change - feels like it was a mistake or a test

RubenGeo: Yes this was a mistake.

Vladimir-Petkov (approved):

Vladimir-Petkov (approved):

PR #5359, opened 2024-05-30
SWC: test that migrations still work on a production server

aberonni:
Do not merge - test only. This is to test #5355 on a production environment


aberonni:
Successfully deployed to the Rwanda instance without any issues. Reverted the DB changes re-deployed the previous version.

PR #5364, opened 2024-06-03
Change instance to organization AB#28493

RubenGeo:
## Describe your changes


aberonni:
@RubenGeo can you rebase & solve the conflicts?

services/121-service/src/shared/enum/question.enums.ts
```diff
@@ -1,4 +1,4 @@
-import { LocalizedString } from '@121-service/src/shared/enum/language.enums';
```
aberonni, line 1: How was this working? 🤔  Could it be that this file isn't actually used anywhere?

services/121-service/src/seed-data/organization/organization-anonymous.json
```diff
@@ -0,0 +1,6 @@
+{
```
aberonni, line 1: Looks like the deletion of the instance seed data slipped between the cracks

aberonni (changes requested):

services/121-service/src/seed-data/organization/organization-anonymous.json
RubenGeo: I don't get it? Was this file deleted in main? @aberonni

services/121-service/src/shared/enum/question.enums.ts
RubenGeo: Not completely sure what you mean here. I think in the swc branch I put: LanguageEnum and LocalizedString in seperate files. Is that what you mean?

services/121-service/src/seed-data/organization/organization-anonymous.json
aberonni: No, I mean that these seed files in `src/seed-data/organization` seem to be replacing the ones in `src/seed-data/instance` so I would assume that the latter folder would be deleted (but it doesn't seem that way).

services/121-service/src/shared/enum/question.enums.ts
aberonni: You fixed an import in this file, which was broken. I don't understand how the broken import got onto main in the first place, and why there are no complaints about the broken import in CI. So I was wondering whether `question.enums.ts` is actually not imported anywhere and that is why there are no complaints. But just a theory/doubt

RubenGeo: Not sure how this happened, but `question.enum.ts` is imported in a bunch of files

services/121-service/src/seed-data/organization/organization-anonymous.json
RubenGeo: Ah good one!

services/121-service/src/shared/enum/question.enums.ts
aberonni: 🤷‍♂️ 😄

aberonni (approved):

PR #5376, opened 2024-06-04
AB#28412 Add files and test for PV retry

Piotrk39:
## Describe your changes

Adds e2e test for retry of PV payments


e2e/test-registration-data/test-registrations-PV.csv
```diff
@@ -0,0 +1,21 @@
+referenceId,preferredLanguage,paymentAmountMultiplier,maxPayments,fullName,phoneNumber,fspName,whatsappPhoneNumber,addressStreet,addressHouseNumber,addressHouseNumberAddition,addressPostalCode,addressCity
```
RubenGeo, line 1: Are the 4 files in e`2e/test-registration-data` a copy of what is in `features/test-registration-data/`?

It is generally better to not duplicate files in 2 places. Otherwise if something changes later on you also have to adjust 2 files. Is it possible to import from that the current folder in playwright? Or does it make sense to create a root level `/test-data` folder with `/registration-data` in it?

e2e/tests/121-Portal/MakeNewPayment/PV/RetryPaymentPV.spec.ts
```diff
@@ -14,19 +14,19 @@ import {
   getAccessToken,
   resetDB,
 } from '@121-service/test/helpers/utility.helper';
-import { programIdOCW } from '@121-service/test/registrations/pagination/pagination-data';
+import { programIdPV } from '@121-service/test/registrations/pagination/pagination-data';
```
RubenGeo, line 17: You both import programIdPV here but you also set it at line 23. Only one of the 2 is needed :)

RubenGeo (changes requested):

e2e/test-registration-data/test-registrations-PV.csv
RubenGeo: Maybe you can already move all files from `features/test-registration-data/` to `e2e/test-registration-data` than. 

I don't think it makes sense to store the csv in the api test folder (`services/121-service/test`). Since we do not use csv's in our API tests right?

Piotrk39: We do use files for bulk-update in API tests. I could maybe move the rest of the files from the features folder for now and clean up unused files from API tests?

e2e/tests/121-Portal/MakeNewPayment/PV/RetryPaymentPV.spec.ts
Piotrk39: Done

e2e/test-registration-data/test-registrations-PV.csv
Piotrk39: So what I did was to redistribute files from feature folder and place them accordingly in API and E2E test folders. The ones in E2E are going to be used there sooner or later in accordance with test scenarios so I think it is a good idea to keep them there. Folder from features was removed. @RubenGeo you can take a look again once the tests pass..

RubenGeo (approved):

PR #5386, opened 2024-06-05
Release v1.130.0

aberonni:
AB#28525


CHANGELOG.md
```diff
@@ -5,10 +5,22 @@ All notable changes to this project will be documented in this file.
...
 
 ---
 
+## [1.130.0](https://github.com/global-121/121-platform/compare/v1.120.0...v1.130.0)- 2024-06-05
+
+### Added
+
+- [AB#28422](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28422) Show completed status on people affected page and default sort as registrationCreated for PAs
+
+### Technical
```
aberonni, line 18: I added this for our own internal "FYI" - but I wouldn't add this to an external stakeholder message

elwinschmitz (approved):
All these mentioned topic are already available in the “compare”-view, right?

the only point of the “unreleased”-section is to keep track of these things *_while working between releases_*…

not to “make it look like we kept track” after-the-fact.

aberonni:
@elwinschmitz I think the answer to your question is yes, but I'm not sure I follow 🙈

elwinschmitz:
> @elwinschmitz I think the answer to your question is yes, but I'm not sure I follow 🙈

That’s where the conversation during refinement is for. ;)
(I’ll try an prepare something I guess)

PR #5387, opened 2024-06-10
Pause card reimplementation

PeterSmallenbroek:
[AB#28246](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28246)

## Describe your changes


interfaces/Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -545,16 +545,15 @@ export class ProgramsServiceApiService {
...
     tokenCode: string,
     block: boolean,
   ): Promise<any> {
-    return await this.apiService.post(
+    return await this.apiService.patch(
       environment.url_121_service_api,
-      `/programs/${programId}/financial-service-providers/intersolve-visa/wallets/${tokenCode}/${
-        block ? 'block' : 'unblock'
-      }`,
+      `/programs/${programId}/registrations/${referenceId}/financial-service-providers/intersolve-visa/cards/${tokenCode}?pause=${block}`,
```
diderikvw, line 556: Small thing for making it cleaner: IMO we should not use the term "block" here anymore. Maybe just {pause}? I am not familiar with the context in which this is called, so not sure on the best name.

services/121-service/src/payments/fsp-integration/intersolve-visa/dto/intersolve-block.dto.ts
diderikvw, line 1: This file should be in dto/internal/intersolve-api/

services/121-service/src/payments/fsp-integration/intersolve-visa/dto/intersolve-block.dto.ts
```diff
@@ -12,7 +12,7 @@ export enum UnblockReasonEnum {
   UNBLOCK_GENERAL = 'UNBLOCK_GENERAL',
 }
 
-export class IntersolveBlockWalletResponseDto {
+export class IntersolveApiBlockWalletResponseDto {
```
diderikvw, line 15: AFAIR we agreed that since this is an internal DTO, which sits in the internal/api-intersolve folder, the prefix "Intersolve" is not needed anymore. And the prefix "IntersolveAPI" I do not recall having aligned on.

services/121-service/src/payments/fsp-integration/intersolve-visa/dto/intersolve-block.dto.ts
```diff
@@ -12,7 +12,7 @@ export enum UnblockReasonEnum {
...
   public data: IntersolveBlockWalletResponseBodyDto;
```
diderikvw, line 16: This partial DTO is only used in this DTO. We agreed to refactor that to an inline part of the DTO and not in a separate class.

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.api.service.ts
```diff
@@ -616,33 +617,41 @@ export class IntersolveVisaApiService {
...
-  public async toggleBlockWallet({
-    tokenCode,
-    payload,
-    block,
-  }: {
-    tokenCode: string | null;
-    payload: IntersolveBlockWalletDto;
-    block: boolean;
-  }): Promise<IntersolveBlockWalletResponseDto> {
+  public async setTokenBlocked(
```
diderikvw, line 620: Consider deconstructing into an object.

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.api.service.ts
```diff
@@ -23,8 +23,9 @@ import { IssueTokenResultDto } from '@121-service/src/payments/fsp-integration/i
 import { IssueTokenDto } from '@121-service/src/payments/fsp-integration/intersolve-visa/dto/internal/issue-token.dto';
 import { TransferDto } from '@121-service/src/payments/fsp-integration/intersolve-visa/dto/internal/transfer.dto';
 import {
-  IntersolveBlockWalletDto,
-  IntersolveBlockWalletResponseDto,
+  BlockReasonEnum,
+  IntersolveApiBlockWalletResponseDto,
+  UnblockReasonEnum,
 } from '@121-service/src/payments/fsp-integration/intersolve-visa/dto/intersolve-block.dto';
```
diderikvw, line 29: To do: refactor this DTO file (file location, file name, split into parts? Combining DTOs?)

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.api.service.ts
```diff
@@ -616,33 +617,41 @@ export class IntersolveVisaApiService {
...
-      block ? 'block' : 'unblock'
+      blocked ? 'block' : 'unblock'
     }`;
     const headers = [
       { name: 'Authorization', value: `Bearer ${authToken}` },
       { name: 'Tenant-ID', value: process.env.INTERSOLVE_VISA_TENANT_ID },
     ];
+    const payload = {
+      reasonCode: blocked
+        ? BlockReasonEnum.BLOCK_GENERAL
```
diderikvw, line 637: I think we can put a refactor comment here to take out this complex DTO - enum structure. It is used only here. Except for short-term also in the migrate code. So after that is done, it can be refactored :)

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.controller.ts
```diff
@@ -157,7 +103,7 @@ export class IntersolveVisaController {
     @Param() params,
     @Param('programId', ParseIntPipe)
     programId: number,
-  ): Promise<IntersolveBlockWalletResponseDto> {
+  ): Promise<IntersolveApiBlockWalletResponseDto> {
```
diderikvw, line 106: To refactor: we should not use an internally used Intersolve DTO also in our 121 Service API. Separation of concerns. Design the DTO we use as response body in our API according to what the consumer needs/expects.

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.service.ts
```diff
@@ -796,4 +704,24 @@ export class IntersolveVisaService
       return 0;
     }
   }
+
+  public async pauseCardOrThrow(tokenCode: string, blockToken: boolean) {
```
diderikvw, line 708: Is it our practice that any method where we deliberately throw an error if something goes wrong suffix it with "OrThrow"? If so, then I have some refactoring to do :)

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.service.ts
```diff
@@ -796,4 +704,24 @@ export class IntersolveVisaService
...
+    const wallet =
+      await this.intersolveVisaChildWalletScopedRepository.findOneOrFail({
+        where: { tokenCode: Equal(tokenCode) },
+      });
+    if (wallet.isTokenBlocked === blockToken) {
+      throw new Error('Token is already in this state');
```
diderikvw, line 714: Consider throwing the state that is being set, so "Cannot pause token, it is already paused." and v.v.

services/121-service/src/registration/registrations.service.ts
```diff
@@ -895,4 +899,32 @@ export class RegistrationsService {
       - See to be removed method: IntersolveVisaService.sendMessageReissueCard()
     */
   }
+
+  async pauseCardAndSendMessage(
```
diderikvw, line 903: Explicitly add access operator (public) like we "always do" (?)
Explicitly state the typed return variable, also if it is void, like we "always do"(?)

services/121-service/src/registration/registrations.service.ts
```diff
@@ -895,4 +899,32 @@ export class RegistrationsService {
...
+    _programId: number,
```
diderikvw, line 904: The _ should not be here anymore.

services/121-service/src/registration/registrations.service.ts
```diff
@@ -895,4 +899,32 @@ export class RegistrationsService {
...
+        HttpStatus.NOT_FOUND,
+      );
+    }
+    await this.intersolveVisaService.pauseCardOrThrow(tokenCode, blockToken);
+    await this.queueMessageService.addMessageToQueue({
+      registration: registration,
+      messageTemplateKey: blockToken
+        ? ProgramNotificationEnum.blockVisaCard
+        : ProgramNotificationEnum.unblockVisaCard,
+      messageContentType: MessageContentType.custom,
```
diderikvw, line 925: Refactor use of term "block" into "pause" in template keys.

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -720,8 +721,45 @@ export class RegistrationsController {
...
+    @Param('tokenCode') tokenCode: string,
+    @Query('pause', ParseBoolPipe) pause: boolean,
+  ) {
+    if (pause === undefined) {
+      throw new HttpException(
+        'No pause value (true/false) provided in query parameter',
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+    return await this.registrationsService.pauseCardAndSendMessage(
```
diderikvw, line 758: As far as I can see, this will not return anything, since the function that is called does not return anything.

AFAIK good practice in REST API design is that a PATCH endpoint returns the new representation of the entity it is called on. So the same data structure as with the GET request.

diderikvw (changes requested):
Looks good Peter!
Some small comments that IMO would make the code a bit cleaner. Probably some debatable :)

Also I see some things under the "Unchanged files with check annotations" section here in GH. Probably these are easily fixed.

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.api.service.ts
PeterSmallenbroek: I'll add typechecking for reasonCode and remove/simplify the enum

services/121-service/src/registration/registrations.controller.ts
PeterSmallenbroek: I'll update the ApiResponse to be a 200 and return the updated entity

PR #5398, opened 2024-06-11
AB#28579 load programs fast

RubenGeo:
AB#28579 

## Describe your changes

Load programs faster by doing less inefficient typeorm joins and only getting ids on the assigned programs endpoint

### DIFF
--- a/interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
+++ b/interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
@@ -27,10 +27,15 @@ export class ProgramsListComponent implements OnInit {
 
   private async refresh() {
     this.loading = true;
-    const programs = await this.programsService.getAllPrograms();
+    const programIds = await this.programsService.getAllProgramIds();
     this.programStats = await this.programsService.getAllProgramsStats(
-      programs.map((p) => p.id),
+      programIds.map((id) => id),
     );
+    const programs = [];
+    for (const programId of programIds) {
+      const program = await this.programsService.getProgramById(programId);
+      programs.push(program);
+    }
     this.items = this.translateProperties(programs).sort((a, b) =>
       a.created <= b.created ? -1 : 1,
     );
--- a/services/121-service/src/programs/programs.service.ts
+++ b/services/121-service/src/programs/programs.service.ts
@@ -29,7 +29,6 @@ import { DefaultUserRole } from '@121-service/src/user/user-role.enum';
 import { UserService } from '@121-service/src/user/user.service';
 import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { omit } from 'lodash';
 import { DataSource, In, QueryFailedError, Repository } from 'typeorm';
 
 interface FoundProgram
@@ -77,7 +76,6 @@ export class ProgramService {
     }
 
     const relations = [
-      'programQuestions',
       'financialServiceProviders',
       'financialServiceProviders.questions',
       'programFspConfiguration',
@@ -92,11 +90,14 @@ export class ProgramService {
       throw new HttpException({ errors }, HttpStatus.NOT_FOUND);
     }
 
-    // Program attributes are queried separately because the performance is bad when using relations
+    // Program attributes and questions are queried separately because the performance is bad when using relations
     program.programCustomAttributes =
       await this.programCustomAttributeRepository.find({
         where: { program: { id: programId } },
       });
+    program.programQuestions = await this.programQuestionRepository.find({
+      where: { program: { id: programId } },
+    });
 
     program.editableAttributes =
       await this.programAttributesService.getPaEditableAttributes(program.id);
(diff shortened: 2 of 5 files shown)


interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
```diff
@@ -27,10 +27,15 @@ export class ProgramsListComponent implements OnInit {
 
   private async refresh() {
     this.loading = true;
-    const programs = await this.programsService.getAllPrograms();
+    const programIds = await this.programsService.getAllProgramIds();
     this.programStats = await this.programsService.getAllProgramsStats(
-      programs.map((p) => p.id),
+      programIds.map((id) => id),
```
aberonni, line 32: Can't this simply be like this?

```suggestion
      programIds
```

interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
```diff
@@ -27,10 +27,15 @@ export class ProgramsListComponent implements OnInit {
...
     );
+    const programs = [];
+    for (const programId of programIds) {
+      const program = await this.programsService.getProgramById(programId);
+      programs.push(program);
+    }
```
aberonni, line 38: TS likes it when you use `Array.map`, because then the resulting `programs` will have the correct type.
```suggestion
    const programs = Promise.all(programIds.map(async (programId) =>
      await this.programsService.getProgramById(programId);
    ));
```

(Check the syntax because I'm not 100% sure about it)

services/121-service/src/programs/programs.service.ts
```diff
@@ -154,37 +155,12 @@ export class ProgramService {
     const programIds = user.programAssignments.map((p) => p.program.id);
     const programs = await this.programRepository.find({
```
aberonni, line 156: This is now a bit confusing. I would maybe rename the variables.

```suggestion
    const userProgramIds = user.programAssignments.map((p) => p.program.id);
    const programIds = await this.programRepository.find({
```

aberonni (dismissed):
I like the semplification! I added some comments but they are not must-haves.

interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
```diff
@@ -27,10 +27,15 @@ export class ProgramsListComponent implements OnInit {
 
   private async refresh() {
     this.loading = true;
-    const programs = await this.programsService.getAllPrograms();
+    const programIds = await this.programsService.getAllProgramIds();
```
elwinschmitz, line 30: @RubenGeo 

Weird idea: The Portal can already 'know' these Program-IDs by looking at the "currently logged-in user"'s permissions...
(So a method needs to be added to the AuthService/UserState-service; To return only the keys/indexes from the `permissions`-object)

That would prevent 1 full API-call... (and the corresponding back-end-code).

So this component would then "show all the details of the programs the user has access to"...

RubenGeo: Nice idea. I am just wondering if you would need to login in and out to see newly assigned programs. Or is that needed anyway to get access to them?

interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
RubenGeo: Hahaha yes 😄

interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
RubenGeo: Nice suggestion 👍

interfaces/Portal/src/app/pages/home/components/programs-list/programs-list.component.ts
elwinschmitz: We could use some "refresh the current user"-logic somewhere indeed.
But there is already the "current-user"-endpoint... so that is probably useful for it.

RubenGeo: Maybe on refresh of the programs page. But maybe that is scope creep for this PR @elwinschmitz

elwinschmitz: @RubenGeo Its only these changes:
https://github.com/global-121/121-platform/compare/feat.performance-program-page...feat.add-current-user-check

interfaces/Portal/src/app/app.component.ts
```diff
@@ -32,14 +33,23 @@ export class AppComponent implements OnInit, OnDestroy {
     }
   }
 
-  public ngOnInit(): void {
+  public async ngOnInit() {
```
aberonni, line 36:
```suggestion
  public ngOnInit() {
```

I don't see an `await` so I don't think this is needed

services/121-service/src/user/user.controller.ts
```diff
@@ -280,12 +280,15 @@ export class UserController {
     description: 'No user detectable from cookie or no cookie present',
   })
   public async findMe(@Req() req): Promise<UserRO> {
-    const username = req.user.username;
-    if (!username) {
+    if (!req.user || !req.user.username) {
```
aberonni, line 283: Perhaps? (Not sure if it works)
```suggestion
    if (!req.user?.username) {
```

aberonni (approved):
Some very minor things - up to you if you wanna address them

services/121-service/src/user/user.controller.ts
RubenGeo: @elwinschmitz We are merging anyway

PR #5411, opened 2024-06-12
Clean dist on start dev

RubenGeo:
[AB#28624](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28624)

## Describe your changes
 If you remove a typescript files the compiled js file remains in the dist folder. This can be annoying when creating and deleting migrations files. Now the way to fix this is: 1) delete dist folder 2) restart 121-service. With this commit you do not have to do 1 as it is done on startup

### DIFF
--- a/services/121-service/nest-cli.json
+++ b/services/121-service/nest-cli.json
@@ -13,6 +13,7 @@
       }
     ],
     "watchAssets": true,
+    "deleteOutDir": true,
     "builder": {
       "type": "swc",
       "typeCheck": true,


elwinschmitz:
There is also the "deleteOutDir"-option in the `nest-cli.json`-file:
https://docs.nestjs.com/cli/monorepo#global-compiler-options

Does that do the same trick? Or does that also slow it down a bit?

With only a configuration-option, we wouldn't need an extra (dev-)dependency...

RubenGeo:
> There is also the "`deleteOutDir`"-option in the `nest-cli.json`-file: https://docs.nestjs.com/cli/monorepo#global-compiler-options
> 
> Does that do the same trick? Or does that also slow it down a bit?
> 
> With only a configuration-option, we wouldn't need an extra (dev-)dependency...

Nice it works. Pushed a that change

PeterSmallenbroek (approved):

elwinschmitz (approved):

PR #5416, opened 2024-06-13
AB#28629 add e2e test for OCW & PV

Piotrk39:
## Describe your changes

Tests xlsx download of debit card usage, parses data from xlsx file and validates expected column names


RubenGeo (approved):
Hey looks great! I think we can already merge this. But I do think it is good to also check some of the content of the export. Like is the balance actually 25 and other checks like that

Piotrk39:
@RubenGeo I think it is indeed good idea. I was not sure initially what data should be checked so I left it on the headers. But I'll add this validation as well.

PR #5418, opened 2024-06-14
AB#28652 Block empty dropdown value on import

RubenGeo:
## Describe your changes

### DIFF
--- a/services/121-service/src/registration/validators/registrations-input-validator.ts
+++ b/services/121-service/src/registration/validators/registrations-input-validator.ts
@@ -191,20 +191,26 @@ export class RegistrationsInputValidator {
             const optionNames = att.options
               ? att.options?.map((option) => option.option)
               : [];
-            if (!optionNames.includes(row[att.name])) {
-              const optionNamesErrorString = optionNames.length
+
+            if (optionNames.includes(row[att.name])) {
+              // Validation passed
+              importRecord[att.name] = row[att.name];
+              return;
+            }
+
+            // validation error
+            const optionNamesErrorString =
+              optionNames.length > 0
                 ? optionNames.join(', ')
                 : 'No options available';
-              const errorObj = {
-                lineNumber: i + 1,
-                column: att.name,
-                value: row[att.name],
-                error: `Value '${row[att.name]}' is not in the allowed options: '${optionNamesErrorString}' for attribute '${att.name}'`,
-              };
-              errors.push(errorObj);
-            } else {
-              importRecord[att.name] = row[att.name];
-            }
+            const errorObj = {
+              lineNumber: i + 1,
+              column: att.name,
+              value: row[att.name],
+              error: `Value '${row[att.name]}' is not in the allowed options: '${optionNamesErrorString}' for attribute '${att.name}'`,
+            };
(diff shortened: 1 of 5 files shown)


services/121-service/src/registration/validators/registrations-input-validator.ts
```diff
@@ -187,6 +187,27 @@ export class RegistrationsInputValidator {
...
+                value: row[att.name],
+                error: `Value '${row[att.name]}' is not in the allowed options: '${optionNamesErrorString}' for attribute '${att.name}'`,
+              };
+              errors.push(errorObj);
+            } else {
+              importRecord[att.name] = row[att.name];
+            }
+            return;
+          }
+
```
aberonni, line 210: Whenever possible, I prefer to [return early](https://medium.com/swlh/return-early-pattern-3d18a41bba8). I think it's a nice habit to keep in mind, to avoid nested or long if/elses which make code hard to read. Not necessarily a big issue here, especially if this aligns with what is already there, but something nice to keep in mind perhaps in general.

```suggestion
          if (att.type === AnswerTypes.dropdown) {
            const optionNames = att.options?.map((option) => option.option);
            if (optionNames.includes(row[att.name])) {
              // validation passed
              importRecord[att.name] = row[att.name];
              return;
            }
            
            // validation error
            const optionNamesErrorString = optionNames.length > 0
                ? optionNames.join(', ')
                : 'No options available';
            const errorObj = {
              lineNumber: i + 1,
              column: att.name,
              value: row[att.name],
              error: `Value '${row[att.name]}' is not in the allowed options: '${optionNamesErrorString}' for attribute '${att.name}'`,
            };
            errors.push(errorObj);
            return;
          }

```

aberonni (approved):

services/121-service/src/registration/validators/registrations-input-validator.ts
RubenGeo: Thanks for the feedback, I will implement it per your suggestion! Except I think this is not possible because optionName is possibly `undefined` like this.

`           const optionNames = att.options?.map((option) => option.option);
            if (optionNames.includes(row[att.name])) {`

services/121-service/test/commercial-bank-ethiopia/export-validation-report.test.ts
```diff
@@ -50,7 +50,12 @@ describe('Export CBE validation report', () => {
 
   it('should succesfully generate a report of CBE validation data', async () => {
     // // Arrange
-    await importRegistrations(programId, [registrationCbe], accessToken);
+    const result = await importRegistrations(
+      programId,
+      [registrationCbe],
+      accessToken,
+    );
+    console.log('🚀 ~ it ~ result:', result);
```
aberonni, line 58:
```suggestion
```

aberonni (changes requested):
I think a console.log slipped through the cracks :D

services/121-service/src/registration/validators/registrations-input-validator.ts
aberonni: Ah yes, good point! My bad

aberonni (approved):

PR #5431, opened 2024-06-18
fix: Prevent redacting(/breaking) payloads that are not objects (strings, etc.)

elwinschmitz:
Fixes what was introduced by: #5397

See [AB#28900](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28900)

### DIFF
--- a/services/121-service/src/shared/services/custom-http.service.ts
+++ b/services/121-service/src/shared/services/custom-http.service.ts
@@ -3,6 +3,7 @@ import { maskValueKeepStart } from '@121-service/src/utils/mask-value.helper';
 import { HttpService } from '@nestjs/axios';
 import { Injectable } from '@nestjs/common';
 import { TelemetryClient } from 'applicationinsights';
+import { isPlainObject } from 'lodash';
 import { catchError, lastValueFrom, map, of } from 'rxjs';
 
 export class Header {
@@ -254,7 +255,7 @@ export class CustomHttpService {
    * @returns - A copy of the input-object with some specific data overwritten/redacted
    */
   private redactSensitiveDataProperties(data: any) {
-    if (typeof data !== 'object' || data === null) {
+    if (!isPlainObject(data)) {
       return data;
     }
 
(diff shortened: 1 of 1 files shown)


services/121-service/src/shared/services/custom-http.service.ts
```diff
@@ -254,6 +255,10 @@ export class CustomHttpService {
    * @returns - A copy of the input-object with some specific data overwritten/redacted
    */
   private redactSensitiveDataProperties(data: any) {
```
aberonni, line 257:
```suggestion
  private redactSensitiveDataProperties(data: unknown) {
```

I imagine this would be safer, but possibly also harder to handle. I'll let you decide if it's worth investigating.

aberonni (approved):
FWIW the AB# link in the description of the PR does not work for me

services/121-service/src/shared/services/custom-http.service.ts
elwinschmitz: I don't have any strong opinion about using `any` over `unknown` (or vise versa).
Its mainly because I don't know what will be thrown at this method, because any request can trigger this.

So partly, I want to 'signal' to the reader of this code "don't expect anything specific, because we cannot know" and at the same time, I don't want to compiler to be too strict all of a sudden that will break any existing code. (code that I can't really check at this point)..

But I _do_ want to manipulate the input (after some checks) and always "almost the same thing"... So I also want to document that.. that what comes out of this method is of the type/shape of "what went into it" (but I couldn't really find how to do that quickly with TypeScript...).

aberonni: Then I think for your use case, a generic like this should work:

```suggestion
  private redactSensitiveDataProperties<T>(data: T): T {
```

I'm going off memory so I might be wrong, but in theory the generic will be inferred automatically by the callee (whenever TS can infer it automatically) and it should work the way you expect it to.

aberonni: Let me know if you would like to look at this together.

elwinschmitz: This does not seem to work/help.

PR #5433, opened 2024-06-18
Re-implement re-issue visa card

diderikvw:
## Describe your changes


services/121-service/src/registration/registrations.service.ts
```diff
@@ -882,37 +887,131 @@ export class RegistrationsService {
...
+      brandCode: intersolveVisaConfig.find(
+        (c) => c.name === FinancialServiceProviderConfigurationEnum.brandCode,
+      )?.value as string, // This must be a string. If it is not, the intersolve API will return an error (maybe).
+      coverLetterCode: intersolveVisaConfig.find(
+        (c) =>
+          c.name === FinancialServiceProviderConfigurationEnum.coverLetterCode,
+      )?.value as string, // This must be a string. If it is not, the intersolve API will return an error (maybe).
+    };
+
+    await this.intersolveVisaService.reissueCard(reissueCardDto);
```
diderikvw, line 993: Instead of creating the DTO object first and then inputting that into the method, in other places I create the object in the method call. According to Dom better, as this then does not need the DTO import and is better readable. I do not necessarily agree, but do think 1 way to do this everywhere is better for readability than various ways throughout the code.

services/121-service/src/registration/repositories/registration-scoped.repository.ts
```diff
@@ -111,9 +112,22 @@ export class RegistrationScopedRepository extends RegistrationScopedBaseReposito
     return this.repository.update(criteria, partialEntity);
   }
 
-  async getRegistrationByReferenceId(referenceId: string) {
+  public async getRegistrationByReferenceId({
```
diderikvw, line 115: Can the signature of this function be simpler? Without optional params and without a "relations" setting?

services/121-service/src/registration/repositories/registration-scoped.repository.ts
```diff
@@ -111,9 +112,22 @@ export class RegistrationScopedRepository extends RegistrationScopedBaseReposito
...
+    programId,
+    relations = [],
+  }: {
+    referenceId: string;
+    programId?: number;
+    relations?: string[];
+  }) {
+    const whereOptions = {
+      referenceId: Equal(referenceId),
+      ...(programId != undefined ? { programId: Equal(programId) } : {}),
```
diderikvw, line 126: Does this not circumvent the security fix for including programId in the query?

services/121-service/src/registration/repositories/registration-scoped.repository.ts
```diff
@@ -111,9 +112,22 @@ export class RegistrationScopedRepository extends RegistrationScopedBaseReposito
...
+    };
     return await this.repository.findOne({
-      where: { referenceId: referenceId },
+      where: whereOptions,
```
diderikvw, line 129: I think putting the where options in-line here would make the code easier to read.

diderikvw (commented):
See comments. Also see the 1st annotation remark, seems relevant.

services/121-service/src/registration/repositories/registration-scoped.repository.ts
PeterSmallenbroek: Well, yes & no. This is a side effect of this method being used in several places/contexts. So to simplify this specific signature we would need define more methods/functions with the needed signatures.

services/121-service/src/registration/repositories/registration-scoped.repository.ts
PeterSmallenbroek: I checked in code and apparently there's situations where we don't (want?) to do this. So that's why I made optional.

services/121-service/src/registration/repositories/registration-scoped.repository.ts
diderikvw: Yes, indeed. IMO that would be cleaner (i.e. simpler to understand and less likely to introduce a bug/security risk).
So something like various functions with names like:
getRegistrationAnd...(relation name)
getRegistrationByReferenceIdAndProgramId

The current method can become private and used by these various public methods, to avoid code duplication.

PR #5445, opened 2024-06-21
fix: typeorm undefined value return AB#28797 AB#28796

Vladimir-Petkov:
## Describe your changes


services/121-service/src/registration/services/registrations-bulk.service.spec.ts
```diff
@@ -34,18 +34,23 @@ describe('RegistrationBulkService', () => {
...
+          programIdValue === programId &&
+          typeValue === RegistrationStatusEnum.registered
+        ) {
           return Promise.resolve({
             id: 1,
             name: 'test',
             language: LanguageEnum.en,
-            type: 'test',
-            message: 'test',
+            type: RegistrationStatusEnum.registered,
```
aberonni, line 48: Why this change? It seems like this is less explicit and more prone to failures but I'm guessing there is a reason I cannot see. I'm also surprised this doesn't break any existing tests 🤔

services/121-service/src/scripts/seed-helper.ts
```diff
@@ -291,8 +291,13 @@ export class SeedHelper {
...
   public async assignAdminUserToProgram(programId: number): Promise<void> {
     const userRepository = this.dataSource.getRepository(UserEntity);
+    const adminEmail = process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN
+      ? Equal(process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN)
+      : undefined;
     const adminUser = await userRepository.findOneOrFail({
-      where: { username: process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN },
+      where: {
+        username: adminEmail,
+      },
```
aberonni, line 300: I think it would be safer to assume this is always defined. Also, the point of this change is to never assign `undefined` inside the `where` clause to avoid the [original bug](https://github.com/typeorm/typeorm/issues/2500#issuecomment-1916774575) (which is why we are making these changes in the first place).

```suggestion
      where: {
        username: Equal(process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN!),
      },
```

services/121-service/src/utils/scope/createFindWhereOptions.helper.spec.ts
```diff
@@ -37,7 +40,34 @@ describe('createFindWhereOptions helper', () => {
...
+      if (typeof obj === 'object' && obj !== null) {
+        return Object.fromEntries(
+          Object.entries(obj).map(([key, value]) => [
+            key,
+            transformToComparableForm(value),
+          ]),
+        );
+      }
+      return obj;
+    };
```
aberonni, line 60: I think this function should exist in a more generic place or in its own file

services/121-service/src/notifications/message-incoming/message-incoming.service.ts
```diff
@@ -145,8 +145,12 @@ export class MessageIncomingService {
...
     ) {
+      const sidCondition = callbackData.SmsSid
+        ? Equal(callbackData.SmsSid)
+        : undefined;
+
       const tryWhatsapp = await this.tryWhatsappRepository.findOne({
-        where: { sid: callbackData.SmsSid },
+        where: sidCondition ? { sid: sidCondition } : {},
         relations: ['registration'],
       });
```
aberonni, line 155: This doesn't look safe. I don't think it will return `null` if the object is empty (due to the typeofm bug). Instead, I think `this.tryWhatsappRepository.findOne` should be called conditionally based on whether `callbackData.SmsSid` is defined.

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
```diff
@@ -200,9 +200,13 @@ export class WhatsappService {
...
     callbackData: TwilioStatusCallbackDto,
   ): Promise<void> {
+    const sidCondition = callbackData.SmsSid
+      ? Equal(callbackData.SmsSid)
+      : undefined;
+
     const tryWhatsappTemplateEntity =
       await this.whatsappTemplateTestRepository.findOne({
-        where: { sid: callbackData.SmsSid },
+        where: sidCondition ? { sid: sidCondition } : {},
```
aberonni, line 209: Same comment as in `message-incoming.service.ts`

aberonni (changes requested):
Beside the PR comments, make sure to follow the [contribution guidelines](https://github.com/global-121/121-platform/blob/4c15a7a2d1f1f51dd208e95597e13f3b47602b3c/docs/CONTRIBUTING.md#submitting-changes) and to update your branches via rebase and not via merge.

aberonni (approved):

PR #5471, opened 2024-06-25
chore: add instance/version endpoint to mock-service

aberonni:
[AB#28880](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28880)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ This project uses the [`CalVer`](https://calver.org/#scheme)-format: `YY.MM-MICR
 
 - Version-format changed from `SemVer`("v`MAJOR`.`MINOR`.`PATCH`"; i.e. `v123.0.0`) to `CalVer`("v`YY`.`MM`-`MICRO`"; i.e `v24.6-0`)
 
+### Added
+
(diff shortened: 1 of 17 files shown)


elwinschmitz (approved):

elwinschmitz (changes requested):
Oh, wait.

The magic that happens in `package.json` in the 121-service also needs to be used here...

elwinschmitz:
This magic: https://github.com/global-121/121-platform/blob/main/services/121-service/package.json#L15-L16

It needs to grab the version-number from the build (text-)file at start-up/build. (I'm not sure which of these 2 is _really_ used... but I guess its copied to the build-step out of caution(?) )

elwinschmitz:
This new endpoint could also be used as a 'healthcheck' maybe.. (for lack of any better alternative):

https://github.com/global-121/121-platform/blob/main/services/docker-compose.yml#L47-L48

elwinschmitz:
> https://github.com/global-121/121-platform/blob/main/services/docker-compose.yml#L47-L48

Also about that line; we should make that port-number configurable; like how `PORT_121_SERVICE` is handled.

arsforza (approved):

PR #5477, opened 2024-06-26
Update Dependabot config to include Portalicious

elwinschmitz:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -116,11 +116,11 @@ updates:
     open-pull-requests-limit: 5
     versioning-strategy: increase
     groups:
-      dependencies:
-        patterns:
-          - "*"
       angular:
         patterns:
           - '@angular/*'
           - '@angular-devkit/*'
           - 'zone.js'
+      dependencies:
+        patterns:
+          - '*'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,6 +17,7 @@ updates:
     groups:
       development:
         patterns:
+          - '@swc/*'
           - '@types/*'
           - '*eslint*'
           - '*prettier*'
@@ -45,6 +46,7 @@ updates:
     groups:
       development:
         patterns:
+          - '@swc/*'
           - '@types/*'
           - '*eslint*'
           - '*prettier*'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -107,3 +107,23 @@ updates:
         update-types:
           ['version-update:semver-major', 'version-update:semver-minor']
       - dependency-name: 'zone.js'
+
+  - package-ecosystem: npm
+    directory: '/interfaces/Portalicous'
+    schedule:
+      interval: weekly
+      day: tuesday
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -105,3 +107,23 @@ updates:
...
+      day: tuesday
+    open-pull-requests-limit: 5
+    versioning-strategy: increase
+    groups:
+      development:
+        patterns:
+          - '@types/*'
+          - '*eslint*'
+          - '*prettier*'
+          - 'karma*'
```
aberonni, line 124: I've been wondering whether we could get away with just grouping everything to avoid having to approve/merge/rebase so many PRs.

For example in my projects I [usually do](https://github.com/aberonni/improvdb/blob/main/.github/dependabot.yml#L13-L15):

```suggestion
    groups:
      dependencies:
        patterns:
          - "*"
```

If not that, then maybe we could at least add a catch-all group
```suggestion
    groups:
      development:
        patterns:
          - '@types/*'
          - '*eslint*'
          - '*prettier*'
          - 'karma*'
          - '*'
```

aberonni (dismissed):
Curious to hear your thoughts on my comment, but I'm ok with it being merged as-is if there are reasons

.github/dependabot.yml
elwinschmitz: We could try it for a while and see how annoying/convenient it is..

The main separation for me is on the "can blindly merge if tests are green"-things (like the `eslint-*`-stuff) and the "Angular stuff needs to be updated together manually anyway"-stuff, etc.

I have created/used a specific workflow for those Angular updates: https://github.com/rodekruis/helpful-information/blob/main/.github/workflows/upgrade-dependencies.yml
But its not a silver bullet...

elwinschmitz: There will be some issues with some packages that have a different understanding of "breaking changes in a minor version" and with their matching peer-dependencies...

But starting with `*` and only adding the exceptions-when-necessary seems like a good approach. 👍

.github/dependabot.yml
```diff
@@ -105,3 +107,20 @@ updates:
...
+    directory: '/interfaces/Portalicous'
+    schedule:
+      interval: weekly
+      day: tuesday
+    open-pull-requests-limit: 5
+    versioning-strategy: increase
+    groups:
+      dependencies:
+        patterns:
+          - "*"
```
elwinschmitz, line 121:
```suggestion
          - '*'
```

.github/dependabot.yml
```diff
@@ -105,3 +107,20 @@ updates:
...
+      angular:
+        patterns:
+          - '@angular/*'
+          - '@angular-devkit/*'
+          - 'zone.js'
```
elwinschmitz, line 126:
```suggestion
    groups:
      angular:
        patterns:
          - '@angular/*'
          - '@angular-devkit/*'
          - 'zone.js'
      dependencies:
        patterns:
          - '*'
```

aberonni (approved):
I like it!

PR #5481, opened 2024-06-27
chore: add generic to redactSensitiveDataProperties

aberonni:
## Describe your changes

Following up on this discussion: https://github.com/global-121/121-platform/pull/5431#discussion_r1647529254

I wanted to add an example of how to use generics in these circumstances.

### DIFF
--- a/services/121-service/src/shared/services/custom-http.service.ts
+++ b/services/121-service/src/shared/services/custom-http.service.ts
@@ -254,7 +254,7 @@ export class CustomHttpService {
    * @param data - Any key-value object
    * @returns - A copy of the input-object with some specific data overwritten/redacted
    */
-  private redactSensitiveDataProperties(data: any) {
+  private redactSensitiveDataProperties<T extends object>(data: T): T {
     if (!isPlainObject(data)) {
       return data;
     }
@@ -274,7 +274,10 @@ export class CustomHttpService {
     }
 
     // Explicitly mask the username/email:
-    if (redactedData.username) {
+    if (
+      'username' in redactedData &&
+      typeof redactedData.username === 'string'
+    ) {
       redactedData.username = maskValueKeepStart(redactedData.username, 3);
     }
 
--- a/services/121-service/src/shared/services/custom-http.service.ts
+++ b/services/121-service/src/shared/services/custom-http.service.ts
@@ -254,7 +254,7 @@ export class CustomHttpService {
    * @param data - Any key-value object
    * @returns - A copy of the input-object with some specific data overwritten/redacted
    */
-  private redactSensitiveDataProperties<T extends object>(data: T): T {
+  private redactSensitiveDataProperties<T>(data: T): T {
     if (!isPlainObject(data)) {
       return data;
     }
@@ -275,6 +275,8 @@ export class CustomHttpService {
 
     // Explicitly mask the username/email:
     if (
+      redactedData &&
+      typeof redactedData === 'object' &&
       'username' in redactedData &&
       typeof redactedData.username === 'string'
     ) {


services/121-service/src/shared/services/custom-http.service.ts
```diff
@@ -254,7 +254,7 @@ export class CustomHttpService {
    * @param data - Any key-value object
    * @returns - A copy of the input-object with some specific data overwritten/redacted
    */
-  private redactSensitiveDataProperties(data: any) {
+  private redactSensitiveDataProperties<T extends object>(data: T): T {
```
elwinschmitz, line 257: Hmm. But `data` is sometimes also “only a string”…

Would that create a runtime-issue?


Or would we need to add a `| number | string | boolean` somewhere?

aberonni: > Hmm. But data is sometimes also “only a string”…

I would've expected a compiler error in that case, but now I see that the payloads coming in are also typed as `any` so that doesn't help lol

> Would that create a runtime-issue?

No, it wouldn't. I created a [TS playground](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAJwKYBMCG0DKqwDOMsAbqgCKZSYAKycADqsrKgQDwAqiqAHlPnQFEcAEYArVNAB8ACizUAXIk4BKZdwDeAKESIICAlBQZsA9JWqIAvIk2IAdE4WZEAXwDciAPTfEOAAtMABtguAB3fUYAT0QoOERMEjgYdEQAWxBqWDAAcziA1BFkGFyYMBCRCSkobV1EGGBEWXq9AHIQAmYK9NQ2hqQ0LGgMS1cAMnHWuOimOCahs1GqTAdO7sxem2tbNqMSvLb61TtpxZGLFbWu5B6i3YI4XqgA8ty2j3q3Or00KBBbogXJ9vtooLMigBZaIAGWIUGCqAASqgAI4gNjGWw6PQMTDRMKYdDKTBgaIguraAyEYzIZTQuFQBHItEYow2U64-GE4mIPZPVAvN5HUHUx6IhxhXKyNqcTFtVSfMVwCVS2TnKB4QjEGBkMZ0RjMVgEdUOPEEuBE1SKoA) to mimic this behaviour. But, I still don't like the way this is typed now.

> Or would we need to add a | number | string | boolean somewhere?

Potentially, yes. But I'd rather remove `extends object` from the generic instead. I've pushed a change that takes this into account.

services/121-service/src/shared/services/custom-http.service.ts
```diff
@@ -274,7 +274,12 @@ export class CustomHttpService {
     }
 
     // Explicitly mask the username/email:
-    if (redactedData.username) {
+    if (
+      redactedData &&
+      typeof redactedData === 'object' &&
```
aberonni, line 279: Ideally, this would be inferred automatically with the `isPlainObject` check, but that function would have to return some information that would help TS infer it (which it currently doesn't).

elwinschmitz: Haha, my whole point of using that Lodash-helper was to avoid these verbose type-check-syntax-things.. :)

There are types installed for Lodash (using `@types/lodash`), but apparently that doesn't do the trick?
There will probably be some (TS-specific/native) alternative to the `isPlainObject`-check, right?

aberonni: > There are types installed for Lodash (using @types/lodash), but apparently that doesn't do the trick?

Nope, the [lodash function only returns a boolean](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/880e19374f4034deeacd43785eac84d52b197e00/types/lodash/common/lang.d.ts#L1115-L1138) rather than [using a type-guard](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates).

I assume that is the case also because 

> There will probably be some (TS-specific/native) alternative to the isPlainObject-check, right?

Maybe `typeof data === 'object'`? But I'm not entirely sure what we're trying to achieve/avoid.

PR #5486, opened 2024-06-27
poc: sharing types between BE and FE [simple]

aberonni:
If you're interested in digging deeper, the main relevant change is the one called out in the comments. Most of the other changes inside the `121-service` and `Portal` folders were necessary just to get the Portal compiler not to complain about importing complex things from `@121-service`.

## Dom's opinion: ✅ Recommended solution (for Portalicious)

Please look at the comments in the PR for more details.

Find all of the related PRs to this PoC in the related task: [AB#28945](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28945).

### DIFF
--- a/interfaces/Portal/package.json
+++ b/interfaces/Portal/package.json
@@ -26,7 +26,8 @@
     "test:all": "npm run lint && npm test",
     "test:dev": "ng test",
     "test": "ng test --configuration=ci",
-    "upgrade:angular": "ng update @angular/core @angular/cli @angular-eslint/schematics --create-commits"
+    "upgrade:angular": "ng update @angular/core @angular/cli @angular-eslint/schematics --create-commits",
+    "postinstall": "cd ../../services/121-service && npm ci --omit=optional --no-fund --no-audit"
   },
   "private": true,
   "engines": {
--- a/interfaces/Portal/package.json
+++ b/interfaces/Portal/package.json
@@ -27,7 +27,7 @@
     "test:dev": "ng test",
     "test": "ng test --configuration=ci",
     "upgrade:angular": "ng update @angular/core @angular/cli @angular-eslint/schematics --create-commits",
-    "postinstall": "cd ../../services/121-service && npm ci --omit=optional --no-fund --no-audit"
+    "postinstall": "cd ../../services/121-service && npm run install:dependencies-for-portal"
   },
   "private": true,
   "engines": {
--- a/interfaces/Portal/src/app/models/user.model.ts
+++ b/interfaces/Portal/src/app/models/user.model.ts
@@ -1,14 +1,7 @@
-import Permission from '../auth/permission.enum';
+import { UserController } from '@121-service/src/user/user.controller';
+import { Dto121Service } from '../shared/utils/dto-type';
 
-export class User {
-  username: string;
-  permissions: {
-    [programId: number]: Permission[];
-  };
-  expires?: string;
-  isAdmin?: boolean;
-  isEntraUser?: boolean;
-}
+export type User = Dto121Service<UserController['login']>['user'];
 
 export enum UserType {
   admin = 'admin',
--- a/interfaces/Portal/src/app/shared/utils/dto-type.ts
+++ b/interfaces/Portal/src/app/shared/utils/dto-type.ts
@@ -1,0 +1,73 @@
+/**
+ * This file contains a utility to convert a DTO type from a service controller method
+ * to a type that can be used in the frontend.
+ *
+ * It is based on: https://github.com/tamj0rd2/dto/blob/master/src/dto.ts
+ *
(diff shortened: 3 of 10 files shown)


interfaces/Portal/src/app/models/user.model.ts
```diff
@@ -1,14 +1,7 @@
...
-export class User {
-  username: string;
-  permissions: {
-    [programId: number]: Permission[];
-  };
-  expires?: string;
-  isAdmin?: boolean;
-  isEntraUser?: boolean;
-}
+export type User = Dto121Service<UserController['login']>['user'];
```
aberonni, line 4: This is the main change, and the one that would need to be applied in the rest of the codebase.

As such, I would not recommend actually doing it in `Portal`, and to simply start `Portalicious` with this approach instead.

interfaces/Portal/src/app/shared/utils/dto-type.ts
```diff
@@ -0,0 +1,73 @@
+/**
+ * This file contains a utility to convert a DTO type from a service controller method
+ * to a type that can be used in the frontend.
+ *
+ * It is based on: https://github.com/tamj0rd2/dto/blob/master/src/dto.ts
+ *
+ * For more info: https://dev.to/tamj0rd2/dto-a-typescript-utility-type-4o3m
```
aberonni, line 7: Kudos to Elwin for the tip on this article that turned out very handy.

services/121-service/src/config.ts
```diff
@@ -1,5 +1,3 @@
-import packageJson = require('../package.json');
```
aberonni, line 1: I made this change because somewhere it was complaining about this `require`. I figured we weren't missing out on much without it.

interfaces/Portal/package.json
```diff
@@ -25,7 +25,8 @@
     "test:all": "npm run lint && npm test",
     "test:dev": "ng test",
     "test": "ng test --configuration=ci",
-    "upgrade:angular": "ng update @angular/core @angular/cli @angular-eslint/schematics --create-commits"
+    "upgrade:angular": "ng update @angular/core @angular/cli @angular-eslint/schematics --create-commits",
+    "postinstall": "cd ../../services/121-service && npm ci --omit=optional --no-fund --no-audit"
```
aberonni, line 29: This is necessary to be able to use the full `@121-service` imports (mainly in CI). Without it, the Portal compiler will complain when we reference controllers and such, with errors such as `typeorm could not be found`.

interfaces/Portal/src/app/models/user.model.ts
elwinschmitz: Ah, so it would have a similar 'workaround' like we're already using the `PermissionsEnum`-type from the 121-service, with a sort-of "local proxy"-file...

So components/services in the Portal(icious) would only have to worry about importing this local model/type-file. And only in those "proxy"-files, we would have the "import from `@121-service`-magic..

interfaces/Portal/package.json
elwinschmitz: Aha, because only some "magical mapping aliasses" wont work for the "built-in types" of some TypeORM-related things, right?

Feels a bit weird to 'hard-code' this install-loop in this way... But otherwise we would need to document it somewhere anyway... AND we would need to perform the correct install-steps in any automated test-workflow...

But there is still this risk of "the npm install for the 121-services will run multiple times in a test-run", I think...
And also a "build" of the Portal will require also to "install all dependencies of the 121-service". :(

elwinschmitz:
```suggestion
    "postinstall": "cd ../../services/121-service && npm install"
```

I would let the 121-service manage its own "install-step"; so we don't have to keep any configuration-flags etc in-sync..

interfaces/Portal/src/app/models/user.model.ts
aberonni: Exactamundo 👌

interfaces/Portal/package.json
aberonni: > Aha, because only some "magical mapping aliasses" wont work for the "built-in types" of some TypeORM-related things, right?

Yup

> Feels a bit weird to 'hard-code' this install-loop in this way... But otherwise we would need to document it somewhere anyway... AND we would need to perform the correct install-steps in any automated test-workflow...

Yes, my thinking exactly. This has effectively become part of the installation process of the Portal at this point.

> But there is still this risk of "the npm install for the 121-services will run multiple times in a test-run", I think...
And also a "build" of the Portal will require also to "install all dependencies of the 121-service". :(

Yes and yes, but I can't think of a way around that.

> Or, to make this "feature" more explicit to both the Portal-developers AND the 121-service-developers:

I like this suggestion 👍

PR #5490, opened 2024-06-27
poc: sharing types between BE and FE [openapi-generator]

aberonni:
If you're interested in digging deeper, the main entry point for looking at the changes should be `openapi-generator.sh`. From here, you can see that everything within the `121-service-angular-api` folder has been auto-generated and linked to the Portal.

## Dom's opinion: ❌ Not a valid solution

Please look at the comments in the PR for more details on the reasoning.

Find all of the related PRs to this PoC in the related task: [AB#28945](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28945).


interfaces/Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -36,26 +36,29 @@ import { arrayToXlsx } from '../shared/array-to-xlsx';
...
-      )
+    return this.usersService
+      .userControllerLogin({
+        username,
+        password,
+      })
+      .toPromise()
       .then((response) => {
+        // this works
+        // but response is typed as 'any' 🙈
```
aberonni, line 61: The biggest shortcoming of this approach - it doesn't actually infer types properly. You can see that this is explicitly mis-typed in [`users.service.ts`](https://github.com/global-121/121-platform/pull/5490/files#diff-684af501228b7f3a61c11b131a41436e13c7e73ab9ccde2f16f07ee6390ce28fR517-R520), where the return type is typed as `any`.

So it kind of defeats the purpose of doing this whole thing.

openapi-generator.sh
```diff
@@ -0,0 +1,27 @@
+# Before running this:
```
aberonni, line 1: Currently this only works locally. We would have to figure out how/when to run this in CI, and ideally maybe even publish the npm module rather than linking it locally?

interfaces/Portal/src/app/app.module.ts
```diff
@@ -100,6 +114,7 @@ if (environment.use_sso_azure_entra) {
       innerHTMLTemplatesEnabled: true,
     }),
     AppRoutingModule,
+    ApiModule.forRoot(apiConfigFactory),
```
aberonni, line 117: Another big shortcoming is that this only seems to work with apps that use an AppModule, but that isn't the angular v18 approach AFAIU

PR #5501, opened 2024-06-28
fix: 28979 export only last 5 payments instead of all payments which actually failed due to large number of rows

ashishkumarbasak-accenture:
## Describe your changes

fix: 28979 export only last 5 payments instead of all payments which actually failed due to large number of rows

Task: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28979


services/121-service/package.json
```diff
@@ -13,8 +13,8 @@
     "lint": "eslint \"**/*.{ts,js}\"",
     "fix": "npm run lint -- --fix",
     "build": "nest build -p tsconfig.json",
-    "start": "GLOBAL_121_VERSION=$(test -f \"VERSION.txt\" && cat VERSION.txt || echo $GLOBAL_121_VERSION)  node dist/main.js",
-    "start:dev": "nodemon --watch src/**/*.json --exec npm run start:dev:nest",
+    "start": "GLOBAL_121_VERSION=$(test -f \"VERSION.txt\" && cat VERSION.txt || echo $GLOBAL_121_VERSION)  node --max-old-space-size=4096 dist/main.js",
+    "start:dev": "NODE_OPTIONS='--max-old-space-size=4096' nodemon --watch src/**/*.json --exec npm run start:dev:nest",
```
RubenGeo, line 17: Do you think it makes sense to put this in a separate PR? I am not 100% sure we should immediately release this to production before testing it on our test servers. And it is not needed for this immidiate fix. 

I read you can also put this in the .npmrc file  https://stackoverflow.com/a/75793324

interfaces/Portal/src/assets/i18n/en.json
```diff
@@ -257,7 +257,7 @@
         "titlePortal": "Title portal"
       },
       "program-payout": {
-        "all-closed-payments": "All closed payments",
+        "all-closed-payments": "Last 5 closed payments",
```
RubenGeo, line 260: I think this key also needs an update `all-closed-payments` to something like `last-closed-payments` or something along those lines

Could you also do the dutch translation in the nl.json: 'De laatste 5 gesloten betalingen'

RubenGeo (changes requested):

interfaces/Portal/src/assets/i18n/en.json
RubenGeo: Ah also maybe the 5 could be dynamic in the string?

services/121-service/package.json
ashishkumarbasak-accenture: Yes, we can put it in `.npmrc` as well.
OK I think better to remove this part from this PR. Removing!

RubenGeo (approved):

PR #5508, opened 2024-07-01
test parallel execution update

Piotrk39:
## Describe your changes

I would like to add another second test container to run e2e tests in parallel. I have already split the the test execution between the two. It is kind of proof of concept for splitting the tests. Main goal is to speed up test execution and to minimise time of re-run in case something goes wrong.

Question to reviewer:

How many test containers should run in the same time?

### DIFF
--- a/.github/workflows/test_e2e-parallel-1.yml
+++ b/.github/workflows/test_e2e-parallel-1.yml
@@ -23,9 +23,6 @@ env:
 jobs:
   test:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        partition: [0, 1, 2]
 
     steps:
       - name: Check out repository code
@@ -54,7 +51,7 @@ jobs:
       - name: Run Services with Docker
         working-directory: ./services
         run: |
-          docker-compose -p service-${{ matrix.partition }} up -d --build
+          docker --log-level 'warn' compose -f docker-compose.yml up -d --quiet-pull --wait --wait-timeout 300
 
       - name: Start Portal
         working-directory: ./interfaces/Portal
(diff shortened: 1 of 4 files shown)


elwinschmitz (changes requested):
I think there are other ways to achieve the "parallel-processes" without having to duplicate a lot of code. ;)

I don't know if I'll get back to this quickly though... (Ping me later today. 🤞 )

aberonni (commented):
I agree with Elwin.

In general, the approach works, but it doesn't scale very well. For example, what do we do when we have 1000 tests? 100 yaml files? And how do we make sure we add tests to the yaml files?

Ideally this would be solved more systematically with something like [testcontainers](https://testcontainers.com/)

elwinschmitz:
I think we can come a long way with:
https://docs.github.com/en/actions/using-containerized-services/about-service-containers
and https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container

So my plan would be something like:
- first job: install + build API+Interface
- second job: (needs: first) list all test-scenario's (Length = X), divide list in chuncks/batches of `N` (=X/?) tests.
- third job: (needs: second) Runs first batch of `N` tests
- `Y`nth job: (needs: second) Runs `Y`nth batch of `N` tests
- `Y+1`nth job: (needs: second) Runs `Y+1`nth batch of `N` tests
- ...

This has a similar "scaling up" issue, but maybe there are ways to run jobs as a loop(?), so we don't have to hard-code the chunck-size and iteration-count etc.

Ideally the chuncks-of-tests would be randomly created and run in a different order every time...
Their might be (posiibly, very likely) already tooling available to manage this; (but I don't know it yet)... but it could also be only a small (node/bash-)script that we write ourselves... (but it wouldn't be the most future-proof and stable solution)

PR #5511, opened 2024-07-01
AB#28249 View card visa re-implementation

RubenGeo:
## Describe your changes


services/121-service/src/payments/fsp-integration/intersolve-visa/mappers/intersolve-visa-status.mapper.ts
```diff
@@ -115,15 +101,15 @@ export class IntersolveVisaStatusMappingService {
...
     actions: string,
     linkCreationInfo: VisaCardLinkCreationInfo,
-  ): VisaCardActionLink[] {
+  ): VisaCardActionLinkDto[] {
     if (!actions) {
       return [];
     }
 
-    const links: VisaCardActionLink[] = [];
+    const links: VisaCardActionLinkDto[] = [];
```
RubenGeo, line 112: So in our old code we used to have links according to https://en.wikipedia.org/wiki/HATEOAS, but this functionality is
-  currently not really used by our frontend
- we never use this HATEOAS somewhere else
- complicates these mapper functions because programId and referenceId need to be passes down through all functions
So it does not add a lot of value in my opinion.

diderikvw: @RubenGeo 
These hyperlinks SHOULD have been used by our front-end :)
But if they are not, then how does the front-end currently decide which actions to show?

RubenGeo: image
It only uses the actions part and not the href or method. The href and method are basically stored here where we store all our api requests interfaces/Portal/src/app/services/programs-service-api.service.ts

RubenGeo:
moved here #5553

PR #5513, opened 2024-07-01
fix: 28984 Fixed memory issue for export large number of rows (~ 1.5 millions) in excel 

ashishkumarbasak-accenture:
Task: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28984

## Describe your changes

Currently we are getting out of heap memory allocation of node processor during export large number of rows in a excel file. This changes will allow to export excel with around 1.5 millions of rows at a time with multiple sheets (as a chunk).


services/121-service/src/utils/send-xlsx-response.ts
```diff
@@ -1,5 +1,11 @@
 import { Response } from 'express';
 import * as XLSX from 'xlsx';
+import * as zlib from 'zlib';
+
+// Use compression and chunks to write into xls
+// for each data sheet of xlsx file to avoid memory error
+const CHUNK_SIZE = 300000;
```
RubenGeo, line 7: Hey, I was wondering about this. Does this CHUNK_SIZE corresponds to the number of rows? If so than is this chunk_size of 300.000 not too large since we only support programs of up to 100.000. Or do I misunderstand this @ashishkumarbasak-accenture

services/121-service/package.json
```diff
@@ -13,8 +13,8 @@
     "lint": "eslint \"**/*.{ts,js}\"",
     "fix": "npm run lint -- --fix",
     "build": "nest build -p tsconfig.json",
-    "start": "GLOBAL_121_VERSION=$(test -f \"VERSION.txt\" && cat VERSION.txt || echo $GLOBAL_121_VERSION)  node dist/main.js",
-    "start:dev": "nodemon --watch src/**/*.json --exec npm run start:dev:nest",
+    "start": "GLOBAL_121_VERSION=$(test -f \"VERSION.txt\" && cat VERSION.txt || echo $GLOBAL_121_VERSION) node --max-old-space-size=4096 dist/main.js",
+    "start:dev": "NODE_OPTIONS='--max-old-space-size=4096' nodemon --watch src/**/*.json --exec npm run start:dev:nest",
```
RubenGeo, line 17: @elwinschmitz Do you see any concerns with increasing the memory size?

services/121-service/src/utils/send-xlsx-response.ts
RubenGeo: Do you know how long an export now takes with 1.5 million rows?

ashishkumarbasak-accenture: @RubenGeo This is chunk size to write into excel file data sheet, where each data sheet will contain 300,000 rows. So if we have 700,000 rows, then it will create 3 data sheets in the excel file.

I have tested with chunk size 500,000 rows and it works fine as well, but to be on safe side I made it a little smaller value, 300,000. If you want I can make it to 500,000 as chunk size, but small value than 300,000 will help use to write the file, but in that case the excel file will have many sheets.

Performance wise on my local machine it took around ~ 4 min to complete.

services/121-service/package.json
aberonni: The [docs on this option](https://nodejs.org/docs/latest-v18.x/api/cli.html#--max-old-space-sizesize-in-megabytes) might help in terms of understanding the recommendations around this.

aberonni, line 17: Following up from my question from standup: how did you come up with `4096` specifically? Trial and error? Or...?

elwinschmitz: I wonder how much 'room' there is on the Azure-side of this.

Because I'd think the reality in the local Docker-setup will be different from the production-Azure-App-Services setup...

I can't easily spot the existing memory-use... But the whole "App Service Plan" seem to have 16Gb available (shared between all production-instances)...

ashishkumarbasak-accenture: Actually I got that we have 16GB of shared memory on azure server, So honestly speaking I tried to assign 1/4 of it and then do some tests to adjust by using trial and error. During tests, got that, for 1.5 millions 4096 is enough. I also tried with 3072 but was not enough. 

And I got in our NLRC has ~ 700,000 rows that failed during export, so wanted to make it work for double number of rows ~1.5 millions.

services/121-service/src/utils/send-xlsx-response.ts
```diff
@@ -19,14 +25,23 @@ export function sendXlsxReponse(
...
+  const workbook = XLSX.utils.book_new();
+
+  let dataSheet = 1;
+  for (let index = 0; index < array.length; index += CHUNK_SIZE) {
+    const chunk = array.slice(index, index + CHUNK_SIZE);
+    const worksheet: XLSX.WorkSheet = XLSX.utils.json_to_sheet(chunk, {
+      dense: true,
+    });
+    XLSX.utils.book_append_sheet(workbook, worksheet, `data_${dataSheet++}`);
+  }
```
elwinschmitz, line 37: The performance bottle-neck seem to be here with "parsing the array into a rows-in-a-sheet", right?

And the 'workaround' is to have multiple sheets in the workbook.

Would it also be helping (memory-wise) to do the conversion from JSON-to-sheet in a loop and append all to the same worksheet?

(Following some example-code from: https://docs.sheetjs.com/docs/api/utilities/array#array-of-objects-input )

So, something like:
```suggestion
  const workbook = XLSX.utils.book_new();
  const worksheet = XLSX.utils.aoa_to_sheet([[]]); // Empty worksheet

  array.forEach((row) => {
    XLSX.utils.sheet_add_json(worksheet, row, {
      origin: -1,
    });
  });
  
  XLSX.utils.book_append_sheet(workbook, worksheet);
```

(👆 Not tested in any way.)

ashishkumarbasak-accenture: Let me test it.

services/121-service/package.json
aberonni: I imagine we can assign more than 4GB if the total is 16GB - but I'm not the expert here.

I'm also wondering, do you know what the default is without setting this flag? The docs doesn't say anything about this.

services/121-service/src/utils/send-xlsx-response.ts
ashishkumarbasak-accenture: Its not helping (not working, giving error on `Object.sheet_add_json`), not sure whats the actual reason.

elwinschmitz: Could have something to do with how the sheet-headers are handled?

It the example they 'manually' add the first row etc...

services/121-service/package.json
elwinschmitz: @aberonni I could only find documentation about the Windows-version of Azure App Services:

https://learn.microsoft.com/en-us/azure/app-service/configure-custom-container?pivots=container-windows&tabs=debian#customize-container-memory

There the defaults are mentioned as `1536` Mb for our `P2v3`-tier.

Underneath that table is:

> For example, in an App Service plan with 8 GB RAM, the cumulative total of `WEBSITE_MEMORY_LIMIT_MB` for all the apps must not exceed 8 GB.

So that would limit us to "only make an exception to this limit for a few clients, at the expense of the others". Could still be worth it, for now. But not a good, solid path going forward, I'd think.

But we could try this out "in production" on the Staging-environment...
To see what actually happens...

elwinschmitz: Another idea:

We could also set these ENV-options _only_ in the NLRC-AppService...
Or divide the 16Gbs up into "unequal"-chunks and set the ENV-options on all AppServices (but with different numbers for each)...

That way, we don't have to hard-code anything for production...

PR #5525, opened 2024-07-02
portalicious: build proof of concept with angular-query-experimental for Portalicious

aberonni:
[AB#28824](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28824)

A lot of the changes here are being carried over from #5486 

- [ ] ~I do not need any deviation from our PR guidelines~ I'd like a review from both @arsforza and @elwinschmitz before merging.

### DIFF
--- a/interfaces/Portalicious/src/app/components/health-widget/health-widget.component.ts
+++ b/interfaces/Portalicious/src/app/components/health-widget/health-widget.component.ts
@@ -1,0 +1,24 @@
+import { JsonPipe } from '@angular/common';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
+import {
+  injectQuery,
+  injectQueryClient,
+} from '@tanstack/angular-query-experimental';
+import { ApiEndpoints, ApiService } from '~/services/api.service';
+
+@Component({
+  selector: 'app-health-widget',
+  standalone: true,
+  imports: [JsonPipe],
+  templateUrl: './health-widget.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class HealthWidgetComponent {
+  apiService = inject(ApiService);
+  queryClient = injectQueryClient();
+
+  versionInfo = injectQuery(() => ({
+    queryKey: [ApiEndpoints.versionInfo],
+    queryFn: () => this.apiService.getVersionInfo(),
+  }));
+}
--- a/interfaces/Portalicious/src/app/services/api.service.ts
+++ b/interfaces/Portalicious/src/app/services/api.service.ts
@@ -1,0 +1,21 @@
+import { Injectable, inject } from '@angular/core';
+import { VersionInfo } from '~/models/health.model';
+import { HttpWrapperService } from '~/services/http-wrapper.service';
+
+export enum ApiEndpoints {
+  versionInfo = 'health/version',
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ApiService {
+  private httpWrapperService = inject(HttpWrapperService);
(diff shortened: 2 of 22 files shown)


interfaces/Portalicious/tsconfig.json
```diff
@@ -3,16 +3,22 @@
...
+      "~/*": ["./src/app/*"]
     },
     "outDir": "./dist/out-tsc",
     "strict": true,
+    // TODO: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28804
+    "strictPropertyInitialization": false,
+    // TODO: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/28419
+    "noImplicitAny": false,
+    // TODO: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/27718
+    "useUnknownInCatchVariables": false,
```
aberonni, line 19: I am forced to do this because otherwise, the imports from the 121-service don't work, and the compiler complains. That's because these flags are set to false in the 121-service, so there is lots of code that doesn't comply with these rules.

For things like explicitly using `any`, we have eslint keeping us in check. I would still rather not do this. I spent an hour trying to figure out how to avoid it to no avail. Maybe it's worth some pair programming.
 
Having to make this change makes me very sad.

interfaces/Portalicious/src/app/services/api.service.ts
```diff
@@ -0,0 +1,133 @@
...
+      throw error;
+    }
+  }
+
+  getVersionInfo() {
+    return this.performRequest<VersionInfo>(
+      'GET',
+      `${environment.url_121_service_api}/health/version`,
+    );
+  }
```
aberonni, line 132: My thinking is that given that `api.service` can be radically simplified compared to what is in the Portal (by removing method-specific functions and repetition), we could have a single `ApiService` that does what the `ProgramApiService` does on the existing Portal.

interfaces/Portalicious/src/app/components/health-widget/health-widget.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  imports: [JsonPipe],
+  templateUrl: './health-widget.component.html',
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class HealthWidgetComponent {
+  apiService = inject(ApiService);
+  queryClient = injectQueryClient();
+
+  versionInfo = injectQuery(() => ({
+    queryKey: [ApiEndpoints.versionInfo],
```
aberonni, line 21: I'm not too sure what a good way of handling these query keys would be. Using the endpoint seems like a good way to ensure that it is unique, and reflects the right endpoint, but it also seems like repetition. Maybe there is a smarter way to handle this but haven't thought of an obvious one, and this seems good enough (for now at least). Open to ideas though.

arsforza: just as a thought for the future, will we have the need to move the ApiEndpoints enum outside of the ApiService?

arsforza (approved):

interfaces/Portalicious/src/app/components/health-widget/health-widget.component.ts
aberonni: Yeah, we potentially can. I prefer to keep the number of files to a minimum, because it makes things simpler for me in my mind - but if I'm the minority then I'm easily persuaded to move this to a separate file.

arsforza (approved):

PR #5534, opened 2024-07-04
chore: refactor api.service

aberonni:
Doing this to make it easier to carry over to Portalicious

- avoid repetition
- remove deprecated `toPromise`

### DIFF
--- a/interfaces/Portal/src/app/services/api.service.ts
+++ b/interfaces/Portal/src/app/services/api.service.ts
@@ -6,7 +6,7 @@ import {
   HttpStatusCode,
 } from '@angular/common/http';
 import { Injectable } from '@angular/core';
-import { Observable, of } from 'rxjs';
+import { lastValueFrom, Observable, of } from 'rxjs';
 import { catchError, tap } from 'rxjs/operators';
 import { USER_KEY } from '../auth/auth.service';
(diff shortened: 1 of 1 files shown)


interfaces/Portal/src/app/services/api.service.ts
```diff
@@ -38,215 +38,128 @@ export class ApiService {
...
+      throw error;
+    }
+  }
+
   get(
     endpoint: string,
     path: string,
     anonymous = false,
     responseAsBlob = false,
     params = null,
```
elwinschmitz, line 101: To be complete, we could add: `params: HttpParams = null,` (like in the other methods)

interfaces/Portal/src/app/services/api.service.ts
```diff
@@ -38,215 +38,128 @@ export class ApiService {
...
+    });
   }
 
   delete(
     endpoint: string,
     path: string,
-    body?: object,
-    params: HttpParams = null,
+    body?: unknown,
+    params?: HttpParams,
```
elwinschmitz, line 154: The other method-methods have the format: `params: HttpParams = null,`, would this be better for all?

aberonni:
@elwinschmitz I only created this PR to make sure I didn't break anything. Not really relevant for review because I'll probably rework those methods further.

PR #5542, opened 2024-07-05
portalicious: add localize to Portalicious

aberonni:
[AB#28828](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28828)

## Describe your changes

A lot of this is based on: https://github.com/angular/angular/issues/46851#issuecomment-1925507830

There is also a version that uses JSON files: #5543 

### DIFF
--- a/interfaces/Portalicious/src/app/utils/locale.ts
+++ b/interfaces/Portalicious/src/app/utils/locale.ts
@@ -1,0 +1,48 @@
+import { registerLocaleData } from '@angular/common';
+import { loadTranslations } from '@angular/localize';
+import { xliffToJson } from '~/utils/xliff-to-json';
+
+export const localStorageLocaleKey = 'preferredLanguage';
+
+export function getStoredLanguage(): string {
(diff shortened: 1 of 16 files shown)


interfaces/Portalicious/src/app/utils/locale.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  if (locale === 'en') {
+    // Default behavior, no changes required
+    return;
+  }
+
+  // Fetch XLIFF translation file and transform to JSON format (JSON translations can be used directly)
+  const json = await fetch('/assets/messages.' + locale + '.xlf')
+    .then((r) => r.text())
+    .then((t) => xliffToJson(t));
```
aberonni, line 22: Would be interesting to see if we can just use JSON files directly and avoid the hassle of this conversion.

aberonni: https://github.com/global-121/121-platform/pull/5543

PR #5545, opened 2024-07-08
fix: Cleanup package.json for k6-tests

elwinschmitz:
We don't need properties that will get out-of-sync.
Also, no need to mix licenses within the repository.

There will be no real 'standalone/distributable' package in the `k6`-folder, so the `package.json`-file is only there to manage dependencies etc.


Piotrk39 (approved):
Makes sense. Looks good to me!

PR #5568, opened 2024-07-11
portalicious: add localize to Portalicious

aberonni:
[AB#28828](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28828)


interfaces/Portalicious/src/app/utils/locale.ts
```diff
@@ -0,0 +1,75 @@
+import { isDevMode } from '@angular/core';
+import { environment } from '~environment';
+
+export const localStorageLocaleKey = 'preferredLanguage';
+
+export enum Locale {
+  en = 'en-US',
```
aberonni, line 7: - use GB
- use shorthand version in URL

interfaces/Portalicious/eslint.config.js
```diff
@@ -76,6 +76,13 @@ module.exports = tseslint.config(
       eslintPluginPrettierRecommended,
     ],
     rules: {
+      '@angular-eslint/template/i18n': [
+        'error',
+        {
+          checkId: false,
```
aberonni, line 82: Optionally we could enforce the usage of ids, rather than using auto-generated ones. This would also play nicely with the defaults of [daniel-sc/ng-extract-i18n-merge](https://github.com/daniel-sc/ng-extract-i18n-merge). However, it also makes our source files a bit more verbose.

I am in favour of doing less so I'd prefer auto-generated ids, but I don't have a strong preference and I'm easily persuaded.

interfaces/Portalicious/src/app/components/page-layout/components/footer/footer.component.html
```diff
@@ -3,33 +3,46 @@
...
           <a
+            i18n
             href="https://510.global"
             target="_blank"
             >510.global</a
           >
         </li>
         <li class="mb-2 [&:not(:hover)]:underline">
           <a
+            i18n
```
elwinschmitz, line 45: These three are basically only URLs, so they won't change between translations... ;)

interfaces/Portalicious/src/app/components/page-layout/components/footer/footer.component.html
```diff
@@ -64,8 +85,14 @@ <h3 class="mb-4">Helpful links</h3>
...
+        i18n
+        class="mb-4"
+      >
+        Questions or feedback? Contact us
+      </h3>
       <a
+        i18n
         href="mailto:support@510.global"
         class="p-button p-button-rounded p-button-outlined p-button-secondary [&:not(:hover)]:text-white"
         >support&#64;510.global</a
```
elwinschmitz, line 98: If we're putting the email-address in plain-text in the `href`, then there is no point in obfuscating it here, I'd think.

interfaces/Portalicious/staticwebapp.config.json
```diff
@@ -2,39 +2,25 @@
...
       "headers": {
         "Cache-Control": "immutable, no-transform, max-age=31536000"
       }
     },
     {
-      "route": "/*.css",
-      "headers": {
-        "Cache-Control": "immutable, no-transform, max-age=31536000"
-      }
+      "route": "/index.html",
```
elwinschmitz, line 11: I think by using: `"route": "/*"`, we could 'auto upgrade/redirect' all existing URLs, hopefully?

interfaces/Portalicious/src/app/utils/locale.ts
```diff
@@ -0,0 +1,85 @@
...
+  en = 'en',
+  nl = 'nl',
+}
+
+export function getLocaleLabel(locale: Locale): string {
+  switch (locale) {
+    case Locale.en:
+      return $localize`English`;
+    case Locale.nl:
+      return $localize`Dutch`;
```
elwinschmitz, line 15:
```suggestion
      return 'English';
    case Locale.nl:
      return 'Nederlands';
```

I think languages should have their own 'native' name, to aid in 'discoverability'/scanability. (And also, because we currently have them that way.)

elwinschmitz (changes requested):

interfaces/Portalicious/src/environments/environment.ts
```diff
@@ -8,7 +8,7 @@ export const environment = {
   // Configuration/Feature-switches:
   defaultLocale: 'en',
   envName: '', // To highlight the environment used
-  locales: 'en,ar,fr,nl,es', // Comma-separated string of enabled locales (i.e: 'en,es,nl_BE'). Each should be available in: `./src/app/services/language.service.ts`
+  locales: 'en,nl', // Comma-separated string of enabled locales (i.e: 'en,es,nl_BE'). Each should be available in: `./src/locale`
```
elwinschmitz, line 11:
```suggestion
  locales: 'en,nl', // Comma-separated string of enabled locales. Each should be available in: `./src/locale`
```

To prevent some confusion about the 'format', i.e. useing `_` or `-` as a separator, etc..

interfaces/Portalicious/staticwebapp.config.json
elwinschmitz: I think we can test this locally by using: https://github.com/Azure/static-web-apps-cli

I wonder if its worth it to use that as our "local dev server" and/or "production-mode test-tool"... As it will also check/use/fake the configurations around headers, cache-settings, etc.. I think.

interfaces/Portalicious/.env.example
```diff
@@ -12,7 +12,7 @@ NG_ENV_NAME=
 
 # Optional: Set the available languages/locales (comma-separated list, i.e: "en,nl,ar")
 # Make sure to add each language-code + a name in `./src/app/services/language.service.ts`
```
elwinschmitz, line 14:
```suggestion
# Make sure to add each language-code + a name in `./src/locale`
```

interfaces/Portalicious/eslint.config.js
elwinschmitz: I like that using "ID you have to name yourself" allows to use "names that make sense"... but the downside is "you _have_ to name them yourself". :D

I wonder if these auto-generated IDs will give us noisy/ugly diffs in day-to-day work... Then I think we have a good argument to "do it better by hand", maybe.

interfaces/Portalicious/staticwebapp.config.json
elwinschmitz: Haha, I didn't see your other PR yet about the SWA-CLI :)

But maybe, to prevent yet another fix-rebase-dance; lets combine #5585 into this branch and to a final rebase. ;)

interfaces/Portalicious/eslint.config.js
elwinschmitz: A "realistic" diff would be something like:
https://github.com/global-121/121-platform/pull/5568/commits/33f5c6c1b023e07c5f06b1db63b2067d72d6e7d3#diff-1e9eba834e433eeec035982534cf84aa6206b45805b53ff97ddff026a82751a1

So adding/removing some bits... It doesn't look _that_ bad.. ;)

services/.env.example
```diff
@@ -289,6 +289,5 @@ MOCK_COMMERCIAL_BANK_ETHIOPIA=TRUE
 # ---------------------------------------------------------------------
 
 # Playwright configuration
-BASE_URL=http://localhost:8888
```
elwinschmitz, line 292: I think the ENV-variable should stay in this file, and maybe further explained with a comment above it (explaining/noting that its optional).

So:

```suggestion
# (Optional) URL of the interface/Portal to use in E2E-tests
BASE_URL=
```

e2e/playwright.config.ts
```diff
@@ -46,7 +46,7 @@ export default defineConfig({
   outputDir: './test-results',
   timeout: 60000,
   use: {
-    baseURL: process.env.BASE_URL,
+    baseURL: process.env.BASE_URL ?? 'http://localhost:8888',
```
elwinschmitz, line 49: Part of the goal of setting the defaults in the `.env.example` is to not having to repeat it everywhere in the code..

e2e/portalicious/tests/Navigation/ChangeLanguage.spec.ts
```diff
@@ -0,0 +1,15 @@
+import HomePage from '@121-e2e/portalicious/pages/HomePage';
+import { test } from '@playwright/test';
+
+test('Change Language', async ({ page }) => {
+  await page.goto('/en/');
+
+  const homePage = new HomePage(page);
+  await page.waitForURL((url) => url.pathname.startsWith('/en/'));
+
+  await homePage.changeLanguage('Dutch');
```
elwinschmitz, line 10:
```suggestion
  await homePage.changeLanguage('Nederlands');
```

interfaces/Portalicious/src/app/pages/all-projects/all-projects.component.ts
```diff
@@ -6,7 +7,7 @@ import { PageLayoutComponent } from '~/components/page-layout/page-layout.compon
 @Component({
   selector: 'app-all-projects',
   standalone: true,
-  imports: [PageLayoutComponent, HealthWidgetComponent, RouterLink],
+  imports: [PageLayoutComponent, HealthWidgetComponent, RouterLink, DatePipe],
```
elwinschmitz, line 10:
```suggestion
  imports: [PageLayoutComponent, HealthWidgetComponent, RouterLink],
```

elwinschmitz (commented):
🚧

elwinschmitz (approved):

PR #5572, opened 2024-07-12
Visa error handling

RubenGeo:
## Describe your changes

### DIFF
--- a/services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
+++ b/services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
@@ -75,16 +75,12 @@ export class TransactionJobProcessorsService {
     }
 
     // Check if all required properties are present. If not, create a failed transaction and throw an error.
-    for (const key in input) {
-      if (key === 'addressHouseNumberAddition') continue; // Skip non-required property
+    for (const [name, value] of Object.entries(input)) {
+      if (name === 'addressHouseNumberAddition') continue; // Skip non-required property
 
       // Define "empty" based on your needs. Here, we check for null, undefined, or an empty string.
-      if (
-        input[key] === null ||
-        input[key] === undefined ||
-        input[key] === ''
-      ) {
-        const errorText = `Property ${key} is undefined`;
+      if (value === null || value === undefined || value === '') {
+        const errorText = `Property ${name} is undefined`;
         await this.processTransactionResult({
           jobInput: input,
           calculatedTranserAmount: input.transactionAmount, // TODO: STORE THE CALCULATED AMOUNT HERE AS THIS IS USED ON RETRY AND SHOWN IN PORTAL
--- a/services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
+++ b/services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
@@ -26,6 +26,17 @@ import { IntersolveVisaTransactionJobDto } from '@121-service/src/transaction-qu
 import { getScopedRepositoryProviderName } from '@121-service/src/utils/scope/createScopedRepositoryProvider.helper';
 import { Inject, Injectable } from '@nestjs/common';
 
+interface ProcessTransactionResultInput {
+  jobInput: IntersolveVisaTransactionJobDto;
+  calculatedTranserAmount: number;
+  financialServiceProviderId: number;
+  registration: RegistrationEntity;
+  oldRegistration: RegistrationEntity;
+  isRetry: boolean;
+  status: StatusEnum;
+  errorText?: string;
+}
+
 @Injectable()
 export class TransactionJobProcessorsService {
   public constructor(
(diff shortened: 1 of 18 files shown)


services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
```diff
@@ -74,17 +85,17 @@ export class TransactionJobProcessorsService {
...
-          userId: input.userId,
+          registration,
+          oldRegistration,
+          isRetry: input.isRetry,
           status: StatusEnum.error,
-          errorMessage: errorText,
+          errorText,
         });
-        throw new Error(errorText);
+        return;
```
RubenGeo, line 98: Return instead of error, to not pollute our logs

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
```diff
@@ -134,24 +145,20 @@ export class TransactionJobProcessorsService {
...
         status: StatusEnum.error,
-        errorMessage: `An error occured: ${error}`,
+        errorText: error?.message, // TODO: THIS IS A GENEARAL stack trace catch and i think the error handling should be more specific and tested
       });
-      await this.updatePaymentCountAndStatusInRegistration(
-        registration,
-        input.programId,
-      );
-      throw new Error(error);
+      return;
```
RubenGeo, line 158: Return instead of throw error, to not pollute our logs

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
```diff
@@ -134,24 +145,20 @@ export class TransactionJobProcessorsService {
           )?.value as string, // This must be a string. If it is not, the intersolve API will return an error (maybe).
         });
     } catch (error) {
-      await this.createTransaction({
-        amount: input.transactionAmount,
-        registration: registration,
+      await this.processTransactionResult({
```
RubenGeo, line 148: If a transaction fails you also need to update completed last transaction etc

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
```diff
@@ -134,24 +145,20 @@ export class TransactionJobProcessorsService {
...
+        jobInput: input,
+        calculatedTranserAmount: input.transactionAmount, // TODO: STORE THE CALCULATED AMOUNT HERE AS THIS IS USED ON RETRY AND SHOWN IN PORTAL
```
RubenGeo, line 150: I think we need to adjust the flow to get this earlier

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw: What does "processTransactionResult" conceptually mean?

jobInput: I do not like passing in "big object parameters) into functions and then having functions internally figure out what they need. I know this happens a lot now throughout our code, for example with PaPaymentDetails "blobs" which are inobvious about what they conceptually are/contain, and get passed around throughout a service. IMO every function should have a "clean" interface, only receiving what it needs and passing back what the caller needs.

Yes, good point about storing the transactionAmount in the failed transaction. However, at this point it is not a calculated amount yet, since that happens in the IntersolveVisa service. Unless you mean that it is calculated already with the multiplier. If that is the case, check where that happens. If that is when creating the job, then IMO we should not be talking about "calculated" amount here at all.
So what is "calculatedTrans(f)Amount" conceptually?

Yes, you have a point about not rethrowing the error. There could be a misconfiguration in the program, but that is not "our problem".

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw: Very good point.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw: See above comment.

If we are not talking about the multiplier here (which I believe we are not), but about the calculation that happens related to balance and spent this month, then I think we do not need to adjust the flow. It will just recalculate it on retry. Which is fine. Or perhaps even better.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw: Indeed. Creating a failed transaction is "normal behavior" and should not throw an error.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
```diff
@@ -134,24 +145,20 @@ export class TransactionJobProcessorsService {
...
-      await this.updatePaymentCountAndStatusInRegistration(
-        registration,
-        input.programId,
-      );
-      throw new Error(error);
+      return;
     }
 
-    if (intersolveVisaDoTransferOrIssueCardReturnDto.cardCreated) {
+    if (intersolveVisaDoTransferOrIssueCardReturnDto?.cardCreated) {
```
diderikvw, line 161: Why is the question mark necessary?

I do not like these kind of statements, as this is where bugs hide.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
RubenGeo: Renamed in this [commit](https://github.com/global-121/121-platform/pull/5617/commits/e21c38f551a6a900453abc914b3e9fc6b024b641)

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
RubenGeo: It's was not needed removed in another branch

PR #5582, opened 2024-07-16
feat: Add ApplicationInsights logging in LogService

elwinschmitz:
- optional debug plugin for local development
- Example implementation of custom-events

See: [AB#28898](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28898)


interfaces/Portalicious/src/app/services/log.service.ts
```diff
@@ -0,0 +1,113 @@
+import { Injectable } from '@angular/core';
+import { DebugPlugin } from '@microsoft/applicationinsights-debugplugin-js';
+import { ApplicationInsights } from '@microsoft/applicationinsights-web';
+import { environment } from 'src/environments/environment';
+
+// All properties sorted alphabetically.
+export enum LogEvent {
```
aberonni, line 7: Maybe we wanna add this? https://github.com/infctr/eslint-plugin-typescript-sort-keys/blob/master/docs/rules/string-enum.md

interfaces/Portalicious/src/app/services/log.service.ts
```diff
@@ -0,0 +1,113 @@
...
+  userLogout = 'user/logout',
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LogService {
+  private appInsights: ApplicationInsights;
+  private appInsightsEnabled: boolean;
```
aberonni, line 16: Maybe more indicative of what this represents:

```suggestion
  private appInsightsInitialized: boolean;
```

aberonni (approved):
Some minor things but feel free to merge as-is

interfaces/Portalicious/src/app/services/log.service.ts
elwinschmitz: Ah, yes. If we limit it to TS-only, then there's not really any risk and things can be autofixed...

The only issues in the past was with "hand-editted JSON-files", for those the sorting made it "less understandable"..

interfaces/Portalicious/src/app/services/log.service.ts
elwinschmitz: Ah, yes.. This was a remnant of a time where there where multiple "logging services" being initiated/enabled to use in this service.

But in this case its mainly to prevent "re-initialization", so that name would be better.

interfaces/Portalicious/src/app/services/log.service.ts
```diff
@@ -0,0 +1,113 @@
...
+export enum LogEvent {
+  userLogout = 'user/logout',
+}
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LogService {
+  private appInsights: ApplicationInsights;
+  private appInsightsInitiated: boolean;
```
aberonni, line 16:
```suggestion
  private appInsightsInitialized: boolean;
```

Sorry to be that guy but initiated means something very different :D

aberonni (changes requested):

interfaces/Portalicious/src/app/services/log.service.ts
elwinschmitz: Ha. _Thanks_ Copilot-autocomplete. 😶

interfaces/Portalicious/src/app/services/log.service.ts
elwinschmitz: @aberonni I've found the perfectly named plugin for that:
https://eslint-plugin-perfectionist.azat.io/rules/sort-enums

I'll add in a separate PR. (together with some other ESLint-things)

aberonni (approved):

PR #5585, opened 2024-07-17
portalicious: add e2e tests for i18n and navigation

aberonni:
[AB#29168](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29168)

Besides what's mentioned in the title, I'm also
- replacing `http-servier` with the `swa-cli`. This actually worked surprisingly well out of the box, and made me upset that I didn't discover this when I was trying to get the redirects in the swa config right.
- Removed the `BASE_URL` environment variable from the services .env.example. I don't think it should've been there to begin with.

- [ ] ~I do not need any deviation from our PR guidelines~ this is ready for review but should not be merged before #5568  has been merged


services/.env.example
```diff
@@ -289,6 +289,5 @@ MOCK_COMMERCIAL_BANK_ETHIOPIA=TRUE
 # ---------------------------------------------------------------------
 
 # Playwright configuration
-BASE_URL=http://localhost:8888
```
elwinschmitz, line 292: I think the ENV-variable should stay in this file, and maybe further explained with a comment above it (explaining/noting that its optional).

So:

```suggestion
# (Optional) URL of the interface/Portal to use in E2E-tests
BASE_URL=
```

aberonni: Merging for now. We can resume the convo in the other PR if you deem it necessary.

PR #5589, opened 2024-07-19
portalicious: revert to node v20

aberonni:
## Describe your changes

Reverting to v20 because v22 is unstable and giving us issues:

image

I also removed the package-lock and regenerated it to be on the safe side.


arsforza (approved):

elwinschmitz:
@aberonni Also, maybe an even more pressing argument for this, is that we use the `121-service`(and its dependencies) as a dependency; And that is on v20.

So to prevent any weird issues there, its better to stick to v20 and upgrade all at the same time.

PR #5606, opened 2024-07-25
portalicious: add change password

arsforza:
AB#29275

## Describe your changes


interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -1,12 +1,99 @@
...
+    };
+  }
+
+  changePasswordFormSubmitted = signal(false);
+
+  onChangePassword() {
+    this.changePasswordFormSubmitted.set(true);
+
+    if (this.changePasswordForm.invalid) {
+      console.log('🚀 ~ UserSettingsComponent ~ onChangePassword ~ ostiasa:');
```
elwinschmitz, line 74: `ostiasa`?

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -1,12 +1,99 @@
...
+
+      if (
+        matchingControl?.errors &&
+        !matchingControl.errors.confirmedValidator
+      ) {
+        return null;
+      }
+
+      if (control?.value !== matchingControl?.value) {
+        const error = { confirmedValidator: 'Passwords do not match.' };
```
elwinschmitz, line 58:
```suggestion
        const error = {
          confirmedValidator: $lokalise`Passwords do not match.`
        };
```

Or could we let this string 'come from' the template-file?
So with an additional attribute on the password-field or something?

elwinschmitz (changes requested):

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.html
```diff
@@ -1,3 +1,131 @@
...
+                </div>
+              </ng-template>
+            </p-messages>
+          } @else {
+            <p-messages [severity]="'success'">
+              <ng-template pTemplate>
+                <i class="pi pi-exclamation-triangle"></i>
+                <div
+                  i18n
+                  class="ml-2"
```
elwinschmitz, line 120: We should not use "margin-**left**", as we want to have the interface compatible with RTL-languages..

We should use "margin-inline-**start**", so `ms-2`.

```suggestion
                  class="ms-2"
```

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.html
```diff
@@ -1,3 +1,131 @@
...
+          <p-progressSpinner
+            aria-label="Loading"
+            i18n-aria-label="@@generic-loading"
+          />
+        } @else {
+          @if (changePasswordError()) {
+            <p-messages [severity]="'error'">
+              <ng-template pTemplate>
+                <i class="pi pi-check-circle"></i>
+                <div class="ml-2">
```
elwinschmitz, line 109:
```suggestion
                <div class="ms-2">
```

interfaces/Portalicious/src/app/services/api.service.ts
```diff
@@ -39,4 +40,24 @@ export class ApiService {
...
+    password,
+    newPassword,
+  }: {
+    username: string;
+    password: string;
+    newPassword: string;
+  }) {
+    return this.httpWrapperService.perform121ServiceRequest({
+      method: 'POST',
+      endpoint: 'users/password',
```
elwinschmitz, line 55:
```suggestion
      endpoint: ApiEndpoints.usersChangePassword,
```

elwinschmitz (changes requested):
Just some small "typo's" :P

elwinschmitz (approved):

interfaces/Portalicious/src/app/app.routes.ts
```diff
@@ -42,6 +42,7 @@ export const routes: Routes = [
   },
   {
     path: AppRoutes.userSettings,
+    title: $localize`:Browser-tab-title@@page-title-user-settings:Change password`,
```
aberonni, line 45: I think this might be the only page with a title, and therefore also why it's the only page that persists the title (ie. the bug we saw this morning in the call).

interfaces/Portalicious/src/app/components/page-layout/page-layout.component.html
```diff
@@ -4,7 +4,7 @@
     @if (programId()) {
       <app-program-header [programId]="programId()!" />
     }
-    <div class="px-6 py-4 lg:px-24">
+    <div class="mx-auto max-w-screen-xl px-6 py-4">
```
aberonni, line 7: This is the change I was discussing this morning in the call (ie. the website not being full screen anymore). I will revert this in a separate PR.

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -1,12 +1,182 @@
...
+  ],
   templateUrl: './user-settings.component.html',
   styles: ``,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserSettingsComponent {}
+export class UserSettingsComponent {
+  private authService = inject(AuthService);
+
+  public changePasswordError = signal<null | string>(null);
```
aberonni, line 44: I don't think there is a need for a separate variable. I believe this gets stored in the mutation already, for example: https://github.com/global-121/121-platform/pull/5575/files#diff-0b0127c431f43bdcc94487f9bcc4a3cb61332fe0e1421c240d3e73e59a599332R64

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -1,12 +1,182 @@
...
+  private currentPasswordFormControl = new FormControl('', Validators.required);
+
+  private newPasswordFormControl = new FormControl('', [
+    /* eslint-disable-next-line @typescript-eslint/unbound-method */
+    Validators.required,
+
+    Validators.minLength(8),
+  ]);
+  /* eslint-disable-next-line @typescript-eslint/unbound-method */
+  private confirmPasswordFormControl = new FormControl('', Validators.required);
```
aberonni, line 129: I am worried that it takes this much custom code to get this (relatively simple) validator working. Especially the logic around deleting properties from objects etc. I have to assume there is a simpler way to do this - perhaps on a form level rather than a field level - because I can't believe that such a common scenario requires such jumping through hoops.

interfaces/Portalicious/src/app/pages/user-settings/user-settings.component.ts
```diff
@@ -1,12 +1,182 @@
...
+      currentPassword: this.currentPasswordFormControl,
+      newPassword: this.newPasswordFormControl,
+      confirmPassword: this.confirmPasswordFormControl,
+    },
+    {
+      validators: [this.newPasswordValidator, this.confirmPasswordValidator],
+    },
+  );
+
+  changePasswordFormSubmitted = signal(false);
```
aberonni, line 142: Similar to above, I think we can get this info from the mutation without relying on an extra variable

aberonni (commented):
I'll open a PR for the css change, but maybe we can look at the other things together once I've done some digging of my own.

PR #5616, opened 2024-07-30
try to fix the flaky test

XiaoyuSun-hub:
## Describe your changes
The messages on the the message history dialog sometimes are different when open it directly after payment, so add a step to refresh the page before open message history.
 If the message still not show, open the message history again in the verification function.

### DIFF
--- a/e2e/pages/Payments/PaymentsPage.ts
+++ b/e2e/pages/Payments/PaymentsPage.ts
@@ -296,22 +296,18 @@ class PaymentsPage {
     messageNotification,
     messageContext,
     messageType,
-    locatorNumber = 0,
   }: {
     messageNotification: string;
     messageContext: string;
     messageType: string;
-    locatorNumber?: number;
   }) {
-    const paymentNotificationLocator = this.page
-      .locator(`:text("${messageContext} (${messageType})")`)
-      .nth(locatorNumber);
+    const paymentNotificationLocator = this.page.locator(
+      `:text("${messageContext} (${messageType})")`,
+    );
     const messageNotificationLocator = this.page.locator(
       `:text("${messageNotification}")`,
     );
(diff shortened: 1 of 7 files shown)


Piotrk39 (commented):
From what I see when the test runs in CI it is still flaky

e2e/pages/Payments/PaymentsPage.ts
```diff
@@ -292,5 +291,44 @@ class PaymentsPage {
...
+    messageType: string;
+    locatorNumber?: number;
+  }) {
+    const paymentNotificationLocator = this.page
+      .locator(`:text("${messageContext} (${messageType})")`)
+      .nth(locatorNumber);
+    const messageNotificationLocator = this.page.locator(
+      `:text("${messageNotification}")`,
+    );
+    // await this.page.waitForLoadState('networkidle');
```
Piotrk39, line 312: In general the corrections you have provided look good to me and the test works without flakiness. 

Good job!

Two things here only:
1. Could you remove the commented line if it is not necessary?
2. Can locator be more explicitly defined or the .nth is the only solution? If it is then there is no issue :)

Piotrk39 (changes requested):

e2e/pages/Payments/PaymentsPage.ts
XiaoyuSun-hub: 1. I have removed it
2. No, because it is also be used to verify "Activity overview" on "Registration details". Then there will be multiple items with same text and also same testid.

Piotrk39 (approved):

e2e/pages/Payments/PaymentsPage.ts
Piotrk39: Approved!

PR #5638, opened 2024-08-01
Remove "before all" & change response query for random order assertions

Piotrk39:
## Describe your changes

Now the tests will be able to run in random order. Things that has been changed:
1. Remove "Before all" and add "Before each" so the reset script can now reset before each test
2. Adapt assertions to match the newly returned response


elwinschmitz:
Would it be possible to set `randomize: true`, here:
https://github.com/global-121/121-platform/blob/main/services/121-service/jest.e2e.config.js#L17

That would really check/prove that the tests are independent now...

Piotrk39:
> @Piotrk39 Would it be possible to set `randomize: true`, here: https://github.com/global-121/121-platform/blob/main/services/121-service/jest.e2e.config.js#L17
> 
> That would really check/prove that the tests are independent now...

@elwinschmitz Let me test how this and the other tests behave.

Piotrk39:
@elwinschmitz during the fully randomised test 6 more tests have failed. I'll fix those as well.

Vladimir-Petkov (approved):

PR #5641, opened 2024-08-01
Api test duplicate operation reference AB#29524

RubenGeo:
## Describe your changes

### DIFF
--- a/services/mock-service/src/fsp-integration/intersolve-visa/intersolve-visa.mock.service.ts
+++ b/services/mock-service/src/fsp-integration/intersolve-visa/intersolve-visa.mock.service.ts
@@ -588,7 +588,7 @@ export class IntersolveVisaMockService {
   ): IntersolveVisaMockResponseDto {
     if (amount === MAGIC_FAIL_OPERATION_REFERENCE_AMOUNT) {
       return {
-        status: HttpStatus.NOT_FOUND,
+        status: HttpStatus.BAD_REQUEST,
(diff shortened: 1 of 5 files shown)


services/mock-service/src/fsp-integration/intersolve-visa/intersolve-visa.mock.service.ts
```diff
@@ -578,7 +582,29 @@ export class IntersolveVisaMockService {
...
 
-  public transfer(fromToken: string): IntersolveVisaMockResponseDto {
+  public transfer(
+    fromToken: string,
+    amount: number,
+  ): IntersolveVisaMockResponseDto {
+    if (amount === MAGIC_FAIL_OPERATION_REFERENCE_AMOUNT) {
+      return {
+        status: HttpStatus.NOT_FOUND,
+        statusText: 'Not Found',
```
PeterSmallenbroek, line 592: When I tested this, I received a 400 Bad request. Is there a specific reason we're not mirroring this behaviour?
If that's the case you can resolve this comment and merge it in.

PeterSmallenbroek (changes requested):

services/mock-service/src/fsp-integration/intersolve-visa/intersolve-visa.mock.service.ts
RubenGeo: No reason I'll change it :)

PR #5658, opened 2024-08-07
portalicious: dynamic form component

aberonni:
Hello, I did a thing and I'm still not sure if it was a good idea or not.

Basically I was about to start work on [AB#29544](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29544) and I realised that I was gonna have to copy/paste a lot of code. That's never nice. So I reflected on how to avoid that. And came up with this.

But!

Is this the right approach?

I am not sure.

End of story. Opinions please 😄 

### DIFF
--- a/interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
+++ b/interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
@@ -1,0 +1,52 @@
+<form
+  class="block w-full"
+  [formGroup]="formGroup()"
+  (ngSubmit)="onFormSubmit()"
+>
+  <div class="flex flex-col gap-2">
+    @for (field of fields(); track field.controlName) {
+      <!-- eslint-disable-next-line @angular-eslint/template/label-has-associated-control -->
+      <label>
+        <ng-container>{{ field.label }}</ng-container>
+        @switch (field.type) {
+          @case ('password') {
+            <p-password
+              [formControlName]="field.controlName"
+              [autocomplete]="field.autocomplete"
+              [toggleMask]="true"
+              [feedback]="false"
+              pAutoFocus
+              [autofocus]="field.autoFocus"
+              styleClass="mt-2 w-full [&_input]:w-full"
+            />
+          }
+          @default {
+            <!-- eslint-disable @angular-eslint/template/no-autofocus -->
+            <input
+              pInputText
+              [type]="field.type"
+              [formControlName]="field.controlName"
+              [autocomplete]="field.autocomplete"
+              [placeholder]="field.placeholder"
+              pAutoFocus
+              [autofocus]="field.autoFocus"
+              class="mt-2 block w-full"
+            />
+          }
+        }
+      </label>
+      <app-form-error
+        [visible]="formGroup().controls[field.controlName].touched"
+        [error]="field.validationMessage?.(formGroup())"
+      />
+    }
+    <p-button
+      [label]="submitButtonText()"
+      type="submit"
+      rounded
+      class="mt-8"
+      [loading]="isSubmitting()"
+    />
+    <app-form-error [error]="formError()" />
+  </div>
+</form>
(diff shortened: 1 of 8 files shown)


interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
```diff
@@ -0,0 +1,50 @@
...
+        <ng-container>{{ field.label }}</ng-container>
+        @switch (field.type) {
+          @case ('password') {
+            <p-password
+              [formControlName]="field.controlName"
+              [autocomplete]="field.autocomplete"
+              [toggleMask]="true"
+              [feedback]="false"
+              styleClass="mt-2 w-full [&_input]:w-full"
+            />
```
arsforza, line 19: `<p-password
  [formControlName]="field.controlName"
  [autocomplete]="field.autocomplete"
  [toggleMask]="true"
  [feedback]="false"
  styleClass="mt-2 w-full [&_input]:w-full"
/>`

This could become its own component?
I am saying this because these kind of dynamic components can become monsters.
Also, let's try to have the least amount of exceptions (special fields that need an element other than the default)

interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
```diff
@@ -0,0 +1,50 @@
...
+          }
+          @default {
+            <!-- eslint-disable @angular-eslint/template/no-autofocus -->
+            <input
```
arsforza, line 23: Should we use the standard input or is it better to use PrimeNG components as much as possible?

interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
aberonni: > This could become its own component?

In this case, I don't see the advantage of doing that.

In general, my preference is to abstract only when reusability is necessary. A good example of a "rule" is the [rule of three](https://en.wikipedia.org/wiki/Abstraction_principle_(computer_programming)#:~:text=A%20rule%20of%20thumb%20governing,needs%20to%20be%20abstracted%20out.). According to this rule, we wouldn't abstract because we are only using this in one place.

But, rules can have exceptions. Let's analyze this particular example. 

I assume, with the extrapolation that you suggest, the code would become something like this:

```html
            <app-dynamic-form-password
              [formControlName]="field.controlName"
              [autocomplete]="field.autocomplete"
              styleClass="mt-2 w-full [&_input]:w-full"
            />
```

We would have a lot more code (two new files, a new class/component definition, together with re-defining input props etc.) but I don't necessarily see the advantage. 

> I am saying this because these kind of dynamic components can become monsters.

I agree that dynamic components can become monsters. But I don't think moving this code elsewhere will improve it (if anything, the opposite). I think there might be other ingredients that will make it a monster with time.

> Also, let's try to have the least amount of exceptions (special fields that need an element other than the default)

I think this is unavoidable, at least with types. I imagine that we will at least have types like radio, select, textarea, date etc. at a certain point.

If we don't want that, then maybe we shouldn't have a dynamic component. That might be the right approach, but then I'm not sure how to avoid the original duplication that this PR is addressing 😬 (and maybe that's ok, and we just need to figure something else out).

interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
aberonni: I'm not sure I understand the question 😬  What would "using the standard input" mean?

Anyways, without understanding the question, 

> is it better to use PrimeNG components as much as possible

my answer to this would be yes.

interfaces/Portalicious/src/app/components/dynamic-form/dynamic-form.component.html
aberonni: Poking around the figma files and I see something like this:

image

And it would require a lot more edge cases and options to work with that layout. I am, once again, questioning everything 😄

aberonni:
@arsforza I'm closing this because I prefer what I did in #5672 but feel free to chime in if you still have thoughts/opinions on what was discussed above (or if perhaps you actually prefer this approach)

PR #5667, opened 2024-08-07
Chore add fullname to ocw to better reflect current reality on prod

RubenGeo:
## Describe your changes


e2e/tests/121-Portal/MakeNewPayment/OCW/SendPaymentInstructionsOCW.spec.ts
```diff
@@ -46,7 +46,7 @@ test.beforeEach(async ({ page }) => {
   );
 });
 
-test('[28448] OCW: Send payment instructions', async ({ page }) => {
+test.skip('[28448] OCW: Send payment instructions', async ({ page }) => {
```
aberonni, line 49: It would be best to add a reason why we skip a test whenever we skip. Ideally also a link to a task to work on re-enabling the test.

aberonni:
@PeterSmallenbroek to get this to pass I did a couple of things:
- I rebased on main after #5700 was merged
- I dropped your changes where you removed all the `null`s from the jsons
- I dropped your change to fix the query parameter (because otherwise the snapshots were significantly different)
- I updated the snapshots

Especially on the last point I am not sure that the changes are correct so someone else would need to check them.

PeterSmallenbroek (approved):

PR #5672, opened 2024-08-08
portalicious: another approach to forms

aberonni:
This is an alternative approach to what I had tried in #5658

I prefer this approach better now that I've fleshed it out. I think it's the right balance between reducing duplication, and flexibility.

More details in the comments.

### DIFF
--- a/interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
+++ b/interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
@@ -1,0 +1,39 @@
+<p-button
+  label="Create new project"
+  i18n-label
+  outlined
+  icon="pi pi-plus"
+  class="text-xs sm:text-base"
+  (click)="formVisible.set(true)"
+/>
+<app-form-sidebar
+  [(visible)]="formVisible"
+  [formGroup]="formGroup"
+  [mutation]="createProjectMutation"
+>
+  <h2 i18n>Create Project</h2>
+  <app-form-field-wrapper
+    label="Kobo Asset ID"
+    i18n-label
+    [errorMessage]="formFieldErrors()('assetId')"
+  >
+    <input
+      pInputText
+      type="text"
+      formControlName="assetId"
+      autocomplete="off"
+    />
+  </app-form-field-wrapper>
+  <app-form-field-wrapper
+    label="Kobo Token"
+    i18n-label
+    [errorMessage]="formFieldErrors()('token')"
+  >
+    <p-password
+      formControlName="token"
+      autocomplete="off"
+      [toggleMask]="true"
+      [feedback]="false"
+    />
+  </app-form-field-wrapper>
+</app-form-sidebar>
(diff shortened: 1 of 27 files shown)


interfaces/Portalicious/src/app/pages/all-projects/components/create-project-button/create-project-button.component.html
aberonni, line 1: I still need to fix the design for this page, but this is the best example to illustrate the different approach (because it is a new page).

Essentially, in the other PR, we were defining a JSON with a bunch of fields and passing that to a "dynamic form builder".

Instead, here, we are defining everything in the component itself, while reusing composable components.

interfaces/Portalicious/src/app/utils/form-validation.ts
```diff
@@ -0,0 +1,16 @@
+import { FormGroup } from '@angular/forms';
+
+export function generateFieldErrors<T extends FormGroup>(
```
aberonni, line 3: I'm not sure how confusing this makes the readability of the code in the components.

I added this mainly as a wrapper because I found myself checking `formGroup.controls[controlName].touched` everywhere an error on a field was shown. This way, that check isn't necessary anymore. But maybe it would be better/simpler to just be explicit and duplicate an `if (formGroup.controls[controlName].touched)` everywhere.

aberonni:
@arsforza this isn't ready to be merged because it still needs some tweaking here and there, but it could still use a review for sure - especially on the new form components.

interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
```diff
@@ -0,0 +1,39 @@
+<p-button
+  label="Create new project"
+  i18n-label
+  outlined
+  icon="pi pi-plus"
+  class="text-xs sm:text-base"
+  (click)="formVisible.set(true)"
+/>
+<app-form-sidebar
```
tijsziere, line 9: Can we change the order? Since the token is a password (or password like), it makes more sense to have that second. So first the AssedId and then the Token

interfaces/Portalicious/src/app/services/http-wrapper.service.ts
```diff
@@ -66,12 +78,60 @@ export class HttpWrapperService {
...
-          return of(new Error('Token expired'));
+          return of(new Error($localize`:@@error-token-expired:Token expired`));
         }
       }
 
-      return of(new Error('Not authorized'));
+      return of(new Error($localize`:@@error-not-authorized:Not authorized`));
     }
+
+    let errorMessage = get(error, 'message', undefined);
```
aberonni, line 88: I added a lot of stuff here for default error handling. I'm not sure it's relevant always but I guess we'll find out through time?

interfaces/Portalicious/src/app/utils/form-validation.ts
```diff
@@ -0,0 +1,28 @@
...
+      control: T['controls'][K],
+    ) => string | undefined;
+  },
+) {
+  return function (controlName: keyof T['controls'] & string) {
+    const control = formGroup.controls[controlName];
+    if (!control.touched) {
+      return undefined;
+    }
+    // @ts-expect-error I couldn't get TS to stop complaining but this is correct
```
aberonni, line 25: Trying to fix this.

arsforza (approved):

arsforza (approved):

PR #5381, opened 2024-06-04
feat: Move to multi-tenant SSO AB#27356

elwinschmitz:
[AB#27504](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/27504)
Retry of #5256

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [ ] I have added tests wherever relevant
- [x] I have made sure that all automated checks pass before requesting a review
- [ ] I do not need any deviation from our PR guidelines

PR #5693, opened 2024-08-14
portalicious: project -> teams -> table

aberonni:
[AB#29676](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29676)


interfaces/Portalicious/src/app/auth.guard.ts
```diff
@@ -18,3 +19,30 @@ export const authGuard: CanActivateFn = (_, state) => {
...
+  ) {
+    const authService = inject(AuthService);
+
+    const projectId = route.params.projectId as number;
+
+    if (isNaN(projectId)) {
+      throw new Error('projectId is not a number');
+    }
+
+    if (authService.hasPermission(projectId, permission)) {
```
elwinschmitz, line 37: There is no "needs multiple permissions"-functionality yet, I think?

interfaces/Portalicious/src/app/services/api.service.ts
```diff
@@ -1,15 +1,17 @@
...
 import { User } from '~/models/user.model';
 import { HttpWrapperService } from '~/services/http-wrapper.service';
 
 export enum ApiEndpoints {
   payments = 'payments',
   projects = 'programs',
   projectsMetrics = 'metrics/program-stats-summary',
+  users = 'users',
   usersChangePassword = 'users/password',
+  usersCurrent = 'users/current/',
```
elwinschmitz, line 14: Does it need that last `/`?

interfaces/Portalicious/src/app/services/auth.service.ts
```diff
@@ -153,4 +154,50 @@ export class AuthService {
...
+    // During development: Use this to simulate a user not having a certain permission
+    // user.permissions[programId] = user.permissions[programId].filter(
+    //   (p) => p !== Permission.FspDebitCardBLOCK,
+    // );
+
+    // TODO: Move this to a better place in the flow, so it doesn't have to be checked this often
+    // Check with Azure (again) when user has no permissions (yet)
+    if (!this.isAssignedToProgram(programId, user)) {
+      void this.refreshCurrentUser(); // Don't await, as it will block all permission-checks
+    }
```
elwinschmitz, line 195: 😭😭😭

I hope we can do without taking this skeleton-from-the-old-portal with us.. 🤞

interfaces/Portalicious/.prettierrc.yml
```diff
@@ -11,6 +11,7 @@ plugins:
   - '@prettier/plugin-xml'
 tailwindAttributes:
   - styleClass
+  - icon
```
elwinschmitz, line 14: Isn't it possible to have 1 central `tailwind.config.json` or something?

One that both Prettier AND the VSCode-extension can/should use?

interfaces/Portalicious/eslint.config.js
```diff
@@ -95,6 +95,12 @@ module.exports = tseslint.config(
             'ngSrc',
             'styleClass',
             'position',
+            'pSortableColumn',
+            'field',
+            'display',
+            'sortField',
+            'appendTo',
+            'iconPosition',
```
elwinschmitz, line 103: Isn't there a "PrimeNG-ESLint-plugin" somewhere? So we don't have to maintain this list here ourselves?

interfaces/Portalicious/src/app/auth.guard.ts
```diff
@@ -18,3 +19,30 @@ export const authGuard: CanActivateFn = (_, state) => {
...
+      return true;
+    }
+
+    const router = inject(Router);
+    return router.navigate(['/', AppRoutes.project, projectId], {
+      queryParams: {
+        permissionDenied: true,
```
elwinschmitz, line 44: Is this used somewhere already? Or is there a default-feature of Angular for this?

interfaces/Portalicious/src/app/auth.guard.ts
aberonni: Tbh, I don't know 😬

interfaces/Portalicious/src/app/services/api.service.ts
aberonni: Nope :D I'll get rid of it

interfaces/Portalicious/src/app/services/auth.service.ts
aberonni: Tbh I'm not entirely sure what is going on here. I blindly brought it over but we can decide not to unless we really have to.

interfaces/Portalicious/src/app/auth.guard.ts
aberonni: No. It's something that I forgot to implement 🙈 I will add a TODO message for now and we can decide whether to look at it now or revisit later.

interfaces/Portalicious/.prettierrc.yml
aberonni: Not that I'm aware of.

aberonni: Or better, we do have a central tailwind.config but that doesn't deal with these attributes.

interfaces/Portalicious/eslint.config.js
elwinschmitz: Theoretically we could import a list of attributes from somewhere via code (this is a JS-file with config-as-code after all...

But as that isn't really a commonly used, documented thing (yet), we can make due with this indeed.

---

Some interesting thing I've found:
https://github.com/angular-eslint/angular-eslint/issues/839#issuecomment-1161328555

So we can be more specific to ignore _some_ attributes on _some_ elements only...

interfaces/Portalicious/src/app/auth.guard.ts
elwinschmitz: Ah, yes..  The whole flow of "your session/token/cookie has expired and you need to re-login" needs to be looked at, also with SSO in mind, would deserve its own backlog-item, I think.

elwinschmitz: Oh, wait... there is also this error-pattern:

https://github.com/global-121/121-platform/pull/5381/files?file-filters%5B%5D=.html&file-filters%5B%5D=.ts&show-viewed-files=true#diff-87ffb49babf9b145819c6fdb333f5e849b84c120e52fab27ae5b7b7e430e2a89

Where we transfer 1 (out of 2/multiple)-error-coded between pages/reloads via local-storage...

interfaces/Portalicious/src/app/components/query-table/query-table.component.html
```diff
@@ -0,0 +1,173 @@
...
+    [rowsPerPageOptions]="[10, 20, 40, 80, 100]"
+    [showCurrentPageReport]="true"
+    [currentPageReportTemplate]="currentPageReportTemplate"
+    stateStorage="local"
+    [stateKey]="localStorageKey()"
+  >
+    <ng-template pTemplate="caption">
+      <div class="flex">
+        @if (globalFilterFields()) {
+          <p-iconField iconPosition="left">
```
elwinschmitz, line 26: 😭  Ow, PrimeNG...

interfaces/Portalicious/src/app/components/query-table/query-table.component.html
```diff
@@ -0,0 +1,173 @@
...
+          </p-iconField>
+        }
+        <p-button
+          label="Clear all filters"
+          i18n-label="@@form-clear-filters"
+          [outlined]="true"
+          icon="pi pi-filter-slash"
+          (onClick)="clearAllFilters()"
+          [disabled]="!isFiltered()"
+          class="ms-auto"
```
elwinschmitz, line 45: \o/

interfaces/Portalicious/src/app/components/query-table/query-table.component.html
```diff
@@ -0,0 +1,173 @@
...
+                <app-table-cell-date [value]="item[column.field]" />
+              }
+              @default {
+                <app-table-cell-text [value]="item[column.field]" />
+              }
+            }
+          </td>
+        }
+        @if (contextMenuItems()) {
+          <td class="w-0 py-0 pl-0">
```
elwinschmitz, line 108: `ps-0`?

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.html
```diff
@@ -3,5 +3,17 @@
...
+      <app-add-user-button />
+    }
+  </div>
+  <app-query-table
+    [columns]="columns()"
+    [query]="projectUsers"
+    localStorageKey="project-team-table"
+    [contextMenuItems]="contextMenuItems()"
+    (onUpdateContextMenuItem)="selectedUser.set($event)"
+    [globalFilterFields]="['username', 'roleString', 'scope']"
```
elwinschmitz, line 17: Is there a way to 'connect' this more?
So not using "just strings", but using actual references/enums/etc?

interfaces/Portalicious/src/app/services/api.service.ts
```diff
@@ -106,4 +117,19 @@ export class ApiService {
...
+    const users = await this.httpWrapperService.perform121ServiceRequest<
+      ProjectUser[]
+    >({
+      method: 'GET',
+      endpoint: `${ApiEndpoints.projects}/${projectId.toString()}/${ApiEndpoints.users}`,
+    });
+
+    return users.map((user) => ({
+      ...user,
+      roleString: user.roles.map((role) => role.label).join('; '),
```
elwinschmitz, line 131: `rolesLabels`? or `rolesLabel`?
Or even `allRoles` / `allRolesLabel`

interfaces/Portalicious/src/app/services/auth.service.ts
elwinschmitz: I think we can do without this (for now) and reconsider this while we work on the "add SSO-login"-features...

I'd like to limit all the introduced workarounds/edge-cases we start with.

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -0,0 +1,127 @@
...
+  localStorageKey = input.required<string>();
+  contextMenuItems = input<MenuItem[]>();
+  globalFilterFields = input<(keyof TData & string)[]>();
+  readonly onUpdateContextMenuItem = output<TData>();
+
+  @ViewChild('table') table: Table;
+  @ViewChild('extraOptionsMenu') extraOptionsMenu: Menu;
+
+  currentPageReportTemplate =
+    $localize`Showing [first] to [last] of [totalRecords] records`
```
elwinschmitz, line 76: We can/should add some explanation/warning to the context of this line about the variables/paceholders.

So adding: `$localize`:Make sure to keep placeholders between square-brackets as-is.:Showing [first]...`

See: https://devdocs.io/angular/api/localize/init/$localize

interfaces/Portalicious/src/styles.scss
```diff
@@ -124,4 +124,19 @@ $toolbar-height: 48px;
...
+  .p-card .p-datatable {
+    .p-datatable-header {
+      @apply px-0 pb-4 pt-0;
+    }
+
+    .p-paginator {
+      @apply border-b-0 pb-0 pt-4;
+
+      .p-paginator-current {
+        @apply me-auto ms-0 pl-0 font-display text-sm;
```
elwinschmitz, line 138: Almost...

```suggestion
        @apply me-auto ms-0 ps-0 font-display text-sm;
```
?

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.html
aberonni: It is currently defined in the [query-table component](https://github.com/global-121/121-platform/pull/5693/files#diff-1907126206c4fe9833dff805793412afcfc7a57c079e5ea32e58150178ff7d71R69) as

`globalFilterFields = input<(keyof TData & string)[]>();`

Which means TS will throw an error if you try to write a string that isn't a property of TData. TData is typed dynamically as the return type of the mutation.

elwinschmitz: Aha, so its sort-of a "one way connection you cannot violate from the other side". 👍

interfaces/Portalicious/src/styles.scss
elwinschmitz: Oh, this _is_ a good candidate to "set in stone" in the guidelines/checklist document.

"Make sure to use "`start`/`end`" instead of "`left`/`right`" in positioning/margin properties/classes".

elwinschmitz (approved):

interfaces/Portalicious/src/app/auth.guard.ts
aberonni: I'm resolving this assuming that we will address this weird pattern when we address the TODO

PR #5694, opened 2024-08-14
Portalicious: add new team member form AB#29678

arsforza:
[AB#29678](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29678)

## Describe your changes

### DIFF
--- a/interfaces/Portalicious/eslint.config.js
+++ b/interfaces/Portalicious/eslint.config.js
@@ -7,6 +7,7 @@ const eslintPluginPrettierRecommended = require('eslint-plugin-prettier/recommen
 const eslintPluginNoRelativePaths = require('eslint-plugin-no-relative-import-paths');
 const eslintPluginQuery = require('@tanstack/eslint-plugin-query');
 const eslintPluginPerfectionist = require('eslint-plugin-perfectionist');
+const eslintPluginRegexp = require('eslint-plugin-regexp');
 
 module.exports = tseslint.config(
   {
@@ -22,6 +23,7 @@ module.exports = tseslint.config(
     plugins: {
       'no-relative-import-paths': eslintPluginNoRelativePaths,
       perfectionist: eslintPluginPerfectionist,
+      regexp: eslintPluginRegexp,
     },
     extends: [
       eslint.configs.recommended,
@@ -29,6 +31,7 @@ module.exports = tseslint.config(
       ...tseslint.configs.stylisticTypeChecked,
       ...angular.configs.tsRecommended,
       ...eslintPluginQuery.configs['flat/recommended'],
+      eslintPluginRegexp.configs['flat/recommended'],
       eslintPluginPrettierRecommended,
     ],
     processor: angular.processInlineTemplates,
@@ -89,21 +92,22 @@ module.exports = tseslint.config(
         {
           checkId: false,
           ignoreAttributes: [
+            'app-query-table[localStorageKey]',
             'data-testid',
-            'icon',
-            'iconPos',
-            'ngSrc',
-            'styleClass',
-            'position',
-            'pSortableColumn',
             'field',
-            'display',
-            'sortField',
-            'appendTo',
-            'iconPosition',
-            'stateStorage',
-            'stateKey',
-            'localStorageKey',
+            'img[ngSrc]',
+            'inputStyleClass',
+            'ng-container[slot]',
+            'p-button[icon]',
+            'p-button[iconPos]',
+            'p-columnFilter[display]',
+            'p-contextMenu[appendTo]',
+            'p-iconField[iconPosition]',
+            'p-sidebar[position]',
+            'p-table[stateKey]',
+            'p-table[stateStorage]',
+            'styleClass',
+            'th[pSortableColumn]',
           ],
         },
       ],
--- a/interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
+++ b/interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
@@ -1,12 +1,9 @@
 <!-- eslint-disable-next-line @angular-eslint/template/label-has-associated-control -->
 <label class="block [&_input]:w-full">
-  <strong class="block">
+  <strong class="mb-2 inline-block">
     {{ label() }}
   </strong>
   <ng-content></ng-content>
-  <app-form-error
-    class="block"
-    [error]="errorMessage()"
-  />
+  <app-form-error [error]="errorMessage()" />
 </label>
 <ng-content select="[slot='details']"></ng-content>
--- a/interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
+++ b/interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
@@ -1,6 +1,12 @@
 <!-- eslint-disable-next-line @angular-eslint/template/label-has-associated-control -->
-<label class="[&_.p-password]:block [&_input]:w-full">
-  <span class="mb-2 inline-block">{{ label() }}</span>
+<label class="block [&_input]:w-full">
+  <strong class="block">
+    {{ label() }}
+  </strong>
   <ng-content></ng-content>
+  <app-form-error
+    class="block"
+    [error]="errorMessage()"
+  />
 </label>
-<app-form-error [error]="errorMessage()" />
+<ng-content select="[slot='details']"></ng-content>
--- a/interfaces/Portalicious/src/app/pages/login/login.component.html
+++ b/interfaces/Portalicious/src/app/pages/login/login.component.html
@@ -49,7 +49,6 @@
               autocomplete="current-password"
               [toggleMask]="true"
               [feedback]="false"
-              styleClass="block w-full"
             />
           </app-form-field-wrapper>
         </app-form-default>
--- a/interfaces/Portalicious/src/app/pages/login/login.component.html
+++ b/interfaces/Portalicious/src/app/pages/login/login.component.html
@@ -68,7 +68,7 @@
             href="https://manual.121.global/privacy-policy/"
             target="_blank"
             title="Opens in a new window"
-            i18n-title
+            i18n-title="@@generic-opens-in-new-window"
           >
             <span i18n>Read our Privacy Policy</span>
             <i
--- a/interfaces/Portalicious/src/app/pages/login/login.component.html
+++ b/interfaces/Portalicious/src/app/pages/login/login.component.html
@@ -49,6 +49,7 @@
               autocomplete="current-password"
               [toggleMask]="true"
               [feedback]="false"
+              styleClass="block w-full"
             />
           </app-form-field-wrapper>
         </app-form-default>
--- a/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
+++ b/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
@@ -86,7 +86,7 @@
         </p>
         <p i18n>
           * To use a subscope type a dot between the scope and subscope (e.g.
-          <code>amsterdam.zuid</code>)
+          <code>amsterdam.west</code>)
         </p>
       </ng-container>
     </app-form-field-wrapper>
--- a/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
+++ b/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
@@ -1,9 +1,8 @@
 <p-button
   label="Add team member"
   i18n-label="@@add-team-member"
-  outlined
+  rounded
   icon="pi pi-plus"
-  class="text-xs sm:text-base"
   (click)="formVisible.set(true)"
 />
 
@@ -18,19 +17,16 @@
     label="Choose user"
     i18n-label
     [errorMessage]="formFieldErrors()('userValue')"
-    class="mb-3"
   >
     <p-dropdown
-      [options]="unassignedUsers()"
+      [options]="availableUsers()"
       [optionLabel]="'username'"
       [optionValue]="'id'"
       [filter]="true"
       formControlName="userValue"
       placeholder="Type or select"
-      i18n-placeholder
-      class="block w-full"
-      styleClass="w-full"
-      [loading]="allUsers.isPending() || !projectUsers"
+      i18n-placeholder="@@generic-type-or-select"
+      [loading]="availableUsersIsLoading()"
     >
     </p-dropdown>
   </app-form-field-wrapper>
@@ -39,20 +35,14 @@
     label="Assign roles"
     i18n-label
     [errorMessage]="formFieldErrors()('rolesValue')"
-    class="mb-3"
   >
     <p-multiSelect
-      [options]="roles!.data()"
+      [options]="roles.data()"
       [optionLabel]="'label'"
       [optionValue]="'role'"
       formControlName="rolesValue"
       placeholder="Type or select"
-      i18n-placeholder
-      class="block w-full"
-      styleClass="w-full"
-      [panelStyleClass]="
-        'max-w-full [&_.p-multiselect-item]:items-start [&_.p-multiselect-item]:text-wrap'
-      "
+      i18n-placeholder="@@generic-type-or-select"
       [showToggleAll]="false"
       [overlayOptions]="{
         target: '@parent',
@@ -73,18 +63,20 @@
         type="text"
         formControlName="scopeValue"
         autocomplete="off"
-        i18n-placeholder
-        class="mb-2 font-mono"
+        class="font-mono"
       />
       <ng-container slot="details">
         <p
           i18n
-          class="mb-2"
+          class="my-2"
         >
           * Leaving this field empty means the team member will have access to
           <strong>all PAs</strong>
         </p>
-        <p i18n>
+        <p
+          i18n
+          class="my-2"
+        >
           * To use a subscope type a dot between the scope and subscope (e.g.
           <code>amsterdam.west</code>)
         </p>
--- a/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
+++ b/interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
@@ -1,7 +1,94 @@
 <p-button
-  label="Add user to team"
-  i18n-label
-  rounded
+  label="Add team member"
+  i18n-label="@@add-team-member"
+  outlined
   icon="pi pi-plus"
-  (click)="onAdd()"
+  class="text-xs sm:text-base"
+  (click)="formVisible.set(true)"
 />
+
+<app-form-sidebar
+  [(visible)]="formVisible"
+  [formGroup]="formGroup"
+  [mutation]="assignUserMutation"
+  formTitle="Add team member"
+  i18n-formTitle="@@add-team-member"
+>
+  <app-form-field-wrapper
+    label="Choose user"
+    i18n-label
+    [errorMessage]="formFieldErrors()('userValue')"
+    class="mb-3"
+  >
+    <p-dropdown
+      [options]="unassignedUsers()"
+      [optionLabel]="'username'"
+      [optionValue]="'id'"
+      [filter]="true"
+      formControlName="userValue"
+      placeholder="Type or select"
+      i18n-placeholder
+      class="block w-full"
+      styleClass="w-full"
+      [loading]="allUsers.isPending() || !projectUsers"
+    >
+    </p-dropdown>
+  </app-form-field-wrapper>
+
+  <app-form-field-wrapper
+    label="Assign roles"
+    i18n-label
+    [errorMessage]="formFieldErrors()('rolesValue')"
+    class="mb-3"
+  >
+    <p-multiSelect
+      [options]="roles!.data()"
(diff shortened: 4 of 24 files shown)


interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
```diff
@@ -1,6 +1,9 @@
 <!-- eslint-disable-next-line @angular-eslint/template/label-has-associated-control -->
-<label class="[&_.p-password]:block [&_input]:w-full">
-  <span class="mb-2 inline-block">{{ label() }}</span>
-  <ng-content></ng-content>
```
aberonni, line 4: The content was inside the label for a reason: this allows for the field to be connected to the appropriate label without extra hurdles (eg. making the sure the field has a `for` field that matches the `name` of the field). Connecting the field means (among other things)
- the field is selectable/focusable by clicking on the label
- screen readers correctly identify the field name

IMHO we need to revert this change.

interfaces/Portalicious/eslint.config.js
```diff
@@ -89,21 +92,22 @@ module.exports = tseslint.config(
...
+            'p-button[icon]',
+            'p-button[iconPos]',
+            'p-columnFilter[display]',
+            'p-contextMenu[appendTo]',
+            'p-iconField[iconPosition]',
+            'p-sidebar[position]',
+            'p-table[stateKey]',
+            'p-table[stateStorage]',
+            'styleClass',
+            'th[pSortableColumn]',
```
aberonni, line 110: Was this auto-generated somehow? If not, I fear that the specificity on some of these might come back to bite us (for example, I can imagine `icon` being relevant for some other component too)

interfaces/Portalicious/src/app/pages/login/login.component.html
```diff
@@ -49,6 +49,7 @@ <h1 i18n>Log in</h1>
               autocomplete="current-password"
               [toggleMask]="true"
               [feedback]="false"
+              styleClass="block w-full"
```
aberonni, line 52: Rather than duplicating this (I wonder if it is necessary also on the change-password page, for example?) I think it would be easier to leave the rules that were removed in `form-field-wrapper` component:

`<label class="[&_.p-password]:block [&_input]:w-full">`

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
 <p-button
-  label="Add user to team"
-  i18n-label
-  rounded
+  label="Add team member"
+  i18n-label="@@add-team-member"
+  outlined
```
aberonni, line 4: According to design:
```suggestion
  rounded
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
   icon="pi pi-plus"
-  (click)="onAdd()"
+  class="text-xs sm:text-base"
```
aberonni, line 6: I don't think we need to cater to small screens to this extent. If we must, then I would rather do it globally for primeng components in `styles.scss`. But again, I'd rather we didn't at all, and instead we tackle this systematically if and when it truly becomes relevant.

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+      [options]="unassignedUsers()"
+      [optionLabel]="'username'"
+      [optionValue]="'id'"
+      [filter]="true"
+      formControlName="userValue"
+      placeholder="Type or select"
+      i18n-placeholder
+      class="block w-full"
+      styleClass="w-full"
+      [loading]="allUsers.isPending() || !projectUsers"
```
aberonni, line 33: I think it would be better to pass in the result of `injectQuery` rather than the array directly, so that then we could do:

```suggestion
      [loading]="allUsers.isPending() || projectUsers.isPending()"
```

Or even

```suggestion
      [loading]="unassignedUsersIsPending()"
```

And 

```ts
unassignedUsersIsPending = computed(() => allUsers.isPending() || projectUsers.isPending())
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+    <p-dropdown
+      [options]="unassignedUsers()"
+      [optionLabel]="'username'"
+      [optionValue]="'id'"
+      [filter]="true"
+      formControlName="userValue"
+      placeholder="Type or select"
+      i18n-placeholder
+      class="block w-full"
+      styleClass="w-full"
```
aberonni, line 32: I think these classes which are being duplicated should probably be assigned either
- within `app-form-field-wrapper` or
- within styles.scss

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+  </app-form-field-wrapper>
+
+  <app-form-field-wrapper
+    label="Assign roles"
+    i18n-label
+    [errorMessage]="formFieldErrors()('rolesValue')"
+    class="mb-3"
+  >
+    <p-multiSelect
+      [options]="roles!.data()"
```
aberonni, line 45: I can't imagine the excl being necessary but maybe I'm wrong

```suggestion
      [options]="roles.data()"
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+  [formGroup]="formGroup"
+  [mutation]="assignUserMutation"
+  formTitle="Add team member"
+  i18n-formTitle="@@add-team-member"
+>
+  <app-form-field-wrapper
+    label="Choose user"
+    i18n-label
+    [errorMessage]="formFieldErrors()('userValue')"
+    class="mb-3"
```
aberonni, line 21: This could introduce inconsistencies. Same goes as for what I mentioned below with duplicated classes.
```suggestion
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+  >
+    <p-dropdown
+      [options]="unassignedUsers()"
+      [optionLabel]="'username'"
+      [optionValue]="'id'"
+      [filter]="true"
+      formControlName="userValue"
+      placeholder="Type or select"
+      i18n-placeholder
```
aberonni, line 30: This is being duplicated and could probably use some generic id (example below but feel free to change)

```suggestion
      i18n-placeholder=":@@form-generic-placeholder"
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+    label="Choose user"
+    i18n-label
+    [errorMessage]="formFieldErrors()('userValue')"
+    class="mb-3"
+  >
+    <p-dropdown
+      [options]="unassignedUsers()"
+      [optionLabel]="'username'"
+      [optionValue]="'id'"
+      [filter]="true"
```
aberonni, line 27:
```suggestion
      filter
```

I think? I generally prefer this syntax but maybe just me

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+      [options]="roles!.data()"
+      [optionLabel]="'label'"
+      [optionValue]="'role'"
+      formControlName="rolesValue"
+      placeholder="Type or select"
+      i18n-placeholder
+      class="block w-full"
+      styleClass="w-full"
+      [panelStyleClass]="
+        'max-w-full [&_.p-multiselect-item]:items-start [&_.p-multiselect-item]:text-wrap'
```
aberonni, line 54: Probably worth moving to `styles.scss`

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+      label="Assign scope"
+      i18n-label
+      [errorMessage]="formFieldErrors()('scopeValue')"
+    >
+      <input
+        pInputText
+        type="text"
+        formControlName="scopeValue"
+        autocomplete="off"
+        i18n-placeholder
```
aberonni, line 76:
```suggestion
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
```diff
@@ -1,7 +1,94 @@
...
+        class="mb-2 font-mono"
```
aberonni, line 77:
```suggestion
        class="font-mono"
```

Same as the other margin comments above

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.ts
```diff
@@ -1,23 +1,138 @@
...
+      },
+    },
+  );
+
+  allUsers = injectQuery(this.apiService.getAllUsers());
+
+  unassignedUsers = computed(() => {
+    this.projectUsers();
+    const allUsers = this.allUsers.data();
+    if (!this.projectUsers() || !allUsers) {
```
aberonni, line 100: I think this would be a safer/better check perhaps? Not sure though

```suggestion
    if (this.allUsers.isPending() || !this.projectUsers.isPending()) {
```

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.ts
```diff
@@ -1,23 +1,138 @@
...
+      });
 
-  onAdd(): void {
-    this.toastService.showToast({
-      summary:
-        'If you try really hard, I am sure you can imagine a new user being added to the table.',
-    });
-  }
+      void this.queryClient.invalidateQueries({
+        queryKey: this.apiService.getProjectUsers(this.projectId)().queryKey,
```
aberonni, line 134: Probably safer

```suggestion
        queryKey: this.apiService.getProject(this.projectId)().queryKey,
```

interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
```diff
@@ -22,7 +22,8 @@
       pInputText
       type="text"
       formControlName="assetId"
-      autocomplete="off"
+      autocomplete="one-time-code"
```
aberonni, line 25: 🤯

interfaces/Portalicious/tailwind.config.ts
```diff
@@ -10,6 +10,7 @@ export default {
     fontFamily: {
       display: ['Montserrat', 'sans-serif'],
       body: ['"Open Sans"', 'sans-serif'],
+      mono: ['monospace'],
```
aberonni, line 13:
```suggestion
```

I don't see this in the design system and I would avoid adding something extra that we need to maintain and/or enforce consistency on if it wasn't asked of us.

aberonni (changes requested):

interfaces/Portalicious/src/app/pages/login/login.component.html
elwinschmitz: My thinking was that this place(the login-page) is the only exception to the rule. So therefore I don't want to over-complicate the reusable component and keep that generic/simple.

interfaces/Portalicious/eslint.config.js
elwinschmitz: It was auto-sorted alphabetically. ;)

The condition I used to add a more specific element to the selector was their current use-cases in the code-base.
So, in the future, if there are other elements that really need the exception, the selector can/should be amended.

The easiest automated solution would be to use a list of "supported attributes per component" provided by any library we use (i.e. PrimeNG etc.); But that doesn't excist, so we could build it up over time...
Or use another reminder to "add attributes to support translations"...

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
elwinschmitz: See comment above @arsforza ...

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
elwinschmitz: huh  ¯\_(ツ)_/¯ ?
How did this end up here? ;) And weird that the linter doesn't complain the other way, that it finds i18n-attributes _without_ any source-attribute...

interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
elwinschmitz: Shht... 🤫

interfaces/Portalicious/tailwind.config.ts
elwinschmitz: I'd like to add features that help users, regardless if it is asked of me.

So, I tried initially to use the "Tailwind default font-stacks", but those where probably overridden by this config, so I added the bare-minimum to start with. And at some point other, more specific decisions could be added...

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
elwinschmitz: I prefer the other way. ;)

It emphasizes the 'booleanness' of it, in a way... And is a bit more explicit when reading... But I'm fine with both ways...

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
elwinschmitz: Probably never not a good idea, indeed.

interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
elwinschmitz: Sssh... 🤫

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
elwinschmitz: I thought it "too soon" to draw conclusions on "how we'll style Prime-components"...
But maybe we could already "look ahead" a bit further in the designs to spot some patterns already...

interfaces/Portalicious/src/app/pages/login/login.component.html
aberonni: > My thinking was that this place(the login-page) is the only exception to the rule.

That assumption is incorrect. The change password page is broken on the preview deployment.

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
aberonni: > Lets see if we need to re-align to prevent future rework...

Yes, I'm curious what was the reasoning here too, but we can still go ahead and fix in the meantime.

interfaces/Portalicious/src/app/pages/project/project-team/add-user-button/add-user-button.component.html
aberonni: I'd say it's safe to move to styles.scss like we've done for other components, and then we'll address this when we get to theming.

interfaces/Portalicious/tailwind.config.ts
aberonni: > I'd like to add features that help users, regardless if it is asked of me.

I'm not against that per se, but I see it as a problem when a dev introduces a design change, without input from design/PM. The assumption that "this helps users" is not one that they are necessarily on board with. And the risk of a design change made at this stage in the process is that we either
- End up having to revert the change because design / PM noticed it at a later stage or 
- A developer reverts this change because they notice it and it's a discrepancy with design (this could be me)

If you feel strongly about the monospace being necessary then I would do a couple of things
- apply it to the p-password in styles.scss rather than throughout the codebase (so we only do it one place, and it's easy to revert if we need to)
- add a note that we have done this in the "design review" task for this PBI, so we can revert if we need to

interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
aberonni, line 1: I'm not sure if the regression is in this file, but I noticed that the input fields are not turning red anymore when they have an error.

interfaces/Portalicious/src/app/pages/projects-overview/create-project-button/create-project-button.component.html
```diff
@@ -32,9 +33,10 @@
   >
     <p-password
       formControlName="token"
-      autocomplete="off"
+      autocomplete="one-time-code"
       [toggleMask]="true"
       [feedback]="false"
+      inputStyleClass="font-mono"
```
aberonni, line 39: Can we move this to `style.scss` and apply to all p-password fields?

interfaces/Portalicious/src/styles.scss
```diff
@@ -135,4 +135,22 @@ $toolbar-height: 48px;
       }
     }
   }
+
+  // Form inputs (Overrides for PrimeNG-components)
+  // All inputs used inside a form-field-wrapper, should be full-width, the wrapper should be in control of the width.
+  app-form-field-wrapper {
```
aberonni, line 141: I'm a bit concerned about using one of our custom component selectors here - this could change without us realising that we're breaking this. Can we apply these rules without the wrapper? If not, then this might be an exception where having a dedicated CSS file for a component (app-form-field-wrapper) might make sense.

aberonni (changes requested):
The regression on the errored input fields is really the only important change - the others I'll leave up to you to decide what to do with.

interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
elwinschmitz: They are sometimes at least...
But I did find this situation:
Inconsistent error-styling on change-password form-fields

And also, the `:focus`-style 'win' over the error-state-styles... (So the border of a focussed field will be blue, instead of red...)

interfaces/Portalicious/src/styles.scss
elwinschmitz: Yeah, I was a bit in doubt on how big I wanted the "impact area" to be...
This was my solution to have it work "opt in" (and keep the default styles available on other/future/yet-unknown use-cases)..

But the other approach could be indeed to "change the default" and deal with the scenario's of "but for this situation we need an exception" later...
The only thing is that currently I don't know the difference between a "one-off exceptional change" and a "this is how we always want it to be", so I try to stay on the (too?) cautious side.

interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
aberonni: It still doesn't work for me.

To reproduce:
- Go to projects page (eg. https://portalicious.test.121.global/en/projects)
- Click on "add project"
- Click on both fields and click out

Expected:
I see this on https://portalicious.test.121.global/en/projects

image

Actual:
I see this on https://lively-river-04adce503-5694.westeurope.5.azurestaticapps.net/en/projects
image

interfaces/Portalicious/src/styles.scss
aberonni: I saw the additions to `styles.scss`, including the re-org and new comments. They look great, thanks! ✨

aberonni (changes requested):
Still experiencing the same regression

interfaces/Portalicious/src/app/components/form-field-wrapper/form-field-wrapper.component.html
elwinschmitz: I also see the last screenshot on <https://portalicious.test.121.global/en/projects>
(With any password-manager disabled/not-active).

I think this has more to do with the logic around the `Validators` (and `control.touched`) on the `FormGroup` then on any styling or nesting/structuring changes introduced in this PR.

As shown in my other screenshot, the behavior is not consistent in all situations..
Also consider:
- open the "All Project"-page
- Click on "+ Add Project"
- Directly click on the Submit-button (without touching the form)

So we might want to look into the way the validation-error-message is added (and the status of the field is _not_ updated; or have a different combination of conditions then only `.touched`.

But it doesn't seem directly related to the changes in this PR.

aberonni (approved):

PR #5710, opened 2024-08-20
Improvements Visa migration endpoint

PeterSmallenbroek:
## Describe your changes

### DIFF
--- a/services/121-service/src/migrate-visa/migrate-visa.service.ts
+++ b/services/121-service/src/migrate-visa/migrate-visa.service.ts
@@ -35,7 +35,7 @@ export class MigrateVisaService {
   name = 'VisaMigrateChildParentWallet1717505534921';
   public tokenSet: TokenSet;
 
-  public async migrateData(): Promise<void> {
+  public async migrateData(limit: number): Promise<void> {
     const queryRunner = this.dataSource.createQueryRunner();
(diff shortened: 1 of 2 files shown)


services/121-service/src/migrate-visa/migrate-visa.service.ts
```diff
@@ -321,7 +388,10 @@ export class MigrateVisaService {
       from
         "121-service"."intersolve_visa_customer" i
       left join "121-service".registration r on
-        r.id = i."registrationId" WHERE "programId" = ${programId}`,
+        r.id = i."registrationId" 
+      LEFT JOIN "121-service"."intersolve_migration_progress" imp ON r."referenceId" = imp."referenceId"
```
RubenGeo, line 392: So if we have to run this a second time because of errors. We manually have to clean them from this table?

RubenGeo:
Great idea with the extra table!

Ideas for testing: (which you probably already did):
- With a few manually created registrations against our mock service (and use the mock to let one or two api request fail)
- Testing against the intersolve acceptance/staging server (with scenario's similair to what we had at the acceptance test)
- Maybe it is good to retest using a dump from production data with our mock service. (Also to see/guess how long the migration will take)

RubenGeo (approved):

PR #5711, opened 2024-08-20
PoC: Tweak local infrastructure to ease switching branches

elwinschmitz:
This is a change to the local-development setup.
It will create a "per-branch" environment using Docker.

So the biggest change in workflow would be that you'd need to explicitly "stop" any environment your currently work in BEFORE switching/creating another git-branch.

### Breaking changes
the "121-redis"-container is renamed to "121queue", so any references to it in your local `services/.env`-file need to be changed (see [`.env.example`](services/.env.example).

Also use of the ENV-variable: `COMPOSE_PROJECT_NAME` is no longer necessary.

Any other comments? Let me know!


elwinschmitz:
Branch renamed/moved to: https://github.com/global-121/121-platform/compare/elwin/poc/branches-dockers

PR #5715, opened 2024-08-20
portalicious: fix: all page content has a tooltip

aberonni:
[AB#29765](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29765)

note to self: do not use "title" as a component attribute that wraps the whole page


elwinschmitz (approved):
Ah, wow.

Super-logical that "using a default HTML-attribute" will "just work as it supposed to";

But also, super-weird that the whole "data-binding-magic" isn't a totally-separate thing from DOM-reality...

🤯

PR #5720, opened 2024-08-22
portalicious: context menu alignment fix

aberonni:
[AB#29766](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29766)

### DIFF
--- a/interfaces/Portalicious/src/app/components/query-table/query-table.component.html
+++ b/interfaces/Portalicious/src/app/components/query-table/query-table.component.html
@@ -8,7 +8,7 @@
     [filterLocale]="locale"
     (onFilter)="onFilter($event)"
     [rowHover]="contextMenuItems()"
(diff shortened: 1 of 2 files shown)


elwinschmitz (approved):
Not a super-dirty or super-complicated fix after all...

Although I still think we just don't need the "right-click-override-menu"... Because it messes with selecting text and using "copy" from the browser/OS context-menu, or other OS/browser-features... etc.

aberonni:
> Although I still think we just don't need the "right-click-override-menu"... Because it messes with selecting text and using "copy" from the browser/OS context-menu, or other OS/browser-features... etc.

Peter also wasn't a fan of that. I would bring it up with Tijs if you feel strongly about it. Personally, I don't mind it.

PR #5726, opened 2024-08-23
portalicious: split api service into domains

aberonni:
[AB#29889](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29889)


interfaces/Portalicious/src/app/domains/domain-api.service.ts
```diff
@@ -0,0 +1,54 @@
...
+  >({
+    path,
+    processResponse,
+    ...opts
+  }: {
+    path: Parameters<typeof DomainApiService.prototype.pathToQueryKey>[0];
+    processResponse?: (data: BackendDataShape) => ProcessedResponseShape;
+  } & Partial<UndefinedInitialDataOptions<ProcessedResponseShape>>) {
+    return () => {
+      console.log('path', path);
```
elwinschmitz, line 30:
```suggestion
```

interfaces/Portalicious/src/app/domains/payment/payment.api.service.ts
```diff
@@ -0,0 +1,41 @@
+import { Injectable, Signal } from '@angular/core';
+import { DomainApiService } from '~/domains/domain-api.service';
+import { Payment } from '~/domains/payment/payment.model';
+
+const BASE_ENDPOINT = (paymentId: Signal<number>) => [
+  'programs',
+  paymentId,
+  'payments',
+];
```
elwinschmitz, line 9:
```suggestion
const BASE_ENDPOINT = (projectId: Signal<number>) => [
  'programs',
  projectId,
  'payments',
];
```

Right?

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.ts
```diff
@@ -56,19 +54,19 @@ export class ProjectTeamComponent {
...
-      this.apiService.removeProjectUser(this.projectId(), userId),
+      this.projectApiService.removeProjectUser(this.projectId, userId),
     onSuccess: () => {
       this.toastService.showToast({
         detail: $localize`User removed`,
       });
-      void this.queryClient.invalidateQueries({
-        queryKey: this.apiService.getProject(this.projectId)().queryKey,
-      });
+      void this.projectApiService.invalidateCache(this.projectId);
```
elwinschmitz, line 69: There is currently no way to invalidate the "project-users"-query, right?
Because now its not 'tied' to the endpoint/request specifically.

Here, both the queries `projectApiService.getProject()` and `projectApiService.getProjectUsers()` should be invalidated (separately), right?
(Or is the invalidation of 1 cascading to invalidate the other?)

Of should this be tackled by amending the functionality of `invalidateCache` to include an aditional `path`-parameter and calling `invalidateCache(projectId, 'users');` (Or similar)...

interfaces/Portalicious/src/app/domains/payment/payment.api.service.ts
aberonni: Nee, all of the payments endpoints are a subset of the project controller. This was the thing preventing me from separate payments from projects in the previous iteration. You can, for example, get all of the payments for a specific project and for that you do not need a paymentId but you always need a projectId (because all payments are tied to a project).

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.ts
aberonni: > There is currently no way to invalidate the "project-users"-query, right?

This is still possible, if necessary, with something like this:

```ts
      void this.queryClient.invalidateQueries({
        queryKey: this.apiService.getProjectUsers(this.projectId)().queryKey,
      });
 ```
 
> Here, both the queries projectApiService.getProject() and projectApiService.getProjectUsers() should be invalidated (separately), right?

Right, but not separately.

> (Or is the invalidation of 1 cascading to invalidate the other?)

Kinda. The way I've currently built it, all of the queries that rely on the projectId will be invalidated when `invalidateCache(projectId)` is used, because they all share the same "beginning of the queryKey array" AKA `['programs', projectId()]`

> Of should this be tackled by amending the functionality of invalidateCache to include an aditional path-parameter and calling invalidateCache(projectId, 'users'); (Or similar)...

I don't think that's necessary (now/yet/potentially ever)

File: interfaces/Portalicious/src/app/domains/project/project.api.service.ts [hunk 4]
```diff
@@ -0,0 +1,87 @@
...
+        koboAssetId: assetId,
+      },
+    });
+  }
+
+  removeProjectUser(projectId: Signal<number>, userId: number) {
+    return this.httpWrapperService.perform121ServiceRequest<Project>({
+      method: 'DELETE',
+      endpoint: `${BASE_ENDPOINT}/${projectId().toString()}/users/${userId.toString()}`,
+    });
```
elwinschmitz, line 73: We could potentially also tie the "invalidation of a related cache" at this point?

That would turn this class more into a "project-data management"-service then a "more dumb" "only talk to the API"-service....
I don't know how much that would look like a slippery slope or not though...

It would remove _some_ duplication maybe and would help developers "never forget to invalidate the relevant caches"....

Or could we formalize a rule like: "only data-manipulation, not business-logic" inside these API-Services?

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.ts
elwinschmitz: Ah, the invalidation happens in a "namespaced" way, not on a simple key-value matching only...
https://tanstack.com/query/latest/docs/framework/angular/guides/query-invalidation#query-matching-with-invalidatequeries

So in this use-case "a bit too much" gets invalidated (buts is saver to be on that side then using data that's 'potentially outdated', in this case.)...

Potentially this approach could be a bit harmful; For example that it would also invalidate a (heavy) query for `programs/1/transactions` (or something unrelated to `users`)..
But lets see when we run into some of those 'inefficiencies' in real life. ;)

interfaces/Portalicious/src/app/domains/project/project.api.service.ts
aberonni: > Or could we formalize a rule like: "only data-manipulation, not business-logic" inside these API-Services?

That would be my preference, yes, which is why they are all `api.service.ts`

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.ts
aberonni: > So in this use-case "a bit too much" gets invalidated (buts is saver to be on that side then using data that's 'potentially outdated', in this case.)...

Yes, exactly.

> Potentially this approach could be a bit harmful; For example that it would also invalidate a (heavy) query for programs/1/transactions (or something unrelated to users)..

I think there is a potential for telling angular-query to only allow a specific query to be invalidated with an exact match rather than by namespace - but I'm not entirely sure.

interfaces/Portalicious/src/app/components/form/form.component.ts
```diff
@@ -10,13 +10,13 @@ import { CreateMutationResult } from '@tanstack/angular-query-experimental';
  * - `FormSidebarComponent`
  */
 @Component({
-  selector: 'app-form-base-do-not-use',
+  selector: 'app-abstract-form-do-not-use',
```
elwinschmitz, line 13: You couldn't resist to also include a "naming things"-code change, next the all the "invalidating cache"-code changes, right? :D

interfaces/Portalicious/src/app/pages/project/project-team/project-team.component.ts
elwinschmitz: There's a `exact: true`-parameter I think, looking at that documentation-page linked above...

interfaces/Portalicious/src/app/components/form/form.component.ts
aberonni: image

PR #5734, opened 2024-08-28
feat: add the migration script to update safaricom_requests database table

ashishkumarbasak-accenture:
[AB#29906](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/29906)

## Describe your changes

Added a new migration script to drop un-necessary columns from `safaricom_requests` database table and keep only the required columns with renamed column names according to [this miro board](https://miro.com/app/board/uXjVKzM7CFE=/)

### DIFF
--- a/services/121-service/src/migration/1724754547895-modify-safaricom-requests-table.ts
+++ b/services/121-service/src/migration/1724754547895-modify-safaricom-requests-table.ts
@@ -54,6 +54,24 @@ export class ModifySafaricomRequestsTable1724754547895
     await queryRunner.query(
       `ALTER TABLE "121-service"."safaricom_request" RENAME COLUMN "transactionId" TO "mpesaTransactionId"`,
     );
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."safaricom_request" RENAME COLUMN "originatorConversationID" TO "originatorConversationId"`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."safaricom_request" DROP CONSTRAINT "FK_76a85e893850c2ef7ce4b6441a0"`,
+    );
+    await queryRunner.query(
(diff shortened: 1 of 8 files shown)


services/121-service/src/migration/1724754547895-modify-safaricom-requests-table.ts
```diff
@@ -0,0 +1,66 @@
...
+      `ALTER TABLE "121-service"."safaricom_request" DROP COLUMN "paymentResult"`,
+    );
+
+    // Step 2: Rename the necessary columns
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."safaricom_request" RENAME COLUMN "conversationID" TO "mpesaConversationId"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."safaricom_request" RENAME COLUMN "transactionId" TO "mpesaTransactionId"`,
+    );
```
diderikvw, line 56: Also rename column originatorConversationID to originatorConversationId (looks like a typo in the code and the diagram).

diderikvw (changes requested):
Migration script looks good: in principle it creates the data structure as it should.

Except:
1. A typo: see specific comment below.

2. When running the command to generate a migration, it still generates a migration script, which it should not:
image

Please adjust the migration script so when running this command it comes up with nothing.

Note: it looks like generating a migration on main also results in a migration script: relating to the view and typeorm meta data. So that seems like a problem in main, not per-se this branch.

services/121-service/src/migration/1724754547895-modify-safaricom-requests-table.ts
ashishkumarbasak-accenture: Updated!

ashishkumarbasak-accenture:
> Migration script looks good: in principle it creates the data structure as it should.
> 
> Except:
> 
> 1. A typo: see specific comment below.
> 2. When running the command to generate a migration, it still generates a migration script, which it should not:
>    image
> 
> Please adjust the migration script so when running this command it comes up with nothing.
> 
> Note: it looks like generating a migration on main also results in a migration script: relating to the view and typeorm meta data. So that seems like a problem in main, not per-se this branch.

I have updated the MR.
Now the `npm run migration:generate` command is not generating migration script. I have fixed the issues related to my migration script & db table `safaricom_request`. Could you please recheck again.

Screenshot 2024-08-30 at 08 56 00

PR #5741, opened 2024-08-30
portalicious: edit user on teams page

aberonni:
[AB#29744](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29744)

When discussing with Elwin, we had settled on a different approach: The idea was to get rid of the separate component for managing the form and moving everything into the page.

When I started doing this, though, I realised that maybe there wasn't a need for it, and instead the only logic that needed to be moved to the page was whether to show the form or not.

So, I've also edited the "create-project-button" component accordingly so that they both follow the same pattern.

Still not sure if this is the wisest approach but it works.

EDIT: I also sneaked a couple of design fixes in there too.

### DIFF
--- a/interfaces/Portalicious/src/app/pages/project/project-team/add-user-form/add-user-form.component.ts
+++ b/interfaces/Portalicious/src/app/pages/project/project-team/add-user-form/add-user-form.component.ts
@@ -2,9 +2,10 @@ import {
   ChangeDetectionStrategy,
   Component,
   computed,
+  effect,
   inject,
   input,
-  signal,
+  model,
 } from '@angular/core';
 import {
   FormControl,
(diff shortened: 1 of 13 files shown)


interfaces/Portalicious/src/app/pages/project/project-team/add-user-form/add-user-form.component.ts
```diff
@@ -50,23 +52,49 @@ type AddUserToTeamFormGroup =
     ReactiveFormsModule,
   ],
 })
-export class AddUserButtonComponent {
+export class AddUserFormComponent {
   projectId = input.required<number>();
-  enableScope = input.required<boolean | undefined>();
+  enableScope = input.required<boolean>();
+  formVisible = model.required<boolean>();
```
elwinschmitz, line 58: Aha! `model` was the trick I was looking for... I fiddled way to long with `input`/`signal` etc.. 🤦

interfaces/Portalicious/src/app/pages/project/project-team/add-user-form/add-user-form.component.ts
```diff
@@ -50,23 +52,49 @@ type AddUserToTeamFormGroup =
...
+  userToEdit = input<ProjectUserWithRolesLabel | undefined>();
 
   private projectApiService = inject(ProjectApiService);
   private userApiService = inject(UserApiService);
   private roleApiService = inject(RoleApiService);
   private toastService = inject(ToastService);
 
+  constructor() {
+    effect(
```
elwinschmitz, line 67: This is basically how you would expect/_want_ the `ngOnChanges`-thing to work...

Still a bit magical. ;)

elwinschmitz (approved):
A nice hybrid solution after all. :)

elwinschmitz (approved):

PR #5773, opened 2024-09-03
portalicious: add note form

aberonni:
[AB#29983](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/29983)

### DIFF
--- a/interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
+++ b/interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
@@ -9,10 +9,13 @@
   (onHide)="formGroup().reset()"
 >
   <ng-template pTemplate="headless">
+    <!-- eslint-disable-next-line @angular-eslint/template/interactive-supports-focus -->
     <form
       class="flex min-h-full w-full flex-col p-4"
       [formGroup]="formGroup()"
       (ngSubmit)="onFormSubmit()"
+      (keydown.control.enter)="onFormSubmit()"
+      (keydown.meta.enter)="onFormSubmit()"
     >
       <div class="flex grow flex-col gap-2">
         <h2 class="mb-5">
--- a/interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
+++ b/interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
@@ -1,0 +1,41 @@
+<app-form-sidebar
+  [(visible)]="formVisible"
+  [formGroup]="formGroup"
+  [mutation]="addNoteMutation"
+  formTitle="Add note"
+  i18n-formTitle="@@add-note"
+>
+  <p
+    i18n
+    class="font-semibold"
+  >
+    You are about to add a note to {{ registrationName() }}'s profile.
+  </p>
+  <p
+    i18n
+    class="mb-8"
+  >
+    The note will be visible to other 121 users who are permitted to view
+    personal profiles.
+  </p>
+  <app-form-field-wrapper
+    label="Note"
+    i18n-label
+    [errorMessage]="formFieldErrors()('note')"
+  >
+    <textarea
+      rows="5"
+      pInputTextarea
+      formControlName="note"
+      autocomplete="off"
+    ></textarea>
+    <ng-container slot="details">
+      <p
+        i18n
+        class="text-grey-500 txt-system-s"
+      >
+        Tip: Press Command / Ctrl + Enter to submit.
+      </p>
+    </ng-container>
+  </app-form-field-wrapper>
+</app-form-sidebar>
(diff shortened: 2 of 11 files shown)


interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
```diff
@@ -0,0 +1,33 @@
...
+    label="Note"
+    i18n-label
+    [errorMessage]="formFieldErrors()('note')"
+  >
+    <textarea
+      rows="5"
+      pInputTextarea
+      formControlName="note"
+      autocomplete="off"
+    ></textarea>
```
elwinschmitz, line 31: No hint-text about the keyboard-shortcut here? (like in the screenshot?)

interfaces/Portalicious/src/styles.scss
```diff
@@ -190,7 +190,8 @@ body {
 
   .p-element.ng-touched.ng-invalid {
     .p-inputwrapper,
-    .p-inputtext {
+    .p-inputtext,
+    &.p-inputtextarea {
```
elwinschmitz, line 194: Hmm.. shouldn't `.p-inputtext` _also_ work with the same kind of selector?

I think the `&` is missing there as well...
See:
image

interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
```diff
@@ -9,10 +9,13 @@
   (onHide)="formGroup().reset()"
 >
   <ng-template pTemplate="headless">
+    <!-- eslint-disable-next-line @angular-eslint/template/interactive-supports-focus -->
     <form
       class="flex min-h-full w-full flex-col p-4"
       [formGroup]="formGroup()"
       (ngSubmit)="onFormSubmit()"
+      (keydown.control.enter)="onFormSubmit()"
+      (keydown.meta.enter)="onFormSubmit()"
```
elwinschmitz, line 18: Hmm.. would it maybe be necessary to use a little key-event-specific wrapper-method, in which we could check whether this `form` also currently has focus within (or simply if its visible)?

(Hmm. maybe the angular-logic will already check for whether this key-event is triggered from 'inside' this form?)

To prevent false-positives when there are maybe multiple `form`-elements still available in the DOM at the same time?

elwinschmitz (dismissed):
The bugfix of the `inputtext`-element is maybe out-of-scope for this item/PR... and it would probably need some more comparison/testing..

The rest looks good; And I don't know if my suggestion for the form-submit really makes any difference...

interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
aberonni: > (Hmm. maybe the angular-logic will already check for whether this key-event is triggered from 'inside' this form?)

Indeed, it looks like that is the case (from my very limited testing)

interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
aberonni: I missed it in the screenshot! Adding it now :)

interfaces/Portalicious/src/styles.scss
aberonni: Ha, looks like for input fields we need `&.p-inputtext` and for password fields we need`.p-inputtext` and I am totally not upset with primeng for this lack of consistency

interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
```diff
@@ -0,0 +1,41 @@
...
+    <ng-container slot="details">
+      <p
+        i18n
+        class="text-grey-500 txt-system-s"
+      >
+        Tip: Press Command / Ctrl + Enter to submit.
```
elwinschmitz, line 37: Oe. missed opportunity to use `<kbd>`-elements! :D

elwinschmitz (approved):

interfaces/Portalicious/src/app/components/page-layout/components/registration-header/add-note-form/add-note-form.component.html
aberonni: 🫨  I was today years old when I learned.... I am making this change now :D

aberonni: Nevermind - I tried it and I didn't like it 😄

elwinschmitz: Yeah, I wondered what the default/PrimeNG styling would be.. ;)
But the "semantic"-benefits are negligible, I think...

Maybe, some day, when we add a `⌘+K` power-users keyboard-shortcuts-panel, etc.. ;)

PR #5788, opened 2024-09-05
chore: add findOne endpoint to be used in portalicious

aberonni:
[AB#30126](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30126)

### DIFF
--- a/services/121-service/src/registration/registrations.controller.ts
+++ b/services/121-service/src/registration/registrations.controller.ts
@@ -10,6 +10,7 @@ import {
   ImportRegistrationsDto,
   ImportResult,
 } from '@121-service/src/registration/dto/bulk-import.dto';
+import { MappedPaginatedRegistrationDto } from '@121-service/src/registration/dto/mapped-paginated-registration.dto';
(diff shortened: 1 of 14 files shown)


services/121-service/src/registration/registrations.controller.ts
```diff
@@ -707,4 +708,42 @@ export class RegistrationsController {
...
+    type: 'integer',
+  })
+  @Get('programs/:programId/registrations/:id')
+  public async findOne(
+    @Param('programId', ParseIntPipe) programId: number,
+    @Param('id', ParseIntPipe) id: number,
+  ): Promise<MappedPaginatedRegistrationDto> {
+    const registrationEntity =
+      await this.registrationsService.getPaginateRegistrationById({
+        id,
```
aberonni, line 736: @PeterSmallenbroek FYI I took an executive decision and went ahead and already made the change to use registration id, with the assumption that this is the direction we're headed. We can always switch to paID / programRegistrationId if we have to.

PeterSmallenbroek (approved):

PR #5807, opened 2024-09-10
fix: Stop testing Third-Party API in CI-tests

elwinschmitz:
[AB#30220](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30220)

(Temporarily?) Disable testing of third-party APIs.

### DIFF
--- a/e2e/portalicious/tests/CreateNewProject/CreateNewProjectSuccessfully.spec.ts
+++ b/e2e/portalicious/tests/CreateNewProject/CreateNewProjectSuccessfully.spec.ts
@@ -18,6 +18,8 @@ test.beforeEach(async ({ page }) => {
 });
 
 test('[29635] Create project successfully', async ({ page }) => {
+  test.fixme(!!process.env.CI, 'Disable use of third-party API in CI only');
+
   const homePage = new HomePage(page);
   const createProject = new CreateProject(page);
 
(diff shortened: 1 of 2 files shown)


e2e/portalicious/tests/CreateNewProject/CreateNewProjectSuccessfully.spec.ts
```diff
@@ -18,6 +18,8 @@ test.beforeEach(async ({ page }) => {
 });
 
 test('[29635] Create project successfully', async ({ page }) => {
+  test.fixme(!!process.env.CI, 'Disable use of third-party API in CI only');
```
aberonni, line 21: I'm not sure why we are disabling this test. Assuming there are Reasons™, it would be nice to create task to address those reasons, and to link that task here as a "TODO" comment

aberonni (dismissed):
See comment

e2e/portalicious/tests/CreateNewProject/CreateNewProjectSuccessfully.spec.ts
Piotrk39: I agree with the above... we can disable them for time-being of course but good idea is to think as well about long lasting solution eg. Mock service or using dev server of Kobo for that part.

Piotrk39 (approved):

e2e/portalicious/tests/CreateNewProject/CreateNewProjectSuccessfully.spec.ts
elwinschmitz: The main item to address this in a more future-proof way, is currently AB#30220.

PR #5824, opened 2024-09-12
fix: First slice of fix issues AB#30290

Vladimir-Petkov:
[AB#30237](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30237)


services/121-service/src/events/dto/get-event.dto.ts
```diff
@@ -3,7 +3,7 @@ import { WrapperType } from '@121-service/src/wrapper.type';
 import { ApiProperty } from '@nestjs/swagger';
 
 class AttributesDto {
-  [key: string]: string;
+  [key: string]: string | unknown;
```
aberonni, line 6: I feel like this change might come back to haunt you, and you might have to adjust it further, but it is ok to merge it as is for now.

services/121-service/src/events/events.service.spec.ts
```diff
@@ -192,7 +192,7 @@ describe('EventsService', () => {
       chunk: 2000,
     });
   });
-
+  // I don`t know how to fix this.
```
aberonni, line 195: We can have a look together

services/121-service/src/events/events.service.ts
```diff
@@ -259,19 +269,27 @@ export class EventsService {
 
     const events: EventEntity[] = [];
     for (const fieldName of fieldNames) {
+      const key = fieldName as keyof LogEntity;
```
aberonni, line 272: I wonder if you can make `keyof LogEntity` the return type of `getRelevantRegistrationViewKeys` so that you can remove this ` as `

services/121-service/src/events/events.service.ts
```diff
@@ -259,19 +269,27 @@ export class EventsService {
...
+          isMatch(oldEntity[key], newEntity[key]))
       ) {
         continue;
       }
 
+      // Ensure only string values are passed to the event creation
+      const oldValue =
+        typeof oldEntity[key] === 'string' ? oldEntity[key] : null;
+      const newValue =
+        typeof newEntity[key] === 'string' ? newEntity[key] : null;
```
aberonni, line 287: Maybe we should throw an error instead, because otherwise we are silently "failing"

```suggestion
      if (typeof oldEntity[key] !== 'string') { throw Error('something something'); }
      if (typeof newEntity[key] !== 'string') { throw Error('something something'); }
```

Or could the `createEventForChange` method also support different types? Maybe also numbers?

services/121-service/src/events/events.service.ts
```diff
@@ -285,13 +303,14 @@ export class EventsService {
...
-    registrationdId: number,
+    registrationId: number,
   ): EventEntity {
     const event = new EventEntity();
     event.type = this.getEventType(fieldName);
-    event.registrationId = registrationdId;
+    event.registrationId = registrationId;
 
-    const attributesData = {
+    // Explicitly declare attributesData as Record<string, any> to allow dynamic keys
```
aberonni, line 312:
```suggestion
    // Explicitly declare attributesData as Record<string, unknown> to allow dynamic keys
```

services/121-service/src/events/utils/events.mapper.ts
```diff
@@ -41,10 +41,12 @@ export class EventsMapper {
 
   private static createAttributesObject(
     attributes: EventAttributeEntity[],
-  ): Record<string, any> {
+  ): Record<string, EventAttributeEntity['value']> {
     const attributesObject: Record<string, EventAttributeEntity['value']> = {};
     for (const attribute of attributes) {
-      attributesObject[attribute.key] = attribute.value;
+      if (attribute.value !== null) {
```
aberonni, line 47: This is a change in bevhaiour which probably makes sense but it should also be tested somehow (I'm not sure how)

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -81,12 +82,19 @@ export class RegistrationsController {
   @ApiBody(FILE_UPLOAD_API_FORMAT)
   @UseInterceptors(FileInterceptor('file'))
   public async importRegistrations(
-    @UploadedFile() csvFile,
+    @UploadedFile() csvFile: unknown,
```
aberonni, line 85: Maybe it is 

```suggestion
    @UploadedFile() csvFile: Blob,
```

But I don't know

aberonni (commented):
Great job so far!

services/121-service/src/registration/registrations.controller.ts
Vladimir-Petkov: Agree, and change it.

services/121-service/src/events/utils/events.mapper.ts
Vladimir-Petkov: Agree, but when we start all different tests we can see. I think for now we can keep it like this.

services/121-service/src/events/events.service.ts
Vladimir-Petkov: Done, copy mistake.

services/121-service/src/events/events.service.ts
Vladimir-Petkov: @aberonni  What do you think about this: 
const oldValue =
        typeof oldEntity[key] === 'string'
          ? oldEntity[key]
          : (() => {
              throw Error('something something');
            })();
      const newValue =
        typeof newEntity[key] === 'string'
          ? newEntity[key]
          : (() => {
              throw Error('something something');
            })();

services/121-service/src/events/events.service.ts
Vladimir-Petkov: I have update both methods to make key typed correctly.

services/121-service/src/events/dto/get-event.dto.ts
Vladimir-Petkov: What do you think about this:
class AttributesDto {
  [key: string]: string | number | boolean | unknown;
}

services/121-service/src/events/events.service.spec.ts
Vladimir-Petkov: If you have today, you can write me and check this.

services/121-service/src/events/dto/get-event.dto.ts
Vladimir-Petkov: We have to talk about this change, because in we have different types and I think this will be good to be:
[key: string]: string | number | unknown;

aberonni: I think, as far as typescript is concerned, doing `string | unknown` or `string | number | unknown` is the same as just doing `unknown` because it still doesn't know what type it is.

When I say 

> I feel like this change might come back to haunt you

it is because I think this is used in a lot of places, and if you change it to `unknown` then you have to do type checking in a lot of places.

I would leave it as `string | number` if possible. But you can decide what is best (you have more context) and then change it accordingly.

services/121-service/src/events/events.service.spec.ts
aberonni: If it's ok, I'd say to merge this like this:
```suggestion
  // XXX: I don`t know how to fix this.
```

So we can find it again, and then we come back to it once I'm back from holidays.

services/121-service/src/events/dto/get-event.dto.ts
Vladimir-Petkov: Then, maybe is better to use null, because we have this:
image

services/121-service/src/events/events.service.ts
aberonni: I think maybe this is cleaner:


```suggestion
      if (typeof oldEntity[key] !== 'string') { throw Error('something something'); }
      if (typeof newEntity[key] !== 'string') { throw Error('something something'); }
      
      const oldValue = oldEntity[key] as string;
      const newValue = newEntity[key] as string;
```

PS if you wrap your text with ``` it makes it a lot easier to read, something like:

image

Will render like this:

```ts
const oldValue =
        typeof oldEntity[key] === 'string'
          ? oldEntity[key]
          : (() => {
              throw Error('something something');
            })();
      const newValue =
        typeof newEntity[key] === 'string'
          ? newEntity[key]
          : (() => {
              throw Error('something something');
            })();
```

services/121-service/src/events/events.service.spec.ts
Vladimir-Petkov: okay, I will add comments on everything that is not easy to fix.

services/121-service/src/events/utils/events.mapper.ts
aberonni: > when we start all different tests we can see

I'm not sure what you mean by this? Ideally this branch should be merged into main as is, so any relevant tests should already happen on this branch.

services/121-service/src/events/dto/get-event.dto.ts
aberonni: Ah yes, then `string | null` should be good indeed

aberonni (dismissed):
I'm not too clear on your comment RE: testing the change in logic, but so long as everything works (I didn't check out the code locally or test anything), the code looks good to me.

services/121-service/src/events/dto/get-event.dto.ts
Vladimir-Petkov: Done.

services/121-service/src/events/events.service.ts
Vladimir-Petkov: okay, I will change it.

services/121-service/src/events/events.service.ts
```diff
@@ -254,25 +270,41 @@ export class EventsService {
...
+              Object.values(oldEntity[key])[0]
+            : null;
+
+      const newValue =
+        typeof newEntity[key] === 'string' || typeof newEntity[key] === 'number'
+          ? newEntity[key]
+          : newEntity[key] && typeof newEntity[key] === 'object'
+            ? (newEntity[key] as LocalizedString)?.en ||
+              Object.values(newEntity[key])[0]
+            : null;
```
elwinschmitz, line 302: I think this could use some more readable (and less repetitive) code...
Using a helper-function to write out the 'algorithm' to get to the final value.

Something like: (⚠️ not tested locally! And the exact typings used need a sanity-check)

```suggestion
      function getLocalizedStringOrFallback(value: LocalizedString | {}) {
        return (value as LocalizedString)?.en || Object.values(value)[0];
      }
      
      function getValue(value: unknown): string | number | null {
        if (typeof value === 'string' || typeof value === 'number') {
          return value;
        }
        
        if (isObject(value)) {
          return getLocalizedStringOrFallback(value);
        }
            
        return null;
      }
      
      const oldValue = getValue(oldEntity[key]);
      const newValue = getValue(newEntity[key]);
```

The helper-methods/functions could also be put somewhere more central/useful.. Probably not the only place where we do such tasks/magic. ;)

services/121-service/src/events/events.service.ts
```diff
@@ -343,9 +378,11 @@ export class EventsService {
       'registrationProgramId',
       'personAffectedSequence',
       'lastMessageStatus',
-      'name',
       'inclusionScore',
+      // Add 'name' only if it's part of LogEntity; otherwise, remove it
```
elwinschmitz, line 382: I think this comment in-code is actually a PR-comment/question...

So _is_ the `name`-property part of `LogEntity`? Then it should be added in the array above.

Is it _not_ part of it? Then this comment can also be removed, I'd think.

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -632,18 +698,25 @@ export class RegistrationsController {
...
   ): Promise<BulkActionResultDto> {
-    const userId = req.user.id;
+    const userId = req.user?.id;
+
+    if (typeof userId === 'undefined') {
+      throw new HttpException(
+        'User is not authenticated',
+        HttpStatus.UNAUTHORIZED,
+      );
+    }
```
elwinschmitz, line 712: Isn't this something that can be handled by a helper-method?
So that the code doesn't need to be copy-pasted in every method in every/this controller?

Or isn't it actually the responsibility/functionality of some Nest.js-middleware?
Or could it work similar to how the `ParsIntPipe` (see a few lines up) works? Like a `ValidUserPipe` or `AuthenticatedUserPipe`?

elwinschmitz (changes requested):
Only the "not longer necessary"-comments in the code would be 'required to change' for me...

The logic-simplifications should be tested _for real_ and maybe judged for their 'simplicity' by you. ;)

services/121-service/src/events/events.service.ts
```diff
@@ -254,25 +270,41 @@ export class EventsService {
...
+          : newEntity[key] && typeof newEntity[key] === 'object'
+            ? (newEntity[key] as LocalizedString)?.en ||
+              Object.values(newEntity[key])[0]
+            : null;
+
       const eventForChange = this.createEventForChange(
-        fieldName,
-        oldEntity[fieldName],
-        newEntity[fieldName],
+        key, // The key is now typed correctly
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/events/events.service.ts
Vladimir-Petkov: Done

services/121-service/src/events/events.service.ts
Vladimir-Petkov: Done

services/121-service/src/registration/registrations.controller.ts
Vladimir-Petkov: Done

elwinschmitz (changes requested):

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -559,22 +568,31 @@ export class MetricsService {
...
+
       for (const key in row) {
         if (row[key] != null) {
+          // Check if the value is not null
           filteredRow[key] = row[key];
         }
       }
-      return filteredRow;
+
+      return filteredRow; // Return the filtered row
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -559,22 +568,31 @@ export class MetricsService {
...
+  private filterUnusedColumn(
+    columnDetails: Record<string, unknown>[],
+  ): Record<string, unknown>[] {
     return columnDetails.map((row) => {
-      const filteredRow = {};
+      const filteredRow: Record<string, unknown> = {}; // Define filteredRow explicitly
+
       for (const key in row) {
         if (row[key] != null) {
+          // Check if the value is not null
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -790,37 +809,43 @@ export class MetricsService {
...
-    const result = allRegistrations.map((registration) => {
+    const result = allRegistrations.map((registration: RegistrationType) => {
+      // Ensure registration is of type RegistrationType
       registration =
         this.registrationsService.transformRegistrationByNamingConvention(
           JSON.parse(JSON.stringify(program.fullnameNamingConvention)),
-          registration,
-        );
+          registration as RegistrationType, // Cast the input to RegistrationType
+        ) as RegistrationType; // Ensure the return type is RegistrationType
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -790,37 +809,43 @@ export class MetricsService {
...
 
       // If a mapping exists, get the display name for the preferred language else use the FSP name
       const fspDisplayNameForRegistrationFsp =
-        fspDisplayNameMapping[registration['fsp']];
+        fspDisplayNameMapping[
+          registration['fsp'] as keyof typeof fspDisplayNameMapping // Cast to keyof typeof fspDisplayNameMapping
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -790,37 +809,43 @@ export class MetricsService {
...
+        ];
+
       registration['fsp'] = fspDisplayNameForRegistrationFsp
         ? fspDisplayNameForRegistrationFsp[preferredLanguage]
-        : registration['fsp']?.fsp;
+        : (registration['fsp'] as string); // Cast registration['fsp'] to string
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -839,23 +864,32 @@ export class MetricsService {
     program: ProgramEntity,
     fspQuestions: FspQuestionEntity[],
   ): Record<number, RegistrationDataOptions[]> {
-    const relationOptionsPerFsp = {};
+    // Explicitly type relationOptionsPerFsp as a Record with number keys
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -957,8 +991,10 @@ export class MetricsService {
       (o) => o.newUniqueName,
     );
     const result: object[] = [];
+
     for (const row of rows) {
-      const resultRow = {};
+      const resultRow: Record<string, unknown> = {}; // Define resultRow as Record<string, any>
```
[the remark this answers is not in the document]
Vladimir-Petkov: Agree.

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -699,7 +707,9 @@ export class RegistrationsController {
       'ReferenceId retrieved - NOTE: this endpoint is scoped, depending on program configuration it only returns/modifies data the logged in user has access to.',
   })
   @Get('programs/:programId/registrations/referenceid/:paId')
-  public async getReferenceId(@Param() params): Promise<any> {
+  public async getReferenceId(
+    @Param() params: { programId: number; paId: number },
+  ): Promise<any> {
```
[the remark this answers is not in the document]
Vladimir-Petkov: I fixed it.

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -790,37 +809,43 @@ export class MetricsService {
...
       const registrationsWithSameFspId =
-        await this.getRegistrationsFieldsForDuplicates(
+        (await this.getRegistrationsFieldsForDuplicates(
           program.id,
           relationOptionsPerFsp[fspId]
             ? relationOptionsPerFsp[fspId]
             : relationOptionNoFsp,
           registrationIds.map((r) => r.registrationProgramId),
-        );
+        )) as RegistrationType[]; // Use type assertion here if necessary
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -765,18 +783,19 @@ export class MetricsService {
...
       { registrationProgramId: number; fspId: number }[]
     > = {};
     registrationAndFspId.forEach((registration) => {
       if (!registration.fspId) {
         return;
       }
       const { registrationProgramId, fspId } = registration;
-      if (!groupedRegistrations[fspId]) {
-        groupedRegistrations[fspId] = [];
+      const fspIdStr = fspId.toString(); // Convert fspId to string for indexing
```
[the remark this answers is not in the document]
Vladimir-Petkov: Agree.

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -513,14 +521,15 @@ export class MetricsService {
...
   private async replaceValueWithDropdownLabel(
-    rows: object[],
+    rows: Record<string, unknown>[], // Typing rows as an array of objects with string keys
     relationOptions: RegistrationDataOptions[],
   ): Promise<void> {
     // Creates mapping list of questions with a dropdown
     const valueOptionMappings: {
       questionName: string;
-      options: QuestionOption[] | null;
+      options?: { option: string; label: Record<string, string> }[]; // Added options typing
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

services/121-service/src/metrics/metrics.service.ts
```diff
@@ -513,14 +521,15 @@ export class MetricsService {
   }
 
   private async replaceValueWithDropdownLabel(
-    rows: object[],
+    rows: Record<string, unknown>[], // Typing rows as an array of objects with string keys
```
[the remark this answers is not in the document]
Vladimir-Petkov: Done

elwinschmitz (approved):
🥳

PR #5833, opened 2024-09-17
Add E2E tests for "Project monitoring page"

Piotrk39:
[AB#30324](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30324) 

## Describe your changes

E2E tests that validates elements present on the "Monitoring" page as well as values displayed in cards for Registration and Inclusion

Out of scope:

Iframe cannot be tested for now since there are no elements with any text to be asserted. Same time isVisible() & isHidden() method always returns true on this page.

### DIFF
--- a/e2e/portalicious/pages/ProjectMonitoringPage.ts
+++ b/e2e/portalicious/pages/ProjectMonitoringPage.ts
@@ -28,25 +28,13 @@ class ProjectMonitoring extends BasePage {
   }
 
   async assertMonitoringTabElements() {
-    const registrationsTileLocator = this.peopleRegisteredTile;
-    const includedTileLocator = this.peopleIncludedTile;
-    const remainingBudgetTileLocator = this.remainingBudgetTile;
-    const cashDisbursedTileLocator = this.cashDisbursedTile;
-    const projectDescriptionTileLocator = this.projectDescriptionTile;
-
-    const registrationTileText = await registrationsTileLocator.innerText();
-    const includedTileText = await includedTileLocator.innerText();
-    const remainingBudgetTileText =
-      await remainingBudgetTileLocator.innerText();
-    const cashDisbursedTileText = await cashDisbursedTileLocator.innerText();
-    const projectDescriptionTileText =
-      await projectDescriptionTileLocator.innerText();
-
-    expect(registrationTileText).toContain('People registered');
-    expect(includedTileText).toContain('People included');
-    expect(remainingBudgetTileText).toContain('Remaining budget');
-    expect(cashDisbursedTileText).toContain('Cash disbursed');
-    expect(projectDescriptionTileText).toContain('Project description');
+    await expect(this.peopleRegisteredTile).toContainText('People registered');
+    await expect(this.peopleIncludedTile).toContainText('People included');
+    await expect(this.remainingBudgetTile).toContainText('Remaining budget');
+    await expect(this.cashDisbursedTile).toContainText('Cash disbursed');
+    await expect(this.projectDescriptionTile).toContainText(
+      'Project description',
+    );
 
     // For the moment we are not checking the iframe area because there are no visible elemnts in it
   }
@@ -65,11 +53,10 @@ class ProjectMonitoring extends BasePage {
       'metric-tile-component',
     );
 
-    const registrationTileText = await registrationsTileLocator.innerText();
-    const includedTileText = await includedTileLocator.innerText();
-
-    expect(registrationTileText).toContain(peopleRegistered.toString());
-    expect(includedTileText).toContain(peopleIncluded.toString());
+    await expect(registrationsTileLocator).toHaveText(
+      peopleRegistered.toString(),
+    );
+    await expect(includedTileLocator).toHaveText(peopleIncluded.toString());
   }
 }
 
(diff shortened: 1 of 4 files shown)


e2e/portalicious/pages/ProjectMonitoringPage.ts
```diff
@@ -0,0 +1,76 @@
...
+    const projectDescriptionTileText =
+      await projectDescriptionTileLocator.innerText();
+
+    expect(registrationTileText).toContain('People registered');
+    expect(includedTileText).toContain('People included');
+    expect(remainingBudgetTileText).toContain('Remaining budget');
+    expect(cashDisbursedTileText).toContain('Cash disbursed');
+    expect(projectDescriptionTileText).toContain('Project description');
+
+    // For the moment we are not checking the iframe area because there are no visible elemnts in it
```
elwinschmitz, line 51: We could check "if its there or not", right?

Because for some programs/projects the URL would be configured, so then the `<iframe>` should be rendered.
And otherwise, if the property is missing, the "no iframe here, contact us"-text should be used.

But it could be a test-scenario for another day... ;)

e2e/portalicious/pages/ProjectMonitoringPage.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  async asserValuesInMonitoringTab({
```
elwinschmitz, line 54:
```suggestion
  async assertValuesInMonitoringTab({
```

Tiny typo.

e2e/portalicious/tests/ProjectMonitoring/AllElementsOfProjectMonitoringPageDisplayCorrectData.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+  const projectTitle = 'NLRC OCW program';
+
+  await test.step('Navigate to project`s monitoring page', async () => {
+    await basePage.selectProgram(projectTitle);
+    await projectMonitoring.navigateToProgramPage('Monitoring');
+  });
+
+  await test.step('Check if all elements are displayed', async () => {
+    await projectMonitoring.assertMonitoringTabElements();
+    await projectMonitoring.asserValuesInMonitoringTab({
```
elwinschmitz, line 43:
```suggestion
    await projectMonitoring.assertValuesInMonitoringTab({
```
Tiny typo.

e2e/portalicious/pages/ProjectMonitoringPage.ts
```diff
@@ -0,0 +1,76 @@
...
+    );
+    const includedTileLocator = this.peopleIncludedTile.getByTestId(
+      'metric-tile-component',
+    );
+
+    const registrationTileText = await registrationsTileLocator.innerText();
+    const includedTileText = await includedTileLocator.innerText();
+
+    expect(registrationTileText).toContain(peopleRegistered.toString());
+    expect(includedTileText).toContain(peopleIncluded.toString());
```
elwinschmitz, line 72: Following the warning on: https://playwright.dev/docs/api/class-locator#locator-inner-text

```suggestion
    expect(registrationsTileLocator).toHaveText(peopleRegistered.toString());
    expect(includedTileLocator).toHaveText(peopleIncluded.toString());
```

Wouldn't this be better (and shorter!)?

e2e/portalicious/pages/ProjectMonitoringPage.ts
```diff
@@ -0,0 +1,76 @@
...
+      await remainingBudgetTileLocator.innerText();
+    const cashDisbursedTileText = await cashDisbursedTileLocator.innerText();
+    const projectDescriptionTileText =
+      await projectDescriptionTileLocator.innerText();
+
+    expect(registrationTileText).toContain('People registered');
+    expect(includedTileText).toContain('People included');
+    expect(remainingBudgetTileText).toContain('Remaining budget');
+    expect(cashDisbursedTileText).toContain('Cash disbursed');
+    expect(projectDescriptionTileText).toContain('Project description');
```
elwinschmitz, line 49: For all these, also see comment below about using `locator.toHaveText()`...

elwinschmitz (changes requested):

elwinschmitz (changes requested):
If the suggested use of `.toHaveText` doesn't help or doesn't "do what I think it does"; The rest looks fine. :)

e2e/portalicious/pages/ProjectMonitoringPage.ts
Piotrk39: Done!

e2e/portalicious/pages/ProjectMonitoringPage.ts
Piotrk39: Yes the only issue I had was the iframe really I can take a look again into this. All the other comments should be resolved. In this case I changed "toHeaveText" into "toIncludeText" because it just checks for for one string.

aberonni (approved):
🚀

e2e/portalicious/pages/ProjectMonitoringPage.ts
aberonni: @Piotrk39 I was about to dismiss Elwin's review but I actually think it would be nice to add this.

I assume it would have to be something as simple as (pseudocode) `expect(page).toContainElement('iframe')` which should be true for one program and false for another.

aberonni: Moved to https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/30569

PR #5860, opened 2024-09-26
Validate fsp config and required attributes for Registration data refactor

RubenGeo:
[AB#30474](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30474) 


services/121-service/src/registration/processsors/registrations-update.processor.ts
```diff
@@ -23,10 +23,11 @@ export class RegistrationUpdateProcessor {
       data: jobData.data,
       reason: jobData.reason,
     };
-    await this.registrationsService.updateRegistration(
-      jobData.programId,
-      jobData.referenceId,
-      dto,
-    );
+    await this.registrationsService.validateBodyAndUpdateRegistration({
```
diderikvw, line 26: What is a "Body"?

services/121-service/src/registration/processsors/registrations-update.processor.ts
```diff
@@ -23,10 +23,11 @@ export class RegistrationUpdateProcessor {
...
+      programId: jobData.programId,
+      referenceId: jobData.referenceId,
+      updateRegistrationDto: dto,
+      userId: jobData?.request?.userId,
```
diderikvw, line 30: We can omit the ?? Looks like we know the userId is always there.

services/121-service/src/registration/registrations.service.ts
```diff
@@ -362,21 +323,22 @@ export class RegistrationsService {
...
 
-  public async importRegistrations(
+  public async importRegistrationsFromCsv(
     csvFile: Express.Multer.File,
     programId: number,
     userId: number,
   ): Promise<ImportResult> {
     const program = await this.findProgramOrThrow(programId);
-    return await this.registrationsImportService.importRegistrations(
+    this.throwIfProgramIsNotPublished(program.published);
```
diderikvw, line 332: AFAIK we are removing the published attribute, so maybe no need to introduce a check for it anymore.

services/121-service/src/registration/registrations.service.ts
```diff
@@ -389,28 +351,26 @@ export class RegistrationsService {
...
-  public async importRegistrationFromJson(
-    validatedJsonData: ImportRegistrationsDto[],
+  public async importRegistrationsFromJson(
+    jsonData: Record<string, string | number | boolean>[],
     programId: number,
     userId: number,
   ): Promise<ImportResult> {
     const program = await this.findProgramOrThrow(programId);
-    if (!program?.published) {
+    this.throwIfProgramIsNotPublished(program.published);
```
diderikvw, line 360: See other comment.

services/121-service/src/registration/registrations.service.ts
```diff
@@ -389,28 +351,26 @@ export class RegistrationsService {
...
+    return await this.registrationsImportService.importRegistrations(
```
diderikvw, line 361: Do we not want to check first for the correct contents of the json before calling this? As the old code did check (?) based on the Dto?

services/121-service/src/registration/services/registrations-import.service.ts
```diff
@@ -67,31 +65,21 @@ export class RegistrationsImportService {
...
+    await this.validateBulkUpdateInput(bulkUpdateRecords, programId, userId);
 
     // Prepare the job array to push to the queue
-    const updateJobs: RegistrationUpdateJobDto[] = filteredRegistrations.map(
-      (registration) => {
-        const updateData = { ...registration };
-        delete updateData['referenceId'];
+    const updateJobs: RegistrationUpdateJobDto[] = bulkUpdateRecords.map(
+      (record) => {
+        // const updateData = { ...registration, ...registration.data };
```
diderikvw, line 77: Remove this comment?

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw, line 1: Why is the Safaricom code changed in this item?

services/121-service/test/payment/do-payment-safaricom.test.ts
diderikvw, line 1: Why was Safaricom code changed in this item?

diderikvw (changes requested):
See comments. For a more in-depth review of the changed code, I would need to learn a lot more of the code where the changes take place. I would like to, but may be out of scope for this item.

services/121-service/package.json
```diff
@@ -54,6 +54,7 @@
     "@nestjs/typeorm": "^10.0.2",
     "@types/express": "^4.17.21",
     "@types/jsonwebtoken": "^9.0.7",
+    "@types/multer": "^1.4.12",
```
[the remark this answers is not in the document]
RubenGeo: This is just the typing of the package we are already using. So it basically make the package we are using more strongly typed

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
RubenGeo: We removed the checks that checks if the required FSP attributes (like nationalId/phonenumber) are set for each registration before making the api request to the FSP. The reason for that is that it should be impossible for them not to be set as this is now checked when creating registrations. So this check was redundant These settings are stored in src/financial-service-providers/financial-service-providers.const.ts
image

services/121-service/test/payment/do-payment-safaricom.test.ts
RubenGeo: With the new code introduced in this PR that checks if registrations all have the required attributes it was not possible to remove the phonenumber of a registration. So the test had to be adjusted

services/121-service/src/registration/registrations.service.ts
RubenGeo: not completely sure what you mean. The validation of JSON of the new registrations need the programId to be able to validate the required program registration attributes of a program. So better to first validate if the program stuff is correct, also these program checks are less expensive

services/121-service/src/registration/registrations.service.ts
RubenGeo: [See](https://github.com/global-121/121-platform/pull/5860#discussion_r1818959001)

services/121-service/src/registration/interfaces/validated-registration-input.interface.ts
```diff
@@ -0,0 +1,23 @@
...
+    | 'phoneNumber'
+    | 'preferredLanguage'
+    | 'inclusionScore'
+    | 'paymentAmountMultiplier'
+    | 'maxPayments'
+    | 'scope'
+  >
+>;
+
+export interface ValidatedRegistrationInput
```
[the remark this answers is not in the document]
RubenGeo: > Looks like this interface is mostly used as return type. So define according to guidelines.
It's also used twice as input of a function. I don't think the name 'ValidatedRegistrationInputReturn' will be easier to understand tbh, so leaving it like this

> The most important code in this file is this exported interface, so put that at the top above the type def. See guidelines.
I moved the interface to the top.

services/121-service/src/registration/dto/validate-registration-error-object.dto.ts
```diff
@@ -1,6 +1,6 @@
-export class ValidateRegistrationErrorObjectDto {
-  public lineNumber: number;
-  public column: string;
-  public value: string;
-  public error: string;
+export interface ValidateRegistrationErrorObject {
```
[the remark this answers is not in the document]
RubenGeo: > It looks like this interface is used as a return type for functions only. That means the file should be in the /interfaces folder, not the /dto folder. And the name of the interface should be in principle: FunctionNameReturnType. I think then: "ValidateAndCleanInputReturnType", as that seems the only public function. 

The public function `validateAndCleanInput` does not return the ValidateRegistrationErrorObject so I don't think this is appropriate in this case. This error object is only used in private function in the RegistrationsInputValidator class. Maybe it should live inside that file than instead of having a seperate interface file so it cannot be used in other files. Not sure what our practice is here. 

> All attributes of the interfaces should be defined readonly. This was in the guidelines only for class DTOs, but also goes for interfaces. I updated the guidelines.
(y)

services/121-service/src/payments/fsp-integration/excel/excel.service.ts
```diff
@@ -382,7 +382,7 @@ export class ExcelService
     );
   }
 
-  private joinRegistrationsAndImportRecords(
+  public joinRegistrationsAndImportRecords(
```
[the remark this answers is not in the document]
RubenGeo: Yea, I agree. However I changed this in one of the other refactor registration data branches to private and in this branch I realized that our unit tests now fail. So reverted it back to public for now. Leaving the refactoring of this unit test out of scope for this work.

services/121-service/src/registration/modules/queue-registrations-update/queue-registrations-update.service.ts
```diff
@@ -21,9 +21,16 @@ export class QueueRegistrationUpdateService {
   public async addRegistrationUpdateToQueue(
```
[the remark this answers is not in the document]
RubenGeo: > I am starting to suspect we have some kind of "Registration Update Queue". This is the first time I learn about this, and I do not see it reflected in our Queues diagram. 

We have one. It's used on bulk updating registrations. Functionality that only Sanne and Tijs use atm. 

No other queues that I know of need to be added. I created this [task](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20140?workitem=30968) to add it to our docs

services/121-service/src/registration/processsors/registrations-update.processor.ts
RubenGeo: Changed it to input

services/121-service/src/registration/registrations.service.ts
RubenGeo: Ah okay nice 👍. But now it's easier to remove right, because the error is not thrown in 2 places

services/121-service/src/registration/dto/validate-registration-error-object.dto.ts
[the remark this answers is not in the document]
diderikvw: > The public function validateAndCleanInput does not return the ValidateRegistrationErrorObject so I don't think this is appropriate in this case.

You are right, I mislooked.

> This error object is only used in private functions in the RegistrationsInputValidator class.

This code smells, but refactoring is out of scope.

> What do you think?

I personally like to stick to "all interfaces in their own file". That makes it simple to follow.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
diderikvw: If we want to go this way, then IMO these attributes should not be optional anymore in SafaricomTransactionJobDto.

diderikvw (changes requested):
See comment.

services/121-service/src/transaction-job-processors/transaction-job-processors.service.ts
RubenGeo: > Anyway, of we go this route, then these attributes should not be optional in the job DTO anymore.

Good point! I changed this

jannisvisser (approved):

PR #5866, opened 2024-09-27
fix: filter on referenceId consisting of numbers

PeterSmallenbroek:
[AB#30520](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30520) 

### DIFF
--- a/services/121-service/src/registration/services/registrations-pagination.service.ts
+++ b/services/121-service/src/registration/services/registrations-pagination.service.ts
@@ -164,6 +164,13 @@ export class RegistrationsPaginationService {
 
     queryBuilder = this.addPaymentFilter(queryBuilder, query);
 
+    if (query?.filter?.referenceId) {
+      queryBuilder.andWhere('CAST("referenceId" AS TEXT) = :referenceId', {
+        referenceId: query.filter.referenceId,
+      });
+      delete query.filter.referenceId;
+    }
+
     // PaginateConfig.select and PaginateConfig.relations cannot be used in combi with each other
     // That's why we wrote some manual code to do the selection
     const result = await paginate<RegistrationViewEntity>(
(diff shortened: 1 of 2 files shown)


services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -163,6 +163,13 @@ export class RegistrationsPaginationService {
 
     queryBuilder = this.addPaymentFilter(queryBuilder, query);
 
+    if (query?.filter?.referenceId?.includes('$ilike') === false) {
```
aberonni, line 166:
```suggestion
    if (query?.filter?.referenceId && !query?.filter?.referenceId?.includes('$ilike')) {
```

Should it rather be something like this?

PeterSmallenbroek: It might, what is the improvement here?

aberonni: Two different things:

```ts
    if (!query?.filter?.referenceId?.includes('$ilike')) {
```

^ This is how I would've written your check (`=== true` or `=== false` are generally avoidable)

But also, I believe that your `if` would be true even if `query?.filter?.referenceId = null` but I might be wrong.

PeterSmallenbroek: > if (!query?.filter?.referenceId?.includes('$ilike')) {

This does not achieve the same as it would also 'respond' to undefined/null:
image


> I believe that your if would be true even if query?.filter?.referenceId = null but I might be wrong

I wondered the same so I asked Chatgippiti:
image

aberonni (approved):
I don't know if my approval is sufficient because I don't really understand what the code is doing, but the syntax looks good to me :D

elwinschmitz (approved):

PR #5902, opened 2024-10-04
portalicious: add/edit user form AB#30597 AB#30598

arsforza:
[AB#30597](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30597) [AB#30598](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30598)

### DIFF
--- a/interfaces/Portalicious/src/app/domains/user/user.api.service.ts
+++ b/interfaces/Portalicious/src/app/domains/user/user.api.service.ts
@@ -63,6 +63,34 @@ export class UserApiService extends DomainApiService {
   getAllUsers() {
     return this.generateQueryOptions<User[]>({
       path: [BASE_ENDPOINT],
+      queryKey: [BASE_ENDPOINT],
+    });
+  }
+
+  createUser({
+    username,
+    displayName,
+  }: {
+    username: string;
+    displayName: string;
+  }) {
+    return this.httpWrapperService.perform121ServiceRequest<User>({
+      method: 'POST',
+      endpoint: BASE_ENDPOINT,
+      body: {
+        users: [
+          {
+            username,
+            displayName,
+          },
+        ],
+      },
+    });
+  }
+
+  public invalidateCache(): Promise<void> {
+    return this.queryClient.invalidateQueries({
+      queryKey: [BASE_ENDPOINT],
     });
   }
 }
--- a/interfaces/Portalicious/src/app/domains/user/user.api.service.ts
+++ b/interfaces/Portalicious/src/app/domains/user/user.api.service.ts
@@ -88,6 +88,25 @@ export class UserApiService extends DomainApiService {
     });
   }
 
+  udpateUserDisplayName({
+    id,
+    displayName,
+  }: {
+    id: number | undefined;
+    displayName: string;
+  }) {
+    if (!id) {
+      return Promise.reject(new Error('User ID is required'));
+    }
+    return this.httpWrapperService.perform121ServiceRequest<User>({
+      method: 'PATCH',
+      endpoint: `${BASE_ENDPOINT}/${id.toString()}`,
+      body: {
+        displayName,
+      },
+    });
+  }
+
   public invalidateCache(): Promise<void> {
     return this.queryClient.invalidateQueries({
       queryKey: [BASE_ENDPOINT],
--- a/interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
+++ b/interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
@@ -1,14 +1,14 @@
 <app-form-sidebar
   [(visible)]="formVisible"
   [formGroup]="formGroup"
-  [mutation]="addUserMutation"
+  [mutation]="userMutation"
   formTitle="Add user"
-  i18n-formTitle="@@add-user"
+  i18n-formTitle
 >
   <app-form-field-wrapper
     label="User name"
     i18n-label
-    [errorMessage]="formFieldErrors()('usernameValue')"
+    [errorMessage]="formFieldErrors()('displayNameValue')"
   >
     <input
       pInputText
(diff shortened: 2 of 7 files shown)


interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  formGroup = new FormGroup({
+    usernameValue: new FormControl<string>('', {
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      validators: [Validators.required],
+      nonNullable: true,
+    }),
+    emailValue: new FormControl<string>('', {
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      validators: [Validators.required, Validators.email],
```
aberonni, line 98: @arsforza RE: your comment:

> the email validator here on line 98 is never triggered because of the required validator
> should we remove the latter?
> I think the form would be invalid with an empty email field anyway

I don't see why it should not triggered. On the login page we have the same two validators, and they work correctly:

image

arsforza: it's because it's a custom validation message on control.invalid for email
I added the same

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    }: {
+      usernameValue: string;
+      emailValue: string;
+    }) => {
+      if (this.isEditing() && this.userToEdit()?.id) {
+        return this.userApiService.udpateUserDisplayName({
+          id: this.userToEdit()?.id,
+          displayName: usernameValue,
+        });
+      }
```
aberonni, line 125:
```suggestion
      if (this.isEditing()) {
        return this.userApiService.udpateUserDisplayName({
          id: this.userToEdit()?.id,
          displayName: usernameValue,
        });
      }
```

We don't need the extra check on the id, because it's already happening in udpateUserDisplayName. As-is, there is an unlikely risk that we create user when we mean to edit it.

interfaces/Portalicious/src/app/domains/user/user.api.service.ts
```diff
@@ -63,6 +63,53 @@ export class UserApiService extends DomainApiService {
...
+      return Promise.reject(new Error('User ID is required'));
+    }
+    return this.httpWrapperService.perform121ServiceRequest<User>({
+      method: 'PATCH',
+      endpoint: `${BASE_ENDPOINT}/${id.toString()}`,
+      body: {
+        displayName,
+      },
+    });
+  }
```
aberonni, line 108: Probably unnecessary, and I'll leave it to you whether to change this or not, but I feel like this might be more futureproof.

```suggestion

  udpateUser({
    id,
    patchedUser,
  }: {
    id: number | undefined;
    patchedUser: Partial<User>;
  }) {
    if (!id) {
      return Promise.reject(new Error('User ID is required'));
    }
    return this.httpWrapperService.perform121ServiceRequest<User>({
      method: 'PATCH',
      endpoint: `${BASE_ENDPOINT}/${id.toString()}`,
      body: patchedUser,
    });
  }
```

aberonni (changes requested):
Looks good! I'm very happy that all the changes I'm requesting are just details :)

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
```diff
@@ -0,0 +1,34 @@
...
+    label="User name"
+    i18n-label
+    [errorMessage]="formFieldErrors()('displayNameValue')"
+  >
+    <input
+      pInputText
+      type="text"
+      formControlName="usernameValue"
+      autocomplete="off"
+      class="font-mono"
```
aberonni, line 18:
```suggestion
```

Sorry I didn't notice this before. I don't think this should be mono (according to the designs at least).

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
```diff
@@ -0,0 +1,34 @@
...
+    label="User email"
+    i18n-label
+    [errorMessage]="formFieldErrors()('usernameValue')"
+  >
+    <input
+      pInputText
+      type="email"
+      formControlName="emailValue"
+      autocomplete="off"
+      class="font-mono"
```
aberonni, line 31:
```suggestion
```

Same as above

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
```diff
@@ -0,0 +1,34 @@
...
+>
+  <app-form-field-wrapper
+    label="User name"
+    i18n-label
+    [errorMessage]="formFieldErrors()('displayNameValue')"
+  >
+    <input
+      pInputText
+      type="text"
+      formControlName="usernameValue"
```
aberonni, line 16:
```suggestion
      formControlName="displayNameValue"
```

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
```diff
@@ -0,0 +1,34 @@
...
+  </app-form-field-wrapper>
+  <app-form-field-wrapper
+    label="User email"
+    i18n-label
+    [errorMessage]="formFieldErrors()('usernameValue')"
+  >
+    <input
+      pInputText
+      type="email"
+      formControlName="emailValue"
```
aberonni, line 29:
```suggestion
      formControlName="usernameValue"
```

interfaces/Portalicious/src/app/pages/users/add-user-form/add-user-form.component.html
aberonni, line 1: There are some mismatch in the form control names that might be a consequence of my review comments. Make sure to test all the functionality here because it looks like it would be broken.

aberonni (changes requested):

aberonni (approved):
LGTM! 🚀

PR #5911, opened 2024-10-09
feat.activities: dto mapping to endpoint

aberonni:
[AB#30664](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30664)

### DIFF
--- a/services/121-service/src/activities/dtos/activities.dto.ts
+++ b/services/121-service/src/activities/dtos/activities.dto.ts
@@ -1,18 +1,20 @@
+import { FinancialServiceProviderName } from '@121-service/src/financial-service-providers/enum/financial-service-provider-name.enum';
 import { MessageContentType } from '@121-service/src/notifications/enum/message-type.enum';
-import { NotificationType } from '@121-service/src/notifications/twilio.entity';
 import { TransactionStatusEnum } from '@121-service/src/payments/transactions/enums/transaction-status.enum';
 import { RegistrationStatusEnum } from '@121-service/src/registration/enum/registration-status.enum';
+import { LocalizedString } from '@121-service/src/shared/types/localized-string.type';
 
 export class ActivitiesDto {
   meta: {
     availableTypes: ActivityTypeEnum[];
     count: Partial<Record<ActivityTypeEnum, number>>;
   };
-  data: ActivityType[];
+  data: Activity[];
 }
 
-type ActivityType =
+export type Activity =
   | DataChangeActivity
+  | FinancialServiceProviderChangeActivity
   | MessageActivity
   | NoteActivity
   | StatusChangeActivity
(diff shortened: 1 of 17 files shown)


services/121-service/src/activities/dtos/activities.dto.ts
```diff
@@ -1,92 +1,98 @@
...
+  FinancialServiceProviderChange = 'financial-service-provider-change',
 }
 
 interface BaseActivity {
   id: string;
-  user: {
-    id: number;
-    username: string;
-  };
+  username?: string;
```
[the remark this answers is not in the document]
aberonni: Because of the way we've built the tables in the FE with primeng, the filters only work if they are on the top-level of an object (ie. not nested within an object within an object). This might become an issue in the future, but for the time being this was an easy fix in the BE, rather than having to do data manipulation in the frontend. But if I'm deviating from a convention, I'm happy to revert and add an extra step in the FE.

services/121-service/src/activities/interfaces/data-change-activity.interface.ts
```diff
@@ -0,0 +1,12 @@
+import { ActivityTypeEnum } from '@121-service/src/activities/enum/activity-type.enum';
+import { BaseActivity } from '@121-service/src/activities/interfaces/base-activity.interface';
+
+export interface DataChangeActivity extends BaseActivity {
+  type: ActivityTypeEnum.DataChange;
+  attributes: {
+    fieldName: string | null;
```
diderikvw, line 7: What is the necessity of having the fields to be | null? Not having this would make using the interface simpler.

services/121-service/src/activities/interfaces/financial-service-provider.interface.ts
```diff
@@ -0,0 +1,11 @@
+import { ActivityTypeEnum } from '@121-service/src/activities/enum/activity-type.enum';
+import { BaseActivity } from '@121-service/src/activities/interfaces/base-activity.interface';
+
+export interface FinancialServiceProviderChangeActivity extends BaseActivity {
+  type: ActivityTypeEnum.FinancialServiceProviderChange;
+  attributes: {
+    oldValue: string | null;
```
diderikvw, line 7: What is the necessity of having the fields to be | null? Not having this would make using the interface simpler.

services/121-service/src/activities/dtos/activities.dto.ts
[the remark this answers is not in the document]
diderikvw: Yes, this is deviating from a software architecture convention I am trying to get us to adhere to.
Unless that is not a good idea at all from API consumer perspective. If that is the case, then let's have a conversation. Otherwise, please revert and add the extra step. Assuming that is not very complex or a lot of work.

diderikvw (changes requested):
Ok, great you addressed these.
Just a few small things left. 

One of which here (because I cannot reply there for some reason):
> I've tried to align with the guidelines to have one export per file, but I think this is a great example of why I didn't like this guideline to begin with (as per our old discussions about it).

I am happy to have a conversation about changing this convention. But I do want to have some convention on this and not let it depend on "whatever the developer favors who happens to work on this code". IMO having it "all the same way" does improve readability, even if TS code-wise it is not as efficient as could be.

services/121-service/src/activities/interfaces/financial-service-provider.interface.ts
aberonni: This stems from the input data being nullable:
https://github.com/global-121/121-platform/blob/c431e19a01f8144121a8592c91a4929c62be5920/services/121-service/src/events/dto/get-event.dto.ts#L7

services/121-service/src/activities/interfaces/data-change-activity.interface.ts
diderikvw: Ok, I would like to discuss this with @PeterSmallenbroek , related to the ##TODOs already in the code.

diderikvw (approved):

PR #5912, opened 2024-10-10
fix: referenceId mistakenly cast to number

aberonni:
[AB#30719](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30719)

Regression introduced in nestjs-paginate:
https://github.com/ppetzold/nestjs-paginate/commit/25739bda711c98f7de673a049da1cdf55c8a96f6#diff-87404c91a22dbabcc334f9a5288341281cb2b6625533e1207efcac16dc1e2ffdR279

This is a temporary fix to solve the regression introduced above. TBD whether we need to adjust this code, adjust our code further, or submit a PR to nestjs-paginate.

### DIFF
--- a/services/121-service/src/registration/services/registrations-pagination.service.ts
+++ b/services/121-service/src/registration/services/registrations-pagination.service.ts
@@ -164,14 +164,28 @@ export class RegistrationsPaginationService {
 
     queryBuilder = this.addPaymentFilter(queryBuilder, query);
 
-    // Replacing the filter on referenceId is needed in specific cases where the referenceId only consists of numbers.
-    // The !query.filter.referenceId.includes('$') is needed to check if the query doesn't contain an operator like '$ilike'.
-    // If the filter has a '$' we don't need to replace the filter
-    if (query?.filter?.referenceId && !query.filter.referenceId.includes('$')) {
-      queryBuilder.andWhere('CAST("referenceId" AS TEXT) = :referenceId', {
-        referenceId: query.filter.referenceId,
-      });
-      delete query.filter.referenceId;
+    // Replacing the filter on referenceId is needed in specific cases where the referenceId only consists of numbers
+    // or a string that javascript could mistakenly interpret as a number (ie. "651581942751358e5")
+    if (
+      query?.filter?.referenceId &&
+      typeof query.filter.referenceId === 'string'
+    ) {
+      if (!query.filter.referenceId.includes('$')) {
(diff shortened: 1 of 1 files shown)


elwinschmitz (approved):
As I have "seen it work", I can approve....

But I don't see a clear path to "refactor this away" very easily... :/

Maybe we could put some _real_ restrictions on the format/syntax of Reference-IDs;
Or put some magical "escapeReferenceIdString"-helpers in place....

But besides that, I don't understand enough of the implications of the things happening inside the QueryBuilder and the related real-SQL output/effect of it...

services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -164,14 +164,28 @@ export class RegistrationsPaginationService {
...
+      typeof query.filter.referenceId === 'string'
+    ) {
+      if (!query.filter.referenceId.includes('$')) {
+        // The !query.filter.referenceId.includes('$') is needed to check if the query doesn't contain an operator like '$ilike'.
+        // If the filter has a '$' we don't need to replace the filter
+        queryBuilder.andWhere('CAST("referenceId" AS TEXT) = :referenceId', {
+          referenceId: query.filter.referenceId,
+        });
+        delete query.filter.referenceId;
+      } else if (query.filter.referenceId.includes('$eq')) {
```
elwinschmitz, line 180:
```suggestion
      } else if (query.filter.referenceId.includes('$eq:')) {
```

Should we align this with the `split`-operation?
To prevent 'false positives' to fall between the cracks?

aberonni: lol too late, but I doubt it will be an issue

elwinschmitz: Yeah...
Lets calculate the changes of a randomly generated reference-ID with `$eq` somewhere in it...

elwinschmitz: Well actually... we're talking about "sanitizing user-input", so its not a matter of chance...
But a matter of _risk_.

What would a user be able to trigger when searching with a filter like: `"?referenceId=abc$eq:sudo-get-me-all-the-money();"`... 🤔

aberonni: I'd say leave a comment in the "Evaluate" task with this and your other reflections. I don't wanna invest time into a perfect hack - I'd rather invest time in a better long-term solution.

PR #5916, opened 2024-10-11
chore: refactor twilio code to not rely on "any"

aberonni:
[AB#30733](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30733)

I've refactored all of the code that is somehow involved with the Twilio client and did not take advantage of the typings we get from it out of the box.

The problem is that I don't know how to test any of this code, nor is any of it covered by integration tests. I would happily add integration tests where applicable, but I'm not sure where to start.

### DIFF
--- a/services/121-service/src/notifications/README.md
+++ b/services/121-service/src/notifications/README.md
@@ -1,0 +1,46 @@
+## Testing Twilio API during development
+
+### 1. Use a tool to inspect the responses from the Twilio API.
+
+- Some options:
+  - `ngrok`: <https://ngrok.com>:
+    - See also: <https://www.twilio.com/en-us/blog/6-awesome-reasons-to-use-ngrok-when-testing-webhooks-html>
+    - Make sure to use the correct port(`3000`) of the 121-service.
+  - `Smee`: <https://smee.io/>
+    - You can use the client with:  
+      `npx smee -u https://smee.io/<unique-url>`
+  - Or any other service that gives a public accessible URL to inspect and/or forward to you local instance of the 121-service.
+- Setup steps with `ngrok`:
+  - Install ngrok on your machine, and link to a free public account
+  - Run `ngrok http 3000`
+- Make note of your publicly accessible URL, for example `https://srkgjh-33-444-88-111.ngrok-free.app`
+
+### 2. Login to the [Twilio console](https://console.twilio.com/us1/develop/sms/try-it-out/whatsapp-learn?frameUrl=%2Fconsole%2Fsms%2Fwhatsapp%2Flearn%3Fx-target-region%3Dus1) and link yourself to the sandbox
+
+- Make sure to choose the "development" subaccount from the dropdown menu at the top of the page
+- Make note of the `SID` and `AUTHTOKEN`.
+  - Refer to these as `TWILIO_SID` and `TWILIO_AUTHTOKEN` in the `.env` later
+- From the left menu: go to Messaging > Services and copy the `SID`
+  - Refer to this as `TWILIO_MESSAGING_SID` in the `.env` later
+- Access the "Sandbox" environment from the left menu: Messaging > Try it out > Send a WhatsApp Message
+- Make note of the `WHATSAPP_NUMBER` and then join the conversation with your personal WhatsApp account
+  - Refer to this as `TWILIO_WHATSAPP_NUMBER` in the `.env` later
+- In "Sandbox settings", update the callback URLs to use your publicle accessible URL
+  - For example, `https://srkgjh-33-444-88-111.ngrok-free.app/api/notifications/whatsapp/incoming`
+
+### 3. Update your `services/.env` file
+
+- Set `TWILIO_SID`, `TWILIO_AUTHTOKEN`, `TWILIO_MESSAGING_SID`, and `TWILIO_WHATSAPP_NUMBER` to the information gathered above
+- Set `MOCK_TWILIO=` (empty string)
+- Set `EXTERNAL_121_SERVICE_URL` to the public url generated for example by `ngrok`
+  - eg. `EXTERNAL_121_SERVICE_URL=https://srkgjh-33-444-88-111.ngrok-free.app/`
+
+### 4. Restart the 121-service and test
+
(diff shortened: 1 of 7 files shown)


services/121-service/src/notifications/whatsapp/whatsapp.service.ts
```diff
@@ -73,51 +74,74 @@ export class WhatsappService {
       payload['messageContentType'] = messageContentType;
     }
 
-    let errorOccurred = false;
-    let messageToStore;
     try {
-      messageToStore = await twilioClient.messages.create(payload);
+      const messageToStore = await twilioClient.messages.create(payload);
+      await this.storeSendWhatsapp({
```
aberonni, line 79: The fact that this call was not there before my changes seems like a bug. The code in `finally` seems to assume that the `finally` clause is always called - but I believe that isn't true when there is a `return` statement in the `try` block.

services/121-service/src/notifications/auth.middlewareTwilio.ts
```diff
@@ -12,20 +12,34 @@ export class AuthMiddlewareTwilio implements NestMiddleware {
...
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+
     const validWhatsAppStatus = twilio.validateRequest(
       process.env.TWILIO_AUTHTOKEN,
       twilioSignature,
       EXTERNAL_API.whatsAppStatus,
       req.body,
-      {
```
aberonni, line 26: `twilio.validateRequest` only accepts 4 parameters, so it seems that these extra parameters were ignored. TS was not complaining before, because `twilio` was typed as `any`, therefore we had no insight into what these functions expect.

Maybe a previous version of the twilio library expected this extra info.

services/121-service/src/notifications/auth.middlewareTwilio.ts
aberonni, line 1: I don't know how to test the changes in this file

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
aberonni: I actually tested this out for myself. So, the code from before seems to work - but by chance. With our mix of TS strictness and code, it could've easily broken if we ever added a `return` to the `finally` block (which, normally, TS would complain about).

Here is a [typescript playground](https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsL1RYAbPMAdVWQAsAlAU2QQCcIAKAShgG8AoGbmZOsdlx7ComEOjxUAdHnQBzRgHI4OfIRLlqtBr34BCRcwDcQ4dzo16WAIwnhAXxhRgyKGRiMqrTmZ6iI4pIy8koquATEpJSWOs6uZAbGpmb+gdKyCl52PI5hagI+vk5iEukhyqoRGtHaWHkRidm+FrUwAAxNMPYc3RwA9H0wAHJUVAAmvOgwwGMTwLwgALSochDoFjBkVBsARlTOCCBUADQwAO5RMCB8qFDIMAC26GNUMJinZ6-OEIr3a-ctKwwKgAN22YDOWws-UGAAFkEsVmtoaBILBECgMHVKuoougEMgtFYWIJhHwCskRCUghlQjjIuR8YSYlhyY1KTxATpbMlHHE3B4vKSiqlSsEFBVwrjGQSibEXG52UVigExbSsrzsFKKcrRTTyvVpWQmXLsdqlUUBjAuVgOryehwOIaGcbZSyWFJyFQmEyAA4EmAAXgAfMLuHqyhLndUTSzFKc-QSkslo1FTR6vT6Cf77iGwylqZG6drqqb429s0nOvYkjWjEA) which illustrates this.

I would argue that my code is therefore still an improvement, in that it is much more explicit.

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
```diff
@@ -73,51 +74,74 @@ export class WhatsappService {
...
-    existingSidToUpdate?: string,
-  ): Promise<void> {
+    userId,
+    registrationId,
+    mediaUrl,
+    messageContentType,
+    messageProcessType,
+    existingSidToUpdateDueToFailure,
+  }: {
+    message: Pick<
```
RubenGeo, line 121: Should we consider changing this to a class as we use it in 2 places?

RubenGeo (changes requested):

RubenGeo:
Thanks for the refactor, I think it really improves the code :)

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
aberonni: I considered it, but then I considered that there is a potential scenario where the two functions deviate, and they need different properties from `MessageInstance`. If that isn't likely, then I think a better refactor would be to reuse the same function altogether, rather than reuse the same type. I also considered doing that, but wasn't sure where to put the function so that both services would have access to it.

services/121-service/src/notifications/README.md
```diff
@@ -0,0 +1,43 @@
...
+- Access the "Sandbox" environment from the left menu: Messaging > Try it out > Send a WhatsApp Message
+- Make note of the `WHATSAPP_NUMBER` and then join the conversation with your personal WhatsApp account
+- In "Sandbox settings", update the callback URLs to use your publicle accessible URL
+  - For example, `https://srkgjh-33-444-88-111.ngrok-free.app/api/notifications/whatsapp/incoming`
+
+### 3. Update your `services/.env` file
+
+- Set `TWILIO_SID`, `TWILIO_AUTHTOKEN`, and `TWILIO_WHATSAPP_NUMBER` to the information gathered above
+- Set `MOCK_TWILIO=` (empty string)
+- Ask a colleague for the `TWILIO_MESSAGING_SID`
```
RubenGeo, line 31: - You can go to [messaging -> services](https://console.twilio.com/us1/develop/sms/services) to find the Sid of the messaging service (`TWILIO_MESSAGING_SID`)

RubenGeo (changes requested):

RubenGeo (approved):

PR #5936, opened 2024-10-21
portalicious: fix last payment amount

tijsziere:
[AB#30842](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30842)

### DIFF
--- a/interfaces/Portalicious/src/app/pages/project-monitoring/project-monitoring.page.ts
+++ b/interfaces/Portalicious/src/app/pages/project-monitoring/project-monitoring.page.ts
@@ -96,7 +96,7 @@ export class ProjectMonitoringPageComponent {
       return;
     }
 
-    return this.payments.data()[0].payment;
+    return this.payments.data()[this.payments.data().length - 1].payment;
   });
 
   projectDescription = computed(() =>
--- a/services/121-service/src/payments/payments.service.ts
+++ b/services/121-service/src/payments/payments.service.ts
@@ -135,6 +135,7 @@ export class PaymentsService {
         programId,
       })
       .groupBy('payment')
+      .orderBy('MIN(transaction.created)', 'ASC')
       .getRawMany();
     return payments;
   }


interfaces/Portalicious/src/app/pages/project/project-monitoring/project-monitoring.page.ts
```diff
@@ -96,7 +96,7 @@ export class ProjectMonitoringPageComponent {
       return;
     }
 
-    return this.payments.data()[0].payment;
+    return this.payments.data().length;
```
aberonni, line 99: Off the top of my head, this seems like a more reliable solution (I haven't tested it)

```suggestion
    return this.payments.data()[this.payments.data().length - 1].payment;
```

aberonni (changes requested):
FYI @tijsziere I created a new task rather than re-using the previous one, and I updated the description with it. Also FYI your previous linked task in the PR description was hidden because you didn't remove the comment (lmk if this doesn't make sense).

interfaces/Portalicious/src/app/pages/project/project-monitoring/project-monitoring.page.ts
tijsziere: I had this before, wasnt working
I've refactored the endpoint to return an ordered list now

services/121-service/src/payments/payments.controller.ts
```diff
@@ -77,7 +77,8 @@ export class PaymentsController {
     }[]
   > {
     // TODO: REFACTOR: use a DTO to define stable structure of result body
-    return await this.paymentsService.getPayments(programId);
+    const payments = await this.paymentsService.getPayments(programId);
+    return payments.sort((a, b) => a.payment - b.payment);
```
tijsziere, line 81: I implemented the sorting in the BE instead of FE, but open to suggestions

interfaces/Portal/src/app/services/programs-service-api.service.ts
```diff
@@ -165,7 +165,7 @@ export class ProgramsServiceApiService {
           })
           .sort(
             (a: PaymentData, b: PaymentData) =>
-              // Sort by payment-id (as the back-end doesn't do that)
+              // Sort by payment-id (as the back-end doesn't do that) - Is this still needed now?
```
tijsziere, line 168: Saw this, don't think we need it now, but also don't think its worth the effort refactoring the old portal...

services/121-service/src/payments/payments.controller.ts
aberonni: How were the payments sorted before this change?
Would it make sense to sort them by date (if possible) rather than by id?
Are you sure that you have adjusted/checked all the places this is used in the old portal?

tijsziere: It wasn't sorted before, I now changed it to sort by date (directly in the query). Old portal still works as far as I can see

aberonni (dismissed):
I'm approving based on the assumption that

`Old portal still works as far as I can see`

Means "I searched the whole codebase of the Portal for instances of where we hit this endpoint" and does not mean "I looked at the Portal based on where I think this endpoint is used"

RubenGeo (approved):
well done

PR #5953, opened 2024-10-22
Small fixes visa reissue card

RubenGeo:
[AB#30852](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30852) 

### DIFF
--- a/services/121-service/src/registration/registrations.service.ts
+++ b/services/121-service/src/registration/registrations.service.ts
@@ -1114,7 +1114,7 @@ export class RegistrationsService {
       if (error instanceof IntersolveVisaApiError) {
         throw new HttpException(
           `${IntersolveVisa121ErrorText.reissueCard} - ${error.message}`,
-          HttpStatus.INTERNAL_SERVER_ERROR,
+          HttpStatus.BAD_REQUEST,
         );
       } else {
         throw error;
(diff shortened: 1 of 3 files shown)


services/121-service/src/registration/registrations.service.ts
```diff
@@ -1120,7 +1120,7 @@ export class RegistrationsService {
       if (error instanceof IntersolveVisaApiError) {
         throw new HttpException(
           `${IntersolveVisa121ErrorText.reissueCard} - ${error.message}`,
-          HttpStatus.INTERNAL_SERVER_ERROR,
```
RubenGeo, line 1123: I don't think it's nice to throw an HttpException with code INTERNAL_SERVER_ERROR because we do not get an email from it, which is confusing as we do get one on new Error() 

So I think we should either throw a new Error or a HttpException with 400 or 404.

In this case we sort of take into account that this error can occur, namely when a user changes an address property or a phonenumber to something that is not allowed by intersolve. Therefore I think this error is expected so I used a 400 here and not new Error()

diderikvw: Create a Task to look at alerting for 500 errors.

diderikvw (approved):

PR #5968, opened 2024-10-22
portalicious: custom messaging components

aberonni:
[AB#30847](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30847)

### DIFF
--- a/interfaces/Portalicious/src/app/domains/notification/notification.api.service.ts
+++ b/interfaces/Portalicious/src/app/domains/notification/notification.api.service.ts
@@ -1,7 +1,12 @@
-import { Injectable, Signal } from '@angular/core';
+import { inject, Injectable, LOCALE_ID, Signal } from '@angular/core';
 
 import { DomainApiService } from '~/domains/domain-api.service';
-import { MessageTemplate } from '~/domains/notification/notification.helper';
+import {
+  MessageTemplate,
+  MessageTemplateWithTranslatedLabel,
+} from '~/domains/notification/notification.model';
+import { TranslatableStringService } from '~/services/translatable-string.service';
+import { Locale } from '~/utils/locale';
 
 const BASE_ENDPOINT = (projectId: Signal<number>) => [
   'notifications',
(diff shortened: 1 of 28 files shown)


interfaces/Portalicious/src/app/pages/project/project-registrations/components/custom-message-preview/custom-message-preview.component.ts
```diff
@@ -19,21 +25,51 @@ import { SendMessageData } from '~/domains/registration/registration.api.service
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CustomMessagePreviewComponent {
-  sendMessageData = input<SendMessageData>();
-  // TODO: AB#30847 add additional property for dummy data to be used in the preview
+  readonly projectId = input.required<number>();
+  readonly previewRegistration = input.required<Registration>();
+  readonly messageText = input.required<string>();
```
aberonni, line 30: I changed this to accept `messageText` (the body of the message) instead of `sendMessageData` (body of the message if custom, otherwise template key) because I already had the body of the message to pass on in both scenarios.

However, I'm not sure what will be more convenient for the status change popups. If it's gonna be more convenient to pass in the template key for the scenario of templated messages (because maybe we'll only have easy access to that, and not the body of the message?) then I can change this.

interfaces/Portalicious/src/app/domains/notification/notification.api.service.ts
```diff
@@ -12,12 +17,37 @@ const BASE_ENDPOINT = (projectId: Signal<number>) => [
...
       path: [
         ...BASE_ENDPOINT(projectId as Signal<number>),
         'message-templates',
       ],
+      processResponse: (response) => {
+        return response
+          .filter(
+            (template) =>
+              template.isSendMessageTemplate &&
+              template.language === (this.locale as string),
```
arsforza, line 39: @aberonni the current portal always shows the English template and a program could have some untranslated templates, so for now this should be reverted to just English

```suggestion
              template.language === 'en',
```

arsforza (changes requested):

interfaces/Portalicious/src/app/domains/notification/notification.api.service.ts
aberonni: Checked in with @tijsziere and we can leave it as is: the templates will be loaded in the users' language, under the assumption that translations will be available.

arsforza (approved):

PR #5970, opened 2024-10-23
Fix.excel setup

jannisvisser:
[AB#30281](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30281)

## Describe your changes

- Fixes the Excel setup in the case of uploading multiple fspConfigs in one file. Previously this lead to a duplication of import records in the response. Now it takes the import records as a base, then loops over the fspConfigs, and then updates each base row if a match was found in that fspConfig. This way the number of rows stays constant by definition, instead of blowing up.
- Fixes API-tests
- Small additional fixes in separate commits


services/121-service/src/payments/payments.service.ts
```diff
@@ -1222,48 +1232,40 @@ export class PaymentsService {
...
+        payment,
+        programId,
+        fspConfigs: fspConfigsExcel,
+      });
     }
 
-    // Store the transactions after checking each FSP config
-    for (const importResult of importResults) {
+    for (const fspConfig of fspConfigsWithReconciliation) {
       await this.transactionsService.storeAllTransactionsBulk(
```
RubenGeo, line 1249: So we have 2 methods that are used for storing multiple transactions in the transactionService. Both are only used in the excelService since we removed vodacash. 
1. storeAllTransactionsBulk
  - Used when doing the initial send payment for excel
  - Uses objects which combine PaTransactionResultDto and TransactionRelationDetailsDto (which contains the FspConfigId) 
  - Seems more in line with current transaction-proccessor architecture
2. storeAllTransactions
  - Used when storing reconsiliation data 
  - Only works for one fspConfigId
  - Performance much better as it save in bulk

Should we remove one of these methods? And create a hybrid that accepts the parameters of storeAllTransactions but saves in bulk somehow?

Also should we test te performance of these excel imports after this refactor?

services/121-service/src/payments/fsp-integration/excel/excel.service.ts
```diff
@@ -266,10 +331,7 @@ export class ExcelService
         fspConfig.id,
       );
     if (!registrationsForReconciliation?.length) {
```
RubenGeo, line 333: I think this `if` fits better in the end of the `getRegistrationsForReconciliation` method

services/121-service/src/payments/fsp-integration/excel/excel.service.ts
```diff
@@ -243,21 +250,79 @@ export class ExcelService
...
+        programId,
+        payment,
+        validatedExcelImport,
+        fspConfig,
+        matchColumn,
+      });
+
+      // .. then loop over each row of the original import to update if a match has been found with this fspConfig
+      crossFspConfigImportResults.forEach((row, index) => {
+        const importResultForFspConfigRow = importResultForFspConfig.find(
```
RubenGeo, line 299: Nice that you solved the notFound problem like this.

I wonder about the performance of this a lot of records. Maybe it's better to use a hasmap lookup here instead of an array?

RubenGeo (changes requested):

PR #5990, opened 2024-10-28
Add forbidden and unautherized to all endpoint swagger docs

RubenGeo:
[AB#30959](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30959) 

## Describe your changes

Add forbidden and unauth http codes to all swagger endpoints

### DIFF
--- a/services/121-service/src/guards/authenticated-user.decorator.ts
+++ b/services/121-service/src/guards/authenticated-user.decorator.ts
@@ -1,4 +1,5 @@
-import { SetMetadata } from '@nestjs/common';
+import { applyDecorators, HttpStatus, SetMetadata } from '@nestjs/common';
+import { ApiResponse } from '@nestjs/swagger';
 
 import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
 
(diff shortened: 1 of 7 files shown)


services/121-service/src/guards/authenticated-user.decorator.ts
```diff
@@ -9,10 +10,20 @@ export interface AuthenticatedUserParameters {
...
-    ...parameters,
-    isGuarded: true,
-  });
+export const AuthenticatedUser = (parameters?: AuthenticatedUserParameters) => {
+  return applyDecorators(
+    SetMetadata('authenticationParameters', {
+      ...parameters,
+      isGuarded: true,
+    }),
+    ApiResponse({
```
diderikvw, line 19: Should we not remove the "status: HttpStatus.FORBIDDEN", and "status: HttpStatus.UNAUTHORIZED" API responses from all endpoints that use the AutenticatedUser decorator?

diderikvw (commented):
Nice refactor. See comment.

services/121-service/src/guards/authenticated-user.decorator.ts
RubenGeo: Ah yes thanks! I forgot to remove HttpStatus.UNAUTHORIZED because I added that one later

diderikvw: Still found FORBIDDEN in: services/121-service/src/financial-service-providers/financial-service-provider.controller.ts

diderikvw (changes requested):
See comment.

services/121-service/src/guards/authenticated-user.decorator.ts
RubenGeo: Ah I originally left that one because it had a different text. But I see FORBDDEN is never called in the service so I removed it

diderikvw (approved):

PR #5999, opened 2024-10-29
chore: put export-type enum in separate file

aberonni:
[AB#31009](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31009)

To be able to use ExportType enum easily in Portalicious, we need to separate it into a file of its own. This should be in line with the 121-service guidelines so it shouldn't be an issue, and is worth the refactor on its own IMHO.


aberonni:
@jannisvisser 

> Is this just how it is?

Pretty much so, yes. It's because of this setting that we enabled a while ago when introducing the PR guidelines:
image

So if you're in the midst of "PR merging traffic" then you'll have to update the branch a few times before it gets merged.

But tbh, I most of the time leave it to the author to make sure it gets merged.

That said,

> You have to click 'update branch' then.

doing that on behalf of the author is fine, but then I would make sure you use the dropdown to select "Update with rebase" rather than update with merge. (It would be great if this could be the default but alas [that is not possible](https://github.com/orgs/community/discussions/12032)).

PR #6000, opened 2024-10-29
portalicious: export selected registrations

aberonni:
AB#31003


interfaces/Portalicious/src/app/pages/project-registrations/components/export-registrations/export-registrations.component.ts
```diff
@@ -0,0 +1,155 @@
...
+    LatestExportDateComponent,
+    ButtonMenuComponent,
+    CalendarModule,
+    FloatLabelModule,
+    FormsModule,
+    FormFieldWrapperComponent,
+  ],
+  templateUrl: './export-registrations.component.html',
+  styles: ``,
+  providers: [ToastService],
```
PeterSmallenbroek, line 50: Doesn't need to be here as we're using inject()

aberonni: I don't understand why....but....it does 😅 

I think it has something to do with how the underlying `MessageService` works. But if we don't add the `ToastService` to the `providers`, angular complains.

Maybe worth (separately) doing some digging to understand why this is happening. But for now, I've just done this everywhere it was needed.

PR #6010, opened 2024-10-31
fix api & unit & e2e tests

RubenGeo:
[AB#30189](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30189) 


services/121-service/src/payments/fsp-integration/excel/excel.service.spec.ts
```diff
@@ -81,7 +81,7 @@ describe('ExcelService', () => {
     excelService = moduleRef.get<ExcelService>(ExcelService);
   });
 
-  it.skip('should find and return the matching reconciliation record for a given registration', async () => {
+  it('should find and return the matching reconciliation record for a given registration', async () => {
```
jannisvisser, line 102: I think these tests no longer work like this, because they test joinRegistrationsAndImportRecords which is by now only public for the sake of the unit test, so it actually shouldn't be public any more.

services/121-service/test/payment/do-payment-fsp-excel.test.ts
```diff
@@ -147,64 +149,39 @@ describe('Do payment with Excel FSP', () => {
     });
 
     // ##TODO: wait with fixing this test until endpoint is available to update/delete columnsToExport
-    it.skip('Should return all program-question/program-custom attributes on Get FSP instruction with Excel-FSP when "columnsToExport" is not set', async () => {
+    it('Should return all program-question/program-custom attributes on Get FSP instruction with Excel-FSP when "columnsToExport" is not set', async () => {
```
jannisvisser, line 152: change description to program-registration-attributes (also in shapshot)

services/121-service/test/program/create-program.test.ts
```diff
@@ -84,35 +87,7 @@ describe('Create program', () => {
     // Assert
     // const programId = createProgramResponse.body.id;
     expect(createProgramResponse.statusCode).toBe(HttpStatus.BAD_REQUEST);
-
-    // A new program should not have been created
-    expect(getProgramResponse.statusCode).toBe(HttpStatus.NOT_FOUND);
-  });
-
-  it.skip('should not be able to post a program with double names', async () => {
```
jannisvisser, line 92: why is this test removed?

services/121-service/test/program/update-program.test.ts
```diff
@@ -52,31 +51,6 @@ describe('Update program', () => {
     expect(updateProgramResponse.body.budget).toBe(program.budget);
   });
 
-  // ##TODO Enable/Refacetor/Remove this test after FSP config endpoints are implemented (and the logic is reused in update program). This test should than check if it is possible to add FSP config to a program if that is not aldreaddy covered by some other test
-  it.skip('should add an fsp to a program', async () => {
```
jannisvisser, line 56: is the new logic now covered in another test?

jannisvisser (changes requested):

e2e/test-registration-data/test-registrations-westeros-20.csv
```diff
@@ -1,21 +1,21 @@
...
-11dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-12dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-13dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-14dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238876,gringotts,stark
-15dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238856,gringotts,stark
-16dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-17dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-18dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
-19dc9451-1273-484c-b2e8-ae21b51a96ab,en,,Test,succeed,14155238886,gringotts,stark
+referenceId,preferredLanguage,maxPayments,firstName,lastName,phoneNumber,programFinancialServiceProviderConfigurationName,whatsappPhoneNumber,addressStreet,addressHouseNumber,addressHouseNumberAddition,addressPostalCode,addressCity,house
```
jannisvisser, line 1: all these address things are not attributes of Excel right? I removed them for that reason in another task. Why would you specifically need them back? I would put it in the actual excel attributes then if needed.

services/121-service/src/registration/dto/bulk-import.dto.ts
```diff
@@ -25,7 +25,7 @@ const fspArray = Object.values(FinancialServiceProviders).map((item) =>
   String(item),
 );
 const languageArray = Object.values(LanguageEnum).map((item) => String(item));
-export class BulkImportDto {
+class BulkImportDto {
```
jannisvisser, line 28: This should not have Dto suffix if it's not exported and therefore not used in controller any more right, according to guidelines? And it can lose the swagger decorators (and become an interface?). Related to this there are more things to change in this file I think. ImportResult should have Dto suffix. Split across files. Etc.

services/121-service/src/migrate-visa/migrate-visa.controller.ts
jannisvisser, line 1: you deleted this already assuming this will not be merged before the release I assume. Is there a note on that somewhere?

jannisvisser (changes requested):
separate review of the e2e commit/task in this PR.

services/121-service/test/program/update-program.test.ts
RubenGeo: We removed the functionality to also update program-financial-service-provider-configurations via PATCH /program. Added it to the handover [task](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20141?workitem=30191)

services/121-service/src/registration/dto/bulk-import.dto.ts
RubenGeo: Agreed. But feel like this is a bit out of scope for this work. I just removed the export here because our knip check was complaining about this export

services/121-service/src/migrate-visa/migrate-visa.controller.ts
RubenGeo: The reason I removed these files was because they were filled with typescript errors due to our changes and it seemed like a waste to fix them.

There is no note, I thought that our thinking in general would be that we never want to release those 2 big refactor branches at the same time. So thought it was not needed.

services/121-service/test/program/create-program.test.ts
RubenGeo: It's not right? These is still a `  it('should not be able to post a program with 2 of the same names', async () => {` it's just at line ~54

jannisvisser (approved):

PR #6061, opened 2024-11-08
Portalicious: refactor aggregates card AB#30918

arsforza:
[AB#30918](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30918)

### DIFF
--- a/interfaces/Portalicious/src/app/components/metric-container/metric-container.component.html
+++ b/interfaces/Portalicious/src/app/components/metric-container/metric-container.component.html
@@ -8,7 +8,11 @@
     } @else {
       {{ metric().value ?? '-' }}
       @if (metric().showAlert) {
-        <i class="pi pi-exclamation-triangle text-red-500"></i>
+        <i
+          class="pi pi-exclamation-triangle text-red-500"
+          title="!"
+          aria-label="!"
+        ></i>
       }
     }
   </dd>
--- a/interfaces/Portalicious/src/app/components/metric-container/metric-container.component.html
+++ b/interfaces/Portalicious/src/app/components/metric-container/metric-container.component.html
@@ -1,13 +1,15 @@
-<dt class="px-2 pb-2 pt-1 txt-body-m">
-  {{ label() }}
-</dt>
-<dd class="px-2 font-bold txt-h-2">
-  @if (pending()) {
-    <app-skeleton-inline />
-  } @else {
-    {{ value() ?? '-' }}
-    @if (showAlert()) {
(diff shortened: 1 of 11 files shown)


interfaces/Portalicious/src/app/components/metric-container/metric-container.component.html
```diff
@@ -1,13 +1,15 @@
...
+  <dt class="px-2 pb-2 pt-2 txt-body-m">
+    {{ metric().label }}
+  </dt>
+  <dd class="mt-auto border-b border-grey-300 px-2 pb-2 font-bold txt-h-2">
+    @if (!metric()) {
+      <app-skeleton-inline />
+    } @else {
+      {{ metric().value ?? '-' }}
+      @if (metric().showAlert) {
+        <i class="pi pi-exclamation-triangle text-red-500"></i>
```
elwinschmitz, line 11: Can this have a textual equivalent?

Like:
```suggestion
        <i
          class="pi pi-exclamation-triangle text-red-500"
          title="!"
          aria-label="!"
        ></i>
```

(Not 100% sure this works as-is; Could also need `[attr.aria-label]`...)

Note: I've tried to figure out a good way to make this reusable; But it seems that this emphasize is (currently) only used for the property "Failed transfers", in technically explaining the emphasize could be redundant/confusing.

My alternative suggestion would be to make the `showAlert`-property return a text-string that would contain the actual alert/warning-text.
But in this current use-case, that would result in a redundant warning-text that is also the value's `label`, (i.e. "`Failed transfers`"), so I thought that wasn't  worth it (for now).

interfaces/Portalicious/src/app/domains/metric/metric.model.ts
```diff
@@ -3,3 +3,10 @@ import { ProgramStats } from '@121-service/src/metrics/dto/program-stats.dto';
 import { Dto } from '~/utils/dto-type';
 // TODO: AB#30152 This type should be refactored to use Dto121Service
 export type ProjectMetrics = Dto<ProgramStats>;
+
+export interface SummaryMetric {
+  value: null | number | string | undefined;
+  label: string;
+  type?: string;
```
elwinschmitz, line 10: There is currently not use for this property, right?
Or is there already an implementations somewhere that uses this for switching between `number`s/`Date`s/etc?

elwinschmitz (changes requested):
Besides the small suggestion (and the rebase dance): LGTM(TM)

elwinschmitz (approved):

elwinschmitz (approved):

PR #6063, opened 2024-11-08
chore: remove nestjs-paginate hack

aberonni:
[AB#31237](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31237)

This is removing workarounds in our codebase that were introduced to solve two bugs.

- Bug [AB#30713](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30713) for which an API test was added when the fix was introduced
- Bug [AB#30520](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30520) for which an API test is added in this PR

We should be able to remove our workarounds thanks to [this release](https://github.com/ppetzold/nestjs-paginate/releases/tag/v9.3.1) which contains [this fix](https://github.com/ppetzold/nestjs-paginate/pull/1024).


services/121-service/test/registrations/pagination/filter-registration.test.ts
```diff
@@ -166,6 +166,28 @@ describe('Load PA table', () => {
       expect(meta.totalItems).toBe(1);
     });
 
+    it('should filter with using a referenceId string that could be interpreted as number', async () => {
+      // Arrange
+      const referenceIdInterpretableAsNumber = '651581942751358e5'; // A just-number string like '1234567890' is somehow not sufficiently covering the edge case, as this would not lead to the same bug prior to bugfix AB#30713
```
jannisvisser, line 171: make this a shared variable between all these tests? maybe split these tests off in separate file, now that they are 5?

jannisvisser (changes requested):
@aberonni we also added a check on importing referenceId with $ in it (https://github.com/global-121/121-platform/pull/5866/files), which was put in because of this hack. Should we reverse that? I guess fine to keep, but also a pretty random check now in import.

services/121-service/test/registrations/pagination/filter-registration.test.ts
aberonni: Good call. I've split the tests into a separate file, and also reverted the other change that you mention in your other comment.

jannisvisser (approved):
approved (if tests pass)

PR #6098, opened 2024-11-13
Do not get unneeded data on every request

RubenGeo:
[AB#31311](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31311) 

## Describe your changes

This query is called every time a user validate (so basically every request), but it is getting unneeded data. So I changed it. 

### DIFF
--- a/services/121-service/src/user/user.service.ts
+++ b/services/121-service/src/user/user.service.ts
@@ -522,7 +522,7 @@ export class UserService {
 
   public async findByUsernameOrThrow(
     username: string,
-    relations: FindOptionsRelations<UserEntity> | undefined,
+    relations?: FindOptionsRelations<UserEntity>,
   ): Promise<UserEntity> {
     const user = await this.userRepository.findOne({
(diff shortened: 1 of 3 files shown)


services/121-service/src/user/user.service.ts
```diff
@@ -520,14 +520,13 @@ export class UserService {
     return user;
   }
 
-  public async findByUsernameOrThrow(username: string): Promise<UserEntity> {
+  public async findByUsernameOrThrow(
+    username: string,
+    relations: FindOptionsRelations<UserEntity> | undefined,
```
aberonni, line 525:
```suggestion
    relations?: FindOptionsRelations<UserEntity>,
```

Small nitpick, but I find this generally makes it easier to, at a glance, check which parameters are optional.

aberonni (approved):
Small suggestion but otherwise looks good (and a nice catch)

aberonni (approved):

PR #6105, opened 2024-11-14
portalicious: enable dependabot

aberonni:
[AB#31346](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31346)

I noticed some dependencies were out of date. Then I started digging and noticed none of them had ever been updated by dependabot. That's when I noticed the typo in dependabot.yml.

I updated the dependencies manually while I was at it.


.github/dependabot.yml
```diff
@@ -111,7 +111,7 @@ updates:
       - dependency-name: 'zone.js'
 
   - package-ecosystem: npm
-    directory: '/interfaces/Portalicous'
+    directory: '/interfaces/Portalicious'
```
elwinschmitz, line 114: 🤯

.github/dependabot.yml
```diff
@@ -122,17 +122,14 @@ updates:
...
+          - 'typescript'
           - 'zone.js'
       dependencies:
         patterns:
           - '*'
     ignore:
-      # Dependencies below are ordered alphabetically!
-      # Most/all Angular-related dependencies are ignored for all versions
-      - dependency-name: '@types/node'
-        update-types: ['version-update:semver-major']
```
elwinschmitz, line 133: We want to keep the types in line with the actual Node.js-version we use, so it should not upgrade beyond v20 (at the moment).

There is unfortunately no way to say: `"ignore if version < .node-version"`

elwinschmitz (changes requested):

elwinschmitz (approved):

PR #6108, opened 2024-11-14
portalicious: table column management

aberonni:
[AB#31327](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31327)


interfaces/Portalicious/src/app/components/query-table/components/query-table-global-search/query-table-global-search.component.html
```diff
@@ -0,0 +1,37 @@
...
+        />
+      }
+    </p-iconField>
+  } @else {
+    <p-button
+      icon="pi pi-search"
+      (click)="globalFilterVisible.set(true)"
+      text
+      severity="secondary"
+    />
```
elwinschmitz, line 35:
```suggestion
    <p-button
      icon="pi pi-search"
      (click)="globalFilterVisible.set(true)"
      text
      severity="secondary"
      title="Filter by keyword"
      i18n-title="@@table-quick-search"
      aria-label="Filter by keyword"
      i18n-aria-label="@@table-quick-search"
    />
```

interfaces/Portalicious/src/app/components/query-table/components/query-table-global-search/query-table-global-search.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  globalFilterVisible = model<boolean>(false);
+  readonly onFilterChange = output<string | undefined>();
+
+  @ViewChild('globalFilterContainer')
+  globalFilterContainer: ElementRef<HTMLDivElement>;
+  @ViewChild('globalFilterInput')
+  globalFilterInput: ElementRef<HTMLInputElement>;
+
+  constructor(private renderer: Renderer2) {
+    this.renderer.listen('window', 'click', (e: Event) => {
```
elwinschmitz, line 36: This looks very low-level?

Is there no "dismiss-on-backdrop-click" or something?

And would this still listen after the component is gone, or is the renderer keeping track of that?

Also, I think the same would need to happen when pressing "Esc" on the keyboard?

interfaces/Portalicious/src/app/components/registrations-table/registrations-table.component.ts
```diff
@@ -114,6 +108,13 @@ export class RegistrationsTableComponent {
       getCellChipData: (registration) =>
         getChipDataByRegistrationStatus(registration.status),
     },
+    {
+      field: 'registrationCreated',
+      fieldForFilter: 'registrationCreatedDate',
+      header: $localize`:@@registration-created:Registration created`,
```
elwinschmitz, line 114: I wonder if its possible to get some 'system' in these names...
And how to check/enforce that system... 🤔 

In the old-portal there where also a few concepts used side-by-side.... so I don't know which should be applied everywhere.. ;)

But we should be able to provide some more context to translators with these.. And also to get some more "consistent" translations easily etc.

aberonni: By "system" you mean some kind of convention / naming guideline / taxonomy?

Indeed that would be interesting but I don't have an opinion / recommendation as of now. And also out of scope for this PR (I'm just moving code around here).

interfaces/Portalicious/src/app/components/query-table/components/query-table-global-search/query-table-global-search.component.ts
aberonni: All good questions, but out of scope for the PR. I'm just moving code around. That said, here are the answers.

> Is there no "dismiss-on-backdrop-click" or something?

I found this when googling how to do this stuff in Angular. Couldn't find anything pre-made.

> And would this still listen after the component is gone, or is the renderer keeping track of that?

Good question, I'm not sure.

> Also, I think the same would need to happen when pressing "Esc" on the keyboard?

Nobody asked for it, and this has been reviewed and used by everyone, so it looks like nobody needs/wants it.

interfaces/Portalicious/src/app/components/query-table/components/query-table-column-management/query-table-column-management.component.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  updateColumnVisibility = injectMutation(() => ({
+    // We don't technically need a mutation here, but we're using one
+    // so that we can easily reuse the form-sidebar component
+    mutationFn: () => Promise.resolve(),
+    onSuccess: () => {
+      this.visibleColumns.set(this.formGroup.getRawValue().selectedColumns);
+      this.formVisible.set(false);
+    },
+  }));
```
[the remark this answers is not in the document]
aberonni: Yeah this is kind of a fake mutation...

Maybe a better way to write this would be this now that I take another look at it:

```ts
updateColumnVisibility = injectMutation(() => ({
    // We don't technically need a mutation here, but we're using one
    // so that we can easily reuse the form-sidebar component
    mutationFn: () => Promise.resolve(this.formGroup.getRawValue().selectedColumns),
    onSuccess: (selectedColumns) => {
      this.visibleColumns.set(selectedColumns);
      this.formVisible.set(false);
    },
  }));
  ```
  
  But it doesn't really matter. I'm just using a mutation "for pretend" because we aren't actually mutating anything on the backend.

elwinschmitz (approved):

interfaces/Portalicious/src/app/components/query-table/components/query-table-column-management/query-table-column-management.component.ts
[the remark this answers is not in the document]
aberonni: I update this function with the snippet I shared above. Let me know if you see other improvements.

interfaces/Portalicious/src/app/components/query-table/components/query-table-global-search/query-table-global-search.component.ts
elwinschmitz: Well,  I think "the team" (in/excluding PO) is in charge of the DoD.

Acceptance Criteria is the area of 'outside stakeholders' and PO's, etc... But let's leave the semantics to Diderik. :D

One of the reasons/requirements of the UI-kit/library we searched for was its "accessibility-level".
And although I don't think "creating an accessible product" is "an absolute thing", it is something to aim for; and at least not "omit something that helps". So I consider it the same as some specific TypeScript/ESLint-toggle; "does is help? (to create a better product)", then we should add/include it. "does it make it worse? (or harder to work with in the future)", then we shouldn't.

elwinschmitz (approved):

interfaces/Portalicious/src/app/components/query-table/components/query-table-global-search/query-table-global-search.component.ts
aberonni: I don't disagree with you, I'm more concerned with this only being a priority when you look at a PR, and it being neglected when anyone else does.

PR #6109, opened 2024-11-14
portalicious: payment page breadcrumbs AB#31322

arsforza:
[AB#31322](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31322)

## Describe your changes

- Made breadcrumbs generic
- Created payment page header
- Changed a bit ifs and elses in the layout component

### DIFF
--- a/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
+++ b/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
@@ -1,0 +1,7 @@
+<a
+  [routerLink]="parentLink()"
+  class="font-normal hover:text-purple-700"
+  >{{ parentTitle() }}</a
+>
+<i class="pi pi-chevron-right mx-2"></i>
+{{ childTitle() }}
--- a/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.ts
+++ b/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.ts
@@ -1,0 +1,64 @@
+import {
+  ChangeDetectionStrategy,
+  Component,
+  computed,
+  input,
+} from '@angular/core';
+import { RouterModule } from '@angular/router';
+
+import { AppRoutes } from '~/app.routes';
+
+@Component({
+  selector: 'app-breadcrumbs-title',
+  standalone: true,
+  imports: [RouterModule],
+  templateUrl: './breadcrumbs-title.component.html',
+  styles: ``,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BreadcrumbsTitleComponent {
+  projectId = input.required<number>();
+  registrationId = input<number>();
+  registrationName = input<string>();
+  paymentDate = input<string>();
+
+  parentLink = computed(() => {
+    const base = `/${AppRoutes.project}/${this.projectId().toString()}`;
+
+    if (this.registrationId()) {
+      return `${base}/${AppRoutes.projectRegistrations}`;
+    }
+
+    if (this.paymentDate()) {
+      return `${base}/${AppRoutes.projectPayments}`;
+    }
+
+    return base;
+  });
+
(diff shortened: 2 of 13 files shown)


interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  imports: [RouterModule],
+  templateUrl: './breadcrumbs-title.component.html',
+  styles: ``,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BreadcrumbsTitleComponent {
+  projectId = input.required<number>();
+  registrationId = input<number>();
+  registrationName = input<string>();
+  paymentDate = input<string>();
```
aberonni, line 23: I'm a bit confused by the choice of props here. Instead of passing in context-specific properties, and adding business logic to this component, I would rather pass in context-agnostic data, and move the logic to the declaration of the instances. Then you avoid all the `if`s in this file.

So, something like:

```suggestion
  parentLink = input<RouterLink['routerLink']>();
  parentTitle = input<string>();
  childTitle = input<string>();
```

interfaces/Portalicious/src/app/domains/payment/payment.api.service.ts
```diff
@@ -7,7 +7,7 @@ import {
   PaymentStatus,
 } from '~/domains/payment/payment.model';
 
-const BASE_ENDPOINT = (projectId: Signal<number>) => [
+const BASE_ENDPOINT = (projectId: Signal<number | undefined>) => [
```
aberonni, line 10:
```suggestion
const BASE_ENDPOINT = (projectId: Signal<number>) => [
```

Ideally we should never call this with an undefined projectId. If needed, change the `enabled` of the related call to this method to avoid this. I'm afraid that changing this to allow undefined could have unexpected side effects with regards to caching and query key invalidation.

interfaces/Portalicious/src/app/domains/payment/payment.api.service.ts
```diff
@@ -23,7 +23,10 @@ export class PaymentApiService extends DomainApiService {
...
   }
 
-  getPayment(projectId: Signal<number>, paymentId: Signal<number | undefined>) {
+  getPayment(
+    projectId: Signal<number | undefined>,
+    paymentId: Signal<number | undefined>,
+  ) {
     return this.generateQueryOptions<PaymentAggregate>({
       path: [...BASE_ENDPOINT(projectId), paymentId],
       enabled: () => !!paymentId(),
```
aberonni, line 32: With regards to my previous comment:

```suggestion
      path: [...BASE_ENDPOINT(projectId as Signal<number>), paymentId],
      enabled: () => !!projectID() && !!paymentId(),
```

interfaces/Portalicious/src/app/components/page-layout/components/payment-header/payment-header.component.html
```diff
@@ -0,0 +1,16 @@
+<app-page-layout-title-and-actions>
+  <ng-container page-layout-title>
+    @if (payments.isPending()) {
+      <p-skeleton
+        width="24rem"
+        height="1.6rem"
+      />
```
aberonni, line 7: This `if` is the same as in `registration-header`.

Maybe we can refactor this to avoid this repetition too. An idea would be to add an `isPending` prop to `app-page-layout-title-and-actions`, and then render the skeleton inside that component instead.

Then we can remove this if and the same if from registration header.

aberonni (changes requested):
That `page-layout` file definitely wants a refactor :) But we can already make some improvements before we merge I think :v:

interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
aberonni, line 1: Another small thing: this component would probably be better off (at least for now) within the `components/page-layout/components` folder

aberonni:
Closing this in favour of #6145

PR #6110, opened 2024-11-14
Portalicious: SSO (redirect slice)

PeterSmallenbroek:
[AB#31188](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31188) 


interfaces/Portalicious/src/app/pages/auth-callback/auth-callback.page.ts
```diff
@@ -0,0 +1,18 @@
...
+  standalone: true,
+  imports: [NgComponentOutlet],
+  templateUrl: './auth-callback.page.html',
+  styles: ``,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class AuthCallbackPageComponent {
+  private authService = inject(AuthService);
+
+  CallbackComponent = this.authService.CallbackComponent;
```
aberonni, line 17: Minor detail but I think the AuthCallbackPageComponent should be the one retrieving the returnUrl, and passing it on to the callback component

interfaces/Portalicious/src/app/services/auth.service.ts
```diff
@@ -1,64 +1,50 @@
...
 
   constructor() {
-    const AuthStrategy = environment.use_sso_azure_entra
-      ? MsalAuthStrategy
-      : BasicAuthStrategy;
-
-    this.authStrategy = this.injector.get(AuthStrategy);
+    this.authStrategy = this.injector.get<IAuthStrategy>(AuthStrategy);
   }
+  public authError = signal<string | undefined>(undefined);
```
aberonni, line 47: I'd like to see if we can remove this from here (see my other comments).

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.callback.component.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor() {
+    this.msalSubscription = this.msalService
+      .handleRedirectObservable()
+      .subscribe((data: AuthenticationResult | null) => {
+        if (!data) {
+          return;
+        }
+        void this.authService
+          .processAzureCallback()
```
aberonni, line 40: Related to my comment in auth-service -> processAzureCallback

```suggestion
          .refreshUser()
```

interfaces/Portalicious/src/app/services/auth.service.ts
```diff
@@ -207,4 +185,22 @@ export class AuthService {
...
+      // TODO: AB#31489 Return a more generic endpoint from the back-end
+      if (error instanceof HttpErrorResponse) {
+        const errorMessage = get(error.error, 'message') as string | undefined;
+        if (errorMessage) {
+          this.authError.set(errorMessage);
+        }
+      }
+      throw error;
+    }
+  }
```
aberonni, line 205:
```suggestion
  async refreshUser() {
    const currentUser = await this.queryClient.fetchQuery(
      this.userApiService.getCurrent()(),
    );
    setUserInLocalStorage(currentUser.user);
  }
```

Then the error catching / handling happens in the component. And I think this way we might be able to avoid setting `authError` in the auth service too.

interfaces/Portalicious/src/app/services/auth.service.ts
```diff
@@ -113,12 +87,16 @@ export class AuthService {
...
     returnUrl?: string,
   ) {
     this.logService.logEvent(LogEvent.userLogin);
+    if (returnUrl) {
+      setReturnUrlInLocalStorage(returnUrl);
+    }
     const user = await this.authStrategy.login(credentials);
-    this.setUserInStorage(user);
+    // Note: SSO never resolves so the code below this line is not executed in the SSO case
+    setUserInLocalStorage(user);
```
aberonni, line 99:
```suggestion
    // Note: SSO never resolves so the code below this line is not executed in the SSO case
    setUserInLocalStorage(user);
    
    if (!!this.authStrategy.CallbackComponent) {
      return this.router.navigate([AppRoutes.authCallback]);
    }
```

It's a small addition, but it would make it a bit more explicit.

Also, we could still alternatively _always_ redirect to the callback, and make it mandatory. And then implement a callback for the basic-auth strategy that only does the redirect and nothing else. I think this might be cleaner.

interfaces/Portalicious/src/app/pages/auth-callback/auth-callback.page.html
```diff
@@ -0,0 +1 @@
+<ng-container *ngComponentOutlet="CallbackComponent"></ng-container>
```
aberonni, line 1: Should we maybe add `app-page-layout` here? To minimize page jumping around (especially if we use this also for basic-auth - see the other comment about this).

interfaces/Portalicious/src/app/services/auth/auth-strategy.interface.ts
```diff
@@ -1,13 +1,16 @@
...
 import { User } from '~/domains/user/user.model';
-import { LocalStorageUser } from '~/services/auth.service';
+import { LocalStorageUser } from '~/utils/local-storage';
 
 export abstract class IAuthStrategy {
+  public static readonly APP_PROVIDERS: (EnvironmentProviders | Provider)[];
+
   LoginComponent: Type<unknown>;
   ChangePasswordComponent: null | Type<unknown>;
+  CallbackComponent: null | Type<unknown>;
```
aberonni, line 11:
```suggestion
  CallbackComponent: Type<unknown>;
```

Maybe... (see my other comments)

interfaces/Portalicious/src/app/services/auth/auth-strategy.interface.ts
```diff
@@ -1,13 +1,16 @@
...
 
-  public abstract login(credentials: {
+  public abstract login(credentials?: {
```
aberonni, line 13:
```suggestion
  public abstract login(credentials: {
```

I have a feeling this might've been added before passing the email along.

interfaces/Portalicious/src/app/services/auth.service.ts
```diff
@@ -113,12 +87,16 @@ export class AuthService {
   }
 
   public async login(
-    credentials: { username: string; password?: string },
+    credentials?: { username: string; password?: string },
```
aberonni, line 90:
```suggestion
    credentials: { username: string; password?: string },
```

Same as auth-strategy interface

interfaces/Portalicious/src/app/services/auth/strategies/basic-auth/basic-auth.strategy.ts
```diff
@@ -61,4 +64,6 @@ export class BasicAuthStrategy implements IAuthStrategy {
   public isUserExpired(user: LocalStorageUser | null): boolean {
     return !user?.expires || Date.parse(user.expires) < Date.now();
   }
+
+  public authError: Signal<string | undefined>;
```
aberonni, line 68: Is this used / needed anywhere?

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.callback.component.html
```diff
@@ -0,0 +1,4 @@
+<div class="flex h-screen w-full flex-col items-center justify-center">
+  <p-progressSpinner />
+  <h2 i18n>Redirecting to 121&hellip;</h2>
+</div>
```
aberonni, line 4: I think this should be in the `AuthCallbackPageComponent`, and the `MsalAuthCallbackComponent` component should actually just have empty HTML (you could use `template: ''` instead of `templateUrl`). With the idea that the auth-callback page will always look the same regardless of the strategy.

EDIT: it makes me wonder whether the callback-component is needed at all, or whether the strategy should just implement a "authCallback" function, which in the case of msal-auth would just contain the contents of the constructor below.

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.callback.component.ts
```diff
@@ -0,0 +1,60 @@
...
+            const returnUrl = getReturnUrlFromLocalStorage();
+            if (returnUrl) {
+              await this.router.navigate([returnUrl]);
+            } else {
+              await this.router.navigate(['/', AppRoutes.users]);
+            }
+          })
+          .catch(async () => {
+            await this.router.navigate(['/', AppRoutes.login]);
+          });
```
aberonni, line 51: Related to my comment in auth-service -> processAzureCallback

```suggestion
          .catch(async (error) => {
            let errorMessage: string | undefined;

            // TODO: AB#31489 Return a more generic endpoint from the back-end
            if (error instanceof HttpErrorResponse) {
              errorMessage = get(error.error, 'message') as string | undefined;
            }
            if (!errorMessage) {
              errorMessage = $localize`Some generic error message`;
            }

            // somehow pass the error to the login page (I'm not sure how to do this)
            await this.router.navigate(['/', AppRoutes.login, error: errorMessage]);
          });
```

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.login.component.html
```diff
@@ -0,0 +1,36 @@
...
+@if (authErrorText) {
+  <span class="mt-2 block text-red-700">
+    <i
+      class="pi pi-exclamation-triangle"
+      title="!"
+      aria-label="!"
+    ></i>
+    {{ authErrorText }}
+  </span>
+}
```
aberonni, line 36:
```suggestion
<app-form-error [error]="authErrorText">
```

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.login.component.ts
```diff
@@ -0,0 +1,78 @@
...
+export class MsalAuthLoginComponent {
+  private authService = inject(AuthService);
+  private route = inject(ActivatedRoute);
+  private returnUrl = computed(() => {
+    const returnUrl: unknown = this.route.snapshot.queryParams.returnUrl;
+    if (typeof returnUrl !== 'string') {
+      return undefined;
+    }
+    return returnUrl;
+  });
```
aberonni, line 56: The login page should probably pass this in, now that both strategies are using the same code

```suggestion
  private returnUrl = input<string>();
```

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -1,10 +1,89 @@
...
+    if (!isIframed()) {
+      this.msalService.loginRedirect({
+        scopes: [`api://${environment.azure_ad_client_id}/User.read`],
+        loginHint: credentials.username,
+      });
+    } else {
+      throw new Error('TODO: AB#31469 Implement loginPopup for iframe');
+    }
+
+    // The user is being fetched & set in processAzureCallback
```
aberonni, line 43:
```suggestion
    // The user is being fetched & set in msal-auth.callback.component
```

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -1,10 +1,89 @@
...
+      return;
+    }
+
+    const currentUser = this.msalService.instance.getAccountByUsername(
+      user.username,
+    );
+
+    if (!currentUser) {
+      await this.router.navigate(['/', AppRoutes.login]);
+      return;
```
aberonni, line 62:
```suggestion
      return this.router.navigate(['/', AppRoutes.login]);
```

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -1,10 +1,89 @@
...
+    } else {
+      this.msalService.logoutRedirect(logoutRequest);
+    }
+    return Promise.resolve();
+  }
+
+  public async changePassword() {
+    return Promise.reject(
+      new Error('This should never be called for MSAL service.'),
+    );
```
aberonni, line 82:
```suggestion
    throw new Error('This should never be called for MSAL service.');
```

aberonni (changes requested):
Let me know if anything doesn't make sense :v:

interfaces/Portalicious/src/app/app.routes.ts
```diff
@@ -7,6 +7,7 @@ import { authCapabilitiesGuard } from '~/guards/auth-capabilities.guard';
 import { projectPermissionsGuard } from '~/guards/project-permissions-guard';
 
 export enum AppRoutes {
+  authCallback = 'auth-callback',
```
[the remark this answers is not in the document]
PeterSmallenbroek: OGD needs to add Portalicious anyway as it's a separate URL. Dom and I believe that auth-callback is a better name.

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -1,10 +1,89 @@
...
+  }
+
+  public isUserExpired() {
+    // Not applicable for MSAL
```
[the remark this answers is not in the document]
PeterSmallenbroek: I've parked this (discussion) in AB#31530

RubenGeo (approved):

PR #6111, opened 2024-11-15
portalicious: create payment popup

aberonni:
[AB#31328](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31328)


interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.html
```diff
@@ -0,0 +1,61 @@
...
+        />
+      </p-card>
+    }
+  </div>
+  <ng-template pTemplate="footer">
+    <div class="w-1/2 border-t-2 border-t-purple-300"></div>
+    <div class="flex items-center justify-end p-4">
+      <span
+        i18n
+        class="absolute left-1/2 -translate-x-1/2"
```
elwinschmitz, line 49: Does `start-1/2` also work?

And would there be an equivalent for `translate-x`?

interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.ts
```diff
@@ -0,0 +1,54 @@
...
+  @ViewChild('registrationsTable')
+  registrationsTable: RegistrationsTableComponent;
+
+  today = new Date();
+
+  addSelectedToPayment() {
+    this.toastService.showToast({
+      severity: 'success',
+      summary: 'You did it!',
+      detail: 'You clicked on a button. Well done! You deserve a cookie: 🍪',
```
elwinschmitz, line 51: What if the user has `DoNotTrack` enabled?? :P

interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.html
```diff
@@ -0,0 +1,61 @@
...
+>
+  <ng-template pTemplate="header">
+    <h3>
+      <i class="pi pi-money-bill me-2"></i>
+      <ng-container i18n="@@create-payment">Create new payment</ng-container>
+      - {{ today | date: 'shortDate' }}
+    </h3>
+  </ng-template>
+  <div class="w-full px-6 lg:px-24">
+    @defer (when dialogVisible()) {
```
elwinschmitz, line 25: This syntax is so nice and readable. :)

elwinschmitz (changes requested):
Besides the RTL/LTR-issues: LGTM 👌

interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.html
aberonni: I couldn't find an equivalent for `translate-x`. That said, given that we're centering the item, it wouldn't make a difference. It could be `right-1/2 translate-x-1/2` and it would be the same result.

elwinschmitz: In this case it is "centering in relation to its own width" right?

Then it won't matter, as its parent container would align properly because of the `justify-end`-class, I'd think.

elwinschmitz (dismissed):
I think we will be re-using/changing the stepper-thing in some other items anyway, so we'll see.. :)

I don't know if the E2E-tests failing is related to this PRs changes though?

interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.html
aberonni: I tried both of those already. The first option is differnt from the design. The second option doesn't work:

image

For now I'm leaving it as-is.

interfaces/Portalicious/src/app/app.routes.ts
aberonni, line 1: @elwinschmitz the E2E tests were indeed failing because of my changes. This was a weird one. I fixed it like below, you can see the commit message for more details.

elwinschmitz: Hmm.. weird.
If this fixes it, it smells like "something async"...

So the only change that could possibly be related would be the `@defer`-stuff...
Is there something that we need to configure in our build to start using that? Or are we missing an `async`-keyword somewhere?

aberonni: > Is there something that we need to configure in our build to start using that? Or are we missing an async-keyword somewhere?

Nope, we're already using it elsewhere. But also,

> So the only change that could possibly be related would be the @defer-stuff...

Not quite. During my debugging I commented out _only_ the `app-registrations-table` usage (so leaving the defer in there) and everything worked fine. So it's something specific to using the table in two places.

My guess is something deeply nested that has to do with primeng using modules, and us using standalone components, given that most of the stack overflow results mentioned modules vs standalone. But, tbh, I don't care enough to dig deeper given that this works.

elwinschmitz (approved):

interfaces/Portalicious/src/app/app.routes.ts
elwinschmitz: But the runs on GitHub Actions do not fully agree... it seems. 😿

aberonni: <Insert getting very angry GIF>

I'll take a look

aberonni: Whilst hunting down the issue in the e2e tests, I ended up refactoring some things there to make them easier to understand and more reusable (IMHO). The result is this commit: https://github.com/global-121/121-platform/pull/6111/commits/19411e3f2719ad6de6008117ebccf34280b67c02

Hopefully that fixes the tests.

elwinschmitz (approved):

PR #6118, opened 2024-11-18
Comments didirik refactor registration data second slice

RubenGeo:
[AB#30873](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30873) 

### DIFF
--- a/services/121-service/src/program-financial-service-provider-configurations/program-financial-service-provider-configurations.repository.ts
+++ b/services/121-service/src/program-financial-service-provider-configurations/program-financial-service-provider-configurations.repository.ts
@@ -20,7 +20,7 @@ export class ProgramFinancialServiceProviderConfigurationRepository extends Repo
     );
   }
 
-  public async findByProgramIdAndFinancialServiceProviderName({
+  public async getByProgramIdAndFinancialServiceProviderName({
     programId,
     financialServiceProviderName,
     includeProperties,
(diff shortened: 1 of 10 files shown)


services/121-service/src/program-financial-service-provider-configurations/program-financial-service-provider-configurations.repository.ts
```diff
@@ -67,6 +67,26 @@ export class ProgramFinancialServiceProviderConfigurationRepository extends Repo
     return response;
   }
 
+  // This methods specfically does not throw as it also used to check if the property exists
```
RubenGeo, line 70: I ended up adding this function. Because I specifically needed a function to test if a property exists for the Excel service instead of a function that throws if something does not exist. If we refactor excel maybe we can reconsider having this here

services/121-service/src/program-financial-service-provider-configurations/program-financial-service-provider-configurations.repository.ts
```diff
@@ -20,7 +20,7 @@ export class ProgramFinancialServiceProviderConfigurationRepository extends Repo
     );
   }
 
-  public async findByProgramIdAndFinancialServiceProviderName({
+  public async getByProgramIdAndFinancialServiceProviderName({
     programId,
     financialServiceProviderName,
     includeProperties,
```
diderikvw, line 26: I suggest to remove this parameter and let the function always return the properties. A bit of overhead because data may be retrieved that is not used, but making it simpler for the caller not to be bothered with this IMO outweighs that. At the moment this function is only called once with true.

services/121-service/src/program-financial-service-provider-configurations/program-financial-service-provider-configurations.repository.ts
```diff
@@ -83,14 +67,34 @@ export class ProgramFinancialServiceProviderConfigurationRepository extends Repo
     return response;
   }
 
+  // This methods specfically does not throw as it also used to check if the property exists
+  public async getPropertyValueByName({
```
diderikvw, line 71: Why not make it simpler by having getPropertiesByNames() instead? Why would one function return names and values and another very similar function only value?
Plus I would be fine with the custom repo returning entities. Then the data structure is transparent to the caller, and they can transform it to whatever they need,

diderikvw (approved):

PR #6145, opened 2024-11-20
portalicious: payments breadcrumbs

aberonni:
[AB#31322](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31322)

Continues the work started in #6109 

### DIFF
--- a/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
+++ b/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
@@ -1,0 +1,7 @@
+<a
+  [routerLink]="parentLink()"
+  class="font-normal hover:text-purple-700 focus:text-purple-700"
+  >{{ parentTitle() }}</a
+>
+<i class="pi pi-chevron-right mx-2"></i>
+{{ childTitle() }}
--- a/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.ts
+++ b/interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.ts
@@ -1,0 +1,16 @@
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+import { RouterLink, RouterModule } from '@angular/router';
+
+@Component({
+  selector: 'app-breadcrumbs-title',
+  standalone: true,
+  imports: [RouterModule],
+  templateUrl: './breadcrumbs-title.component.html',
+  styles: ``,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class BreadcrumbsTitleComponent {
+  parentLink = input<RouterLink['routerLink']>();
+  parentTitle = input<string>();
+  childTitle = input<string>();
+}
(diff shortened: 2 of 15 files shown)


interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
```diff
@@ -0,0 +1,7 @@
+<a
+  [routerLink]="parentLink()"
+  class="font-normal hover:text-purple-700"
```
elwinschmitz, line 3:
```suggestion
  class="font-normal focus:text-purple-700 hover:text-purple-700"
```

elwinschmitz: This is one of the real downsides of the approach of Tailwind... that this 'functional' connection between 2 lines of CSS and its use in templates has to be remembered+repeated everywhere and every time..

Instead of a single place in a (S)CSS-file somewhere with:
```
a:hover,
a:focus {
  color: var(--color-link-highlight);
}
```
(Or on something else then `a`-elements, i.e. a `.special-link`-class...)

aberonni: Dammit how many times will you have to point that out until I start remembering it? 😄 FYI I've added this https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/31483

elwinschmitz (changes requested):
I have a lead for a lint-rule/tailwind-plugin fix to automate the "not forgetting to style for all input-modes"...

But in the meantime... see the code-suggestion ;)

interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
aberonni: > I have a lead for a lint-rule/tailwind-plugin fix to automate the "not forgetting to style for all input-modes"...

This would be great.

> This is one of the real downsides of the approach of Tailwind... that this 'functional' connection between 2 lines of CSS and its use in templates has to be remembered+repeate

Yeah. I think it's also "on us" in a way, that we should be [composing our components in a way to avoid this repetition](https://tailwindcss.com/docs/reusing-styles). In most places we're doing a good job of it, but not everywhere. For example, maybe this should become a `app-link` component /shrug

For now, I think we don't need to overengineer it, but we can definitely re-iterate further down the line (especially once we've implemented the primeng theme, and we'll have a better feel for what the final result of our CSS will look like).

elwinschmitz (approved):

interfaces/Portalicious/src/app/components/page-layout/components/breadcrumbs-title/breadcrumbs-title.component.html
elwinschmitz: Some of these issues could be resolved by "storing the thinking in a component" indeed.. 
That will also help our back-end-friends to "know what to use" (and not needing all the knowledge)...

But "creating the perfect set of components" in and of itself is already a big over-engineering-opportunity of course. ;)

PR #6160, opened 2024-11-26
Portalicious: SSO (popup slice)

PeterSmallenbroek:
[AB#31469](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31469) 

## Describe your changes

These are the changes required for the SSO popup scenario. It's a straight forward change, the only thing I'm not super happy with is the need to downgrade the package (a couple minor versions).


interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -98,7 +114,7 @@ export class MsalAuthStrategy implements IAuthStrategy {
     const subscription = this.msalService
       .handleRedirectObservable()
       .subscribe((data: AuthenticationResult | null) => {
-        if (!data) {
+        if (!data && !isIframed()) {
```
aberonni, line 117: Why this change?

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -39,13 +43,25 @@ export class MsalAuthStrategy implements IAuthStrategy {
...
+          if (!data) {
+            throw new Error(
+              'MSAL Strategy: an error occurred while logging in with popup',
+            );
+          }
+          void this.router.navigate(['/', AppRoutes.authCallback]);
+        });
     }
 
     // The user is being fetched & set in msal-auth.callback.component
```
aberonni, line 67: Does this also apply to the popup logic?

interfaces/Portalicious/package.json
```diff
@@ -34,8 +34,8 @@
     "@angular/router": "^18.2.11",
     "@microsoft/applicationinsights-web": "^3.3.4",
     "@tanstack/angular-query-experimental": "^5.60.0",
-    "@azure/msal-angular": "^3.1.0",
-    "@azure/msal-browser": "^3.27.0",
+    "@azure/msal-angular": "3.0.25",
```
aberonni, line 37: If we need to block these versions, we should also enforce it via the dependabot file

aberonni (changes requested):

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
PeterSmallenbroek: I'm not sure I understand your question.
The router.navigate is needed in the popup subscribe because in this scenario we don't have a redirectUrl (like in loginRedirect). We have to handle the 'redirects' ourself by navigating to the page.

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
PeterSmallenbroek: Apparently some handling of this already happens 'under the hood' in the popup subscribe. So in the popup scenario the data here is null. But it does have the token etc. so it's fine to continue.

interfaces/Portalicious/package.json
PeterSmallenbroek: Good one, let me figure that one out

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
aberonni: That makes me wonder whether we need the subscribe at all in that scenario, and whether we can just exit early from this function if `isIframed()`

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
aberonni: Ah ok. Then I think the current code is a bit misleading, but I'll add a suggestion to clarify.

interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts
```diff
@@ -39,13 +43,25 @@ export class MsalAuthStrategy implements IAuthStrategy {
   }
 
   public async login(credentials: { username: string }) {
```
aberonni, line 45: Turns out I can't do a suggestion on the whole function so, something like:

```ts
login() {
  const loginRequest: PopupRequest | RedirectRequest = {
    scopes: [`api://${environment.azure_ad_client_id}/User.read`],
    loginHint: credentials.username,
  };

  if (isIframed()) {
    return new Promise((resolve, reject) => {
      const sub = this.msalService
        .loginPopup(loginRequest)
        .subscribe((data: AuthenticationResult | null) => {
          sub.unsubscribe();
          if (!data) {
            return reject(new Error(
              'MSAL Strategy: an error occurred while logging in with popup',
            ));
          }
          resolve(null); // auth.service needs to be adjusted to support null coming back from here
        });
      });
   }
   
   // here you have what was in this function before
}
``

This way we leave the logic of "where should we go after we've logged in" to the auth service.

File: interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts (hunk 1 above)
Comment (aberonni) [reply interfaces/Portalicious/src/app/services/auth/strategies/msal-auth/msal-auth.strategy.ts:117]: Something like

```ts
public handleAuthCallback(nextPageUrl: string) {
    if (isIframed()) {
      void this.refreshUserAndNavigate(nextPageUrl);
      return;
    }

    const subscription = this.msalService
      .handleRedirectObservable()
      .subscribe((data: AuthenticationResult | null) => {
        if (!data) {
          return;
        }
        subscription.unsubscribe();

        void this.refreshUserAndNavigate(nextPageUrl);
      });
  }
```

aberonni (approved):

interfaces/Portalicious/package.json
elwinschmitz: We _should_ keep an eye on the _next, fixed_ release of both packages though...

So I think we can remove them from the Dependabot-ignore list fairly soon... (especially now that we let it check less often)

elwinschmitz: I've subscribed to the "Releases" of the repo; But don't know yet how often these are created... and if I would notice the GitHub-notification (or maybe it'll be an email)

PR #6174, opened 2024-11-27
portalicious: add FinancialServiceProviderConfigurationApiService

aberonni:
and update payment overview accordingly

[AB#31594](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31594)


services/121-service/src/payments/payments.service.ts
```diff
@@ -329,13 +326,13 @@ export class PaymentsService {
...
   private async validateMissingFspConfigurations(
-    programFinancialServiceProviderConfigurationId: number,
+    programFinancialServiceProviderConfigurationName: string,
   ): Promise<string[]> {
     const config =
       await this.programFinancialServiceProviderConfigurationRepository.findOneOrFail(
         {
           where: {
-            id: Equal(programFinancialServiceProviderConfigurationId),
+            name: Equal(programFinancialServiceProviderConfigurationName),
```
aberonni, line 335: @RubenGeo the API test that is failing seems to be failing (I think) because of this change. I tried debugging it, and we get a config with a different ID - which makes me think that name is not actually unique. Maybe because of something in the test data? I'm not sure.

RubenGeo: Ah shit we are missing a programId: Equal(programId) in this query here next to the name. Thank you api tests! 👍

PR #6154, opened 2024-11-21
portalicious: excel payments

aberonni:
[AB#31330](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31330)

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have added tests wherever relevant
- [x] I have made sure that all automated checks pass before requesting a review
- [x] I do not need any deviation from our PR guidelines

PR #6175, opened 2024-11-27
feat: refactor FSPInstructions and add payment status endpoint

aberonni:
[AB#31608](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31608) [AB#31609](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31609)

While working on Portalicious, I realised there was logic in the Portal and the backend that isn't being used anywhere.

- removes unused ExportFileType
- removes unnecessary CsvInstructions
- updates the Portal and service code accordingly

Also adding a new endpoint that makes polling for this specific information easier

- adds programs/:programId/payments/:payment/status

I combined the above into a single PR because they are both small changes, and they are already sharded off of #6154, but I can split into two separate PRs if necessary / useful.

### DIFF
--- a/services/121-service/src/payments/payments.controller.ts
+++ b/services/121-service/src/payments/payments.controller.ts
@@ -89,6 +89,31 @@ export class PaymentsController {
     return await this.paymentsService.getProgramPaymentsStatus(programId);
   }
 
+  @AuthenticatedUser({ permissions: [PermissionEnum.PaymentREAD] })
+  @ApiOperation({ summary: 'Get status of a specific payment.' })
+  @ApiParam({ name: 'programId', required: true, type: 'integer' })
+  @ApiParam({
+    name: 'payment',
+    required: true,
+    type: 'integer',
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    description: 'Status of the specific payment',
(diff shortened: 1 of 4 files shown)


services/121-service/src/payments/payments.controller.ts
```diff
@@ -89,6 +89,31 @@ export class PaymentsController {
     return await this.paymentsService.getProgramPaymentsStatus(programId);
   }
 
+  @AuthenticatedUser({ permissions: [PermissionEnum.PaymentREAD] })
```
jannisvisser, line 92: do I understand correctly that this new endpoint is to be used in Portalicious only then and not in Portal?

jannisvisser (approved):
looks good and straightforward from code. See 1 question in comment.

services/121-service/src/payments/payments.controller.ts
aberonni: Yes correct, unless it becomes necessary in the Portal (but that would have to be an explicit request which is unlikely).

PR #6181, opened 2024-11-27
feat: retry failed jobs on startup

jannisvisser:
[AB#31519](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31519)

## Describe your changes

Retry failed Redis jobs on 121-service startup (so when back up after being down / deployment / maybe SNAT?). See comments in devops task.


RubenGeo (approved):
Should we cover this in an automated test somehow? @Jannis Visser

An ugly solution I can think of is an api test:

    reset with nlrc-mock
    do payment
    call an endpoint of the 121-service that we introduce that specifically restarts the server
    See if all transaction resolve

Do you think this is worth it or should we look for another solution?






@jannisvisser I was wondering if there is a way we can make sure we do not forget to add new queues to this list? Like can we do this InjectQueue in a loop over all queues somehow so we do not forget. Or import it in a different manner.

Or if that's not possible loop over all QueueNames (from the enums) and for each get the injected queue and throw if one is missing?

RubenGeo (approved):

PR #6183, opened 2024-11-28
portalicious: remove todo comment

aberonni:
AB#31559

I'm just removing the comment and leaving the copy as-is. If Tal / Tys don't like it, we'll hear about it in review.

And I'm also sneaking in there a little fix where we showed "undefined registration(s) imported"


elwinschmitz (approved):
We could've used all these fancy "`{ plural, =0, many }`"-expressions!

Such a shame, all this less-complexity. :D

PR #6188, opened 2024-11-28
Add swagger json automatically

RubenGeo:
AB#[31624](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/31624)
## Describe your changes

[AB#31624](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31624)

### DIFF
--- a/services/121-service/src/main.ts
+++ b/services/121-service/src/main.ts
@@ -73,6 +73,13 @@ function generateModuleDependencyGraph(app: INestApplication): void {
   });
 }
 
+interface MethodInfo {
+  method: string;
+  path: string;
+  params: string[];
+  returnType?: string;
+}
+
 function generateSwaggerSummaryJson(app: INestApplication<any>): void {
   const options = new DocumentBuilder()
     .setTitle(APP_TITLE)
@@ -80,11 +87,7 @@ function generateSwaggerSummaryJson(app: INestApplication<any>): void {
     .build();
   const openApiDocument = SwaggerModule.createDocument(app, options);
 
-  const minimalDocument: {
-    method: string;
-    path: string;
-    returnType: string;
-  }[] = [];
+  const summaryDocument: MethodInfo[] = [];
 
   for (const path in openApiDocument.paths) {
     for (const method in openApiDocument.paths[path]) {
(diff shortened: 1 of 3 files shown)


services/121-service/src/main.ts
```diff
@@ -71,6 +73,39 @@ function generateModuleDependencyGraph(app: INestApplication): void {
...
+  const options = new DocumentBuilder()
+    .setTitle(APP_TITLE)
+    .setVersion(APP_VERSION)
+    .build();
+  const openApiDocument = SwaggerModule.createDocument(app, options);
+
+  const minimalDocument: {
+    method: string;
+    path: string;
+    returnType: string;
```
aberonni, line 86:
```suggestion
    returnType?: string;
```

Instead of setting it to a hardcoded string, I would simply not set it.

services/121-service/src/main.ts
```diff
@@ -71,6 +73,39 @@ function generateModuleDependencyGraph(app: INestApplication): void {
...
+  }[] = [];
+
+  for (const path in openApiDocument.paths) {
+    for (const method in openApiDocument.paths[path]) {
+      const methodInfo = openApiDocument.paths[path][method];
+      const returnType =
+        methodInfo.responses['200']?.content?.['application/json']?.schema?.$ref
+          ?.split('/')
+          .pop() || 'Not set/ undefined';
```
aberonni, line 95:
```suggestion
          .pop();
```
related to my other comment above.

services/121-service/src/main.ts
```diff
@@ -89,9 +124,7 @@ async function bootstrap(): Promise<void> {
     credentials: true,
   });
 
-  if (DEBUG) {
-    generateModuleDependencyGraph(app);
-  }
+  console.log('🚀 ~ bootstrap ~ DEBUG:', DEBUG);
```
aberonni, line 127:
```suggestion
```
Or was this intentional?

services/121-service/src/main.ts
```diff
@@ -71,6 +73,39 @@ function generateModuleDependencyGraph(app: INestApplication): void {
...
 }
 
+function generateSwaggerSummaryJson(app: INestApplication<any>): void {
+  const options = new DocumentBuilder()
+    .setTitle(APP_TITLE)
+    .setVersion(APP_VERSION)
+    .build();
+  const openApiDocument = SwaggerModule.createDocument(app, options);
+
+  const minimalDocument: {
```
aberonni, line 83: Could we also add the parameters to this object?

aberonni (changes requested):
Small feedback, but all in all this is really cool 😄

services/121-service/src/main.ts
aberonni: Endpoint params*

services/121-service/src/main.ts
RubenGeo: Good one

services/121-service/src/main.ts
RubenGeo: Ah yes I missed as I only was using the 200 code. Good one

aberonni (approved):
Nice one! I would recommend sharing a post about this change in the 121 dev channel

PR #6214, opened 2024-12-03
portalicious: payment chart

aberonni:
[AB#31325](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31325)

### DIFF
--- a/interfaces/Portalicious/src/app/pages/project-payment/components/project-payment-chart/project-payment-chart.component.html
+++ b/interfaces/Portalicious/src/app/pages/project-payment/components/project-payment-chart/project-payment-chart.component.html
@@ -1,0 +1,7 @@
+<p-chart
+  type="bar"
+  [data]="chartData()"
+  [options]="chartOptions"
+  responsive
+  class="[&>div]:max-h-96"
+/>
--- a/interfaces/Portalicious/src/app/utils/chart.ts
+++ b/interfaces/Portalicious/src/app/utils/chart.ts
@@ -4,6 +4,9 @@ import ChartDataLabels from 'chartjs-plugin-datalabels';
 export function registerChartDefaults() {
   const documentStyle = getComputedStyle(document.documentElement);
 
+  Chart.defaults.font.family = 'Montserrat, sans-serif';
+  Chart.defaults.font.weight = 500;
+
   Chart.defaults.scale.ticks.color =
     documentStyle.getPropertyValue('--text-color');
 
--- a/services/121-service/src/payments/payments.service.ts
+++ b/services/121-service/src/payments/payments.service.ts
@@ -540,23 +540,7 @@ export class PaymentsService {
 
   public async getProgramPaymentsStatus(
     programId: number,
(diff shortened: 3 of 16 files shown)


interfaces/Portalicious/src/app/utils/chart.ts
```diff
@@ -4,6 +4,9 @@ import ChartDataLabels from 'chartjs-plugin-datalabels';
 export function registerChartDefaults() {
   const documentStyle = getComputedStyle(document.documentElement);
 
+  Chart.defaults.font.family = 'Montserrat, sans-serif';
```
elwinschmitz, line 7: There is no magical `var(--font-display)` Tailwind-utility, right?

Or like with the `--text-color` below?

Using the TailwindConfig as-is wouldn't really work unfortunately... :/

aberonni: No, I tried to pull it out but couldn't find a way to make it happen.

Worst case scenario the fonts in the chart are slightly off - feels like a fair risk to me given that we literally have one chart on the portal (for now).

interfaces/Portalicious/src/app/pages/project-payment/components/project-payment-chart/project-payment-chart.component.html
```diff
@@ -0,0 +1,7 @@
+<p-chart
+  type="bar"
+  [data]="chartData()"
+  [options]="chartOptions"
+  responsive
+  class="[&>div]:max-h-96"
+/>
```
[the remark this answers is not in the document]
aberonni: I didn't try this (nor did I see this in the docs). I just copied what I had done in my prototype (whichwas based on the docs at the time).

services/121-service/src/payments/payments.service.ts
```diff
@@ -540,23 +540,7 @@ export class PaymentsService {
...
   ): Promise<ProgramPaymentsStatusDto> {
-    if (payment) {
-      // TODO: refactor this once the "transaction/payment" entity exists,
-      // so that we can just get the status for a specific entity
-      const payments = await this.getPayments(programId);
-      const latestPayment = Math.max(...payments.map((p) => p.payment));
-      const isLatestPayment = payment === latestPayment;
-
-      if (!isLatestPayment) {
-        // only the latest payment can be in progress
```
aberonni, line 553: FYI I removed this endpoint after realising, with Jannis, that this statement is actually incorrect.

elwinschmitz: I think this wasn't used in the old Portal, right?

elwinschmitz (approved):
Assuming the 121-service change isn't a breaking change for the existing Portal: ✅

services/121-service/src/payments/payments.service.ts
aberonni: Yup, I created it for Portalicious in a separate PR.

PR #6247, opened 2024-12-05
Better error on creating users

RubenGeo:
[AB#31950](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31950) 

### DIFF
--- a/services/121-service/src/user/user.service.ts
+++ b/services/121-service/src/user/user.service.ts
@@ -299,7 +299,7 @@ export class UserService {
     const user = await qb.getOne();
 
     if (user) {
-      const errors = { username: 'Username must be unique.' };
+      const errors = { username: `Username: '${username}' must be unique.` };
       throw new HttpException(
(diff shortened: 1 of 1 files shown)


elwinschmitz (commented):
This endpoint fits within the "stricter rate-limits"-set, right?

Because with this error (also the original error), you could "find out" if certain accounts exist on the system already...

But having _no_ information on an error for the Cash IM users would also not be very helpful...

But still, in essence, this change doesn't really make the already existing 'vulnerability' any greater.. As you're only returning data that's been provided in the request. Its only "more convenient"...

elwinschmitz (approved):

RubenGeo:
Very point Elwin. But considering this is an Organiaztion admin only endpoint it's fine. As that user can already read all users in the system. If we make an open endpoint like this we should reconsider

elwinschmitz:
Ah, yes. Even better. ;)

PR #6253, opened 2024-12-06
Refactor.cascade delete

jannisvisser:
[AB#26365](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/26365)

## Describe your changes

See descriptions in devops tasks.

### DIFF
--- a/services/121-service/src/events/entities/event.entity.ts
+++ b/services/121-service/src/events/entities/event.entity.ts
@@ -16,7 +16,7 @@ import { UserEntity } from '@121-service/src/user/user.entity';
 
 @Entity('event')
 export class EventEntity extends Base121OptionalAuditedEntity {
-  @ManyToOne(() => UserEntity)
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE' })
   @JoinColumn({ name: 'userId' })
   public user: Relation<UserEntity>;
 
@@ -31,7 +31,7 @@ export class EventEntity extends Base121OptionalAuditedEntity {
   @Column({ type: 'character varying' })
   public type: EventEnum;
 
-  @ManyToOne(() => RegistrationEntity)
+  @ManyToOne(() => RegistrationEntity, { onDelete: 'CASCADE' })
   @JoinColumn({ name: 'registrationId' })
   public registration: Relation<RegistrationEntity>;
   @Column()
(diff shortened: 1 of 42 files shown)


jannisvisser:
failing Test Service: Code is because of https://github.com/global-121/121-platform/pull/6252 that needs to be merged

services/121-service/src/events/entities/event.entity.ts
```diff
@@ -16,7 +16,7 @@ import { UserEntity } from '@121-service/src/user/user.entity';
 
 @Entity('event')
 export class EventEntity extends Base121OptionalAuditedEntity {
-  @ManyToOne(() => UserEntity)
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE' })
```
RubenGeo, line 19: I am wondering a bit about this. If you delete a user would you also want to delete all events related to that user (so the audit trail of updating registrations, but also the transactions created by that user). I also found this in the user service at the delete endpoint:

```
    try {
      return await this.userRepository.remove(user);
    } catch (e) {
      if (e.code === '23503') {
        throw new HttpException(
          'User cannot be removed because it is related to other entities for logging purposes',
          HttpStatus.CONFLICT,
        );
      }
      throw e;
    }
```
So before we sort of consciously made a decision no to delete users? 

Maybe we should pick one of these options for users:
1. Make userId optional in transactions and events entities
2. Set the `active` boolean of users to false instead of deleting them. And maybe allowing updates to the username (so you can anomize the email adress)
3. Give extra warnings when deleting users with transactions?

RubenGeo (changes requested):
Maybe we need to add some testing on this? Like deleting the program to see if that keeps working?

Edit: I found that there was a task open for this

services/121-service/src/events/entities/event.entity.ts
jannisvisser: yeah good point. It was still the goal of [this to do task](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20144?workitem=31938) to go over relationships in detail again and notice things like this. 

Let's discuss indeed.

RubenGeo (approved):

PR #6285, opened 2024-12-13
AB#30601 Personal Info generic attributes refactor

RubenGeo:
[AB#30601](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30601) 

### DIFF
--- a/interfaces/Portalicious/src/app/domains/project/project.helper.ts
+++ b/interfaces/Portalicious/src/app/domains/project/project.helper.ts
@@ -1,73 +1,11 @@
 import { FinancialServiceProviderIntegrationType } from '@121-service/src/financial-service-providers/financial-service-provider-integration-type.enum';
-import { RegistrationAttributeTypes } from '@121-service/src/registration/enum/registration-attribute.enum';
-import { LocalizedString } from '@121-service/src/shared/types/localized-string.type';
 
-import { DataListItem } from '~/components/data-list/data-list.component';
 import { getFinancialServiceProviderSettingByName } from '~/domains/financial-service-provider/financial-service-provider.helper';
 import {
   FSPS_WITH_PHYSICAL_CARD_SUPPORT,
   FSPS_WITH_VOUCHER_SUPPORT,
 } from '~/domains/payment/payment.helpers';
-import {
-  AttributeWithTranslatedLabel,
-  Project,
-} from '~/domains/project/project.model';
-
-export const ATTRIBUTE_LABELS: Record<string, string | undefined> = {
-  fspDisplayName: $localize`:@@attribute-label-fspDisplayName:FSP Display Name`,
-  paymentAmountMultiplier: $localize`:@@attribute-label-paymentAmountMultiplier:Transfer value multiplier`,
-  maxPayments: $localize`:@@attribute-label-maxPayments:Max. payments`,
-  paymentCountRemaining: $localize`:@@attribute-label-paymentCountRemaining:Remaining payments`,
-};
-
-export const attributeToDataListItem = (
-  attribute: AttributeWithTranslatedLabel,
-  value: unknown,
-): DataListItem | undefined => {
-  const label = attribute.label;
-
-  switch (attribute.type) {
-    case RegistrationAttributeTypes.multiSelect:
-      // TODO: Implement multiSelect when necessary
-      console.log(
-        'attributeToDataListItem: multiSelect not implemented',
-        value,
-      );
-      return undefined;
-    case RegistrationAttributeTypes.numeric:
-      return {
-        label,
-        type: 'number',
-        value: value as number,
-      };
-    case RegistrationAttributeTypes.numericNullable:
-      return {
-        label,
-        type: 'number',
-        value: value as null | number,
-      };
-    case RegistrationAttributeTypes.date:
-      return {
-        label,
-        type: 'date',
-        value: value as Date,
-      };
-    case RegistrationAttributeTypes.boolean:
-      return {
-        label,
-        type: 'boolean',
-        value: value as boolean,
-      };
-    case RegistrationAttributeTypes.dropdown:
-    case RegistrationAttributeTypes.tel:
-    case RegistrationAttributeTypes.text:
-      return {
-        label,
-        type: 'text',
-        value: value as LocalizedString | string,
-      };
-  }
-};
+import { Project } from '~/domains/project/project.model';
 
 export function projectHasVoucherSupport(project?: Project) {
   return project?.programFinancialServiceProviderConfigurations.some((fsp) =>
@@ -87,6 +25,17 @@ export function projectHasFspWithExportFileIntegration(project?: Project) {
       getFinancialServiceProviderSettingByName(fsp.financialServiceProviderName)
         ?.integrationType === FinancialServiceProviderIntegrationType.csv,
   );
+}
+
+export function projectHasInclusionScore(project?: Project): boolean {
+  if (project?.programRegistrationAttributes) {
+    for (const attribute of project.programRegistrationAttributes) {
+      if (Object.keys(attribute.scoring).length > 0) {
+        return true;
+      }
+    }
+  }
+  return false;
 }
 
 export function financialServiceProviderConfigurationNamesHaveIntegrationType({
--- a/interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
+++ b/interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
@@ -1,6 +1,7 @@
 import { ActivityTypeEnum } from '@121-service/src/activities/enum/activity-type.enum';
 import { VisaCard121Status } from '@121-service/src/payments/fsp-integration/intersolve-visa/enums/wallet-status-121.enum';
 import { RegistrationStatusEnum } from '@121-service/src/registration/enum/registration-status.enum';
+import { LanguageEnum } from '@121-service/src/shared/enum/language.enums';
 
 import { AppRoutes } from '~/app.routes';
 import { ChipVariant } from '~/components/colored-chip/colored-chip.component';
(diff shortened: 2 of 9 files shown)


interfaces/Portalicious/src/app/domains/project/project.helper.ts
```diff
@@ -13,14 +16,62 @@ import {
...
+  paymentCount: $localize`:@@attribute-label-paymentCount:Payment Count`,
   paymentCountRemaining: $localize`:@@attribute-label-paymentCountRemaining:Remaining payments`,
+  scope: $localize`:@@attribute-label-scope:Scope`,
+  status: $localize`:@@attribute-label-status:Status`,
+  registrationProgramId: $localize`:@@attribute-label-registrationProgramId:Registration ID`,
+  registrationCreatedDate: $localize`:@@attribute-label-registrationCreatedDate:Registration Created Date`,
+  inclusionScore: $localize`:@@attribute-label-inclusionScore:Inclusion Score`,
+};
+
+export const getGenericAttributeDataListItem = (
```
RubenGeo, line 36: We (me and @aberonni ) had an open ended discussion if the logic of this function should live should live in the backend or in the frontend. 

At first I came to a conclusion to add it to the backend in [/api/programs/{programId}/attributes](http://localhost:3000/docs/#/programs/ProgramController_getAttributes) under another flag like `includeTemplateDefaultAttributes` and map the typings there. However I thought later it does not make sense to categorize the `GenericRegistrationAttributes` using the `RegistrationAttributeTypes` as this enum is meant for categorizing `ProgramRegistrationAttributeEntities`. For example `programFinancialServiceProviderConfigurationLabel` does not fall in any category and preferredLanguage and programFinancialServiceProviderConfigurationName are sort of dropdown but the do not work the same as the 'normal' dropdown attributes.

The major reason of setting the RegistrationAttributeTypes in the backend is so that it could be used in `projectAttributeToDataListItem` to map it to the type of a `DataListItem (a frontend) conecpt. I therefore decided to create this function `getGenericAttributeDataListItem` in the frontend directly.

For the upcoming work of edit attribute data on the profile page I think that a lot of the editable GenericRegistrationAttributes are sort of `exceptions` that need their own logic (like programFspConfig or the preferredLanguage). So the may also not need the RegistrationAttributeTypes as parameter of they should be edited in the frontend

interfaces/Portalicious/src/app/domains/project/project.helper.ts
```diff
@@ -89,6 +140,17 @@ export function projectHasFspWithExportFileIntegration(project?: Project) {
...
 
+export function projectHasInclusionScore(project?: Project): boolean {
+  if (project?.programRegistrationAttributes) {
+    for (const attribute of project.programRegistrationAttributes) {
+      if (Object.keys(attribute.scoring).length > 0) {
+        return true;
+      }
+    }
+  }
+  return false;
```
aberonni, line 151:
```suggestion
  return (project?.programRegistrationAttributes ?? []).some(attribute => Object.keys(attribute.scoring).length > 0));
```

This is how I would've written it, but I don't care too much. Mostly curious what you prefer.

aberonni: Don't be scared!

interfaces/Portalicious/src/app/pages/project-registration-personal-information/project-registration-personal-information.page.ts
```diff
@@ -62,20 +69,52 @@ export class ProjectRegistrationPersonalInformationPageComponent {
...
 
+  project = injectQuery(this.projectApiService.getProject(this.projectId));
+
   attributeList = computed<DataListItem[]>(() => {
     const list: DataListItem[] = [];
     if (!this.projectAttributes.isSuccess() || !this.registration.isSuccess()) {
       return list;
     }
 
+    const genericAttributeNames =
```
RubenGeo, line 80: A side effect of getting the genericAttributeNames seperate is that they always appears at the start of the data list now. Which I think is nice.

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
```diff
@@ -53,6 +54,18 @@ export const REGISTRATION_STATUS_VERB: Record<RegistrationStatusEnum, string> =
...
   };
 
+export const LANGUAGE_ISO_639_1: Record<LanguageEnum, string> = {
+  ar: $localize`Arabic`,
+  en: $localize`English`,
+  es: $localize`Spanish`,
+  fr: $localize`French`,
+  in: $localize`Indonesian`,
+  nl: $localize`Dutch`,
+  pt_BR: $localize`Portuguese`,
```
elwinschmitz, line 64: `<hat mode="pedantic">`
This one includes a 'region subtag', which is part of https://en.wikipedia.org/wiki/IETF_language_tag
`</hat>`

So the variable-name is a bit _too specific_; Wouldn't `LanguageName` or `LanguageLabel` work?

interfaces/Portalicious/src/locale/messages.nl.xlf
```diff
@@ -910,9 +910,9 @@
         <source>Add new user</source>
         <target>Nieuwe gebruiker toevoegen</target>
       </trans-unit>
-      <trans-unit id="attribute-label-fspDisplayName" datatype="html">
-        <source>FSP Display Name</source>
-        <target>FSP Weergavenaam</target>
+      <trans-unit id="attribute-label-programFinancialServiceProviderConfigurationLabel" datatype="html">
+        <source>FSP</source>
+        <target state="new">FSP Weergavenaam</target>
```
elwinschmitz, line 915:
```suggestion
        <target>FSP</target>
```

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
RubenGeo: I though it would be nice to have a name like this so if we ever add a new language, we don't just make something up and look a the standard. 

What do you think about using a specific const name? And what would be the right specific const name?

interfaces/Portalicious/src/app/pages/project-registration-personal-information/project-registration-personal-information.page.ts
```diff
@@ -62,20 +69,52 @@ export class ProjectRegistrationPersonalInformationPageComponent {
...
+      const dataListItem = projectAttributeToDataListItem(attribute, value);
       if (dataListItem) {
         list.push(dataListItem);
       }
     }
-
     return list;
   });
+
+  private getGenericAttributeNamesForPersonalInformation(): GenericRegistrationAttributes[] {
```
aberonni, line 103:
```suggestion
  private genericAttributeNamesForPersonalInformation = computed<GenericRegistrationAttributes[][]>(() => {
```

Watch out @RubenGeo @elwinschmitz ! Not following this pattern can lead to (hard to catch) bugs.

For additional context if needed: you are calling `getGenericAttributeNamesForPersonalInformation` from a `computed` variable (`attributeList`). That means that you would want `attributeList` to automatically update whenever `getGenericAttributeNamesForPersonalInformation` changes. If you do not make `getGenericAttributeNamesForPersonalInformation` a `computed` variable, then this will not happen, because angular will not know to react to changes within `getGenericAttributeNamesForPersonalInformation` itself (changes that could happen, for example, whenever the contents of `this.project.data()` changes).

interfaces/Portalicious/src/app/domains/project/project.helper.ts
```diff
@@ -12,15 +16,76 @@ import {
   AttributeWithTranslatedLabel,
   Project,
 } from '~/domains/project/project.model';
+import { LANGUAGE_ISO_639_1 } from '~/domains/registration/registration.helper';
 
-export const ATTRIBUTE_LABELS: Record<string, string | undefined> = {
-  fspDisplayName: $localize`:@@attribute-label-fspDisplayName:FSP Display Name`,
+export const ATTRIBUTE_LABELS: Record<GenericRegistrationAttributes, string> = {
```
aberonni, line 21: This file is getting a bit crowded. I wonder if we should have a dedicated `attribute` domain / helper - but not sure if that makes also from a BE perspective.

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
aberonni: +1 to what Elwin suggests. My two cents on the name itself: `LanguageEnumLabel`.

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
```diff
@@ -53,6 +54,18 @@ export const REGISTRATION_STATUS_VERB: Record<RegistrationStatusEnum, string> =
     [RegistrationStatusEnum.deleted]: $localize`Delete`,
   };
 
+export const LANGUAGE_ISO_639_1: Record<LanguageEnum, string> = {
```
aberonni, line 57: I don't think this should live in this file. Maybe `~/utils/locale`?

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
RubenGeo: If I put it in the 121-service it is angry about $localize. I there for just added your iso comment to https://github.com/global-121/121-platform/blob/main/services/121-service/src/shared/enum/language.enums.ts#L1 and change the name of the const to `LANGUAGE_ENUM_LABEL`

interfaces/Portalicious/src/app/domains/project/project.helper.ts
RubenGeo: After a discussion with Peter and Elwin we decided to do 

domains/project/project-attribute.helper.ts

interfaces/Portalicious/src/app/domains/registration/registration.helper.ts
aberonni: I see. If that is solely how `LanguageEnum` is used, then maybe a clearer name for it would be `PreferredLanguageEnum` - but maybe out of scope 😄 

For now this makes sense to me.

elwinschmitz (approved):

PR #6303, opened 2024-12-18
test: unit test PoC for isTokenValid 

jannisvisser:
[AB#31601](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/31601)


package.json
```diff
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@prettier/plugin-xml": "^3.4.1",
     "@types/multer": "^1.4.12",
+    "@types/jest": "^29.5.14",
```
diderikvw, line 46: Why was this not here before? Do we need to add it?

services/121-service/src/utils/token/token-validation.service.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@nestjs/common';
+import { TokenSet } from 'openid-client';
+
+@Injectable()
+export class TokenValidationService {
+  public isTokenValid(
+    tokenSet: TokenSet,
+    expiresAtInSeconds = false, // If true, then expires_at is assumed in milliseconds, as it should be
```
diderikvw, line 8: This is very confusing: so when expiresAtInSeconds = true then it is in milliseconds??
Why "as it should be"?

I suggest that we put a bit more effort into this and solve it to be clean and simple. Ideas:
1. Create our own TokenSet sub-class with at least a comment that expires_at is "in milliseconds". But what does that even mean? A timestamp according to standard blabla? And if we can, add a validation that if someone passes in seconds instead of milliseconds, it starts to complain. Like: this timestamp is likely false, because the date is in the 18th century.
2. Get rid of this boolean parameter in this function.
3. Refactor IntersolveVisa code so that it also works in milliseconds.

services/121-service/src/utils/token/token-validation.service.ts
diderikvw, line 1: I think this is a very cool refactor.

services/121-service/src/utils/token/token-validation.service.ts
```diff
@@ -0,0 +1,23 @@
...
+    if (!tokenSet || !tokenSet.expires_at) {
+      return false;
+    }
+
+    // convert to milliseconds if needed
+    if (expiresAtInSeconds) {
+      tokenSet.expires_at *= 1000;
+    }
+
+    const timeLeftBeforeExpire = tokenSet.expires_at - Date.now();
```
diderikvw, line 19: What is not clear from the name of this function is that it looks at expires_at to determine if the token is still valid (it could also look at other params of TokenSet to determine this) and it is also not clear from the name of this function it looks if the token is still valid for at least 5 minutes.

Ideally we should name the function so that it is clearer what it does. Or else, GEFN, put a comment above the function with a brief explainer. 

See e.g. comment above IntersolveVisaService.retrieveAndUpdateWallet as an example of how we do function comments. In this way VS Code also picks it up and shows it when hovering.

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.ts
```diff
@@ -35,6 +46,7 @@ export class SafaricomApiService {
 
     let errorMessage: string | undefined;
 
+    // ##TODO: unit test this part (and move to helper)
```
diderikvw, line 49: Look at solution in IntersolveVisaApiService. If a similar solution can be applied here (with the customized http service), then consider refactoring that out into a Helper utils class, in the same way as the TokenValidationService.

services/121-service/src/payments/fsp-integration/safaricom/safaricom.service.ts
```diff
@@ -34,6 +34,7 @@ export class SafaricomService
    * Do not use! This function was previously used to send payments.
    * It has been deprecated and should not be called anymore.
    */
+  // ##TODO: remove this method. What is the point of keeping it?
```
diderikvw, line 37: Remove comment. We need to keep it for now because of "implements FinancialServiceProviderIntegrationInterface". When we refactor the PaymentsService, we can remove this.

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    safaricomApiService = module.get<SafaricomApiService>(SafaricomApiService);
+    customHttpService = module.get<CustomHttpService>(CustomHttpService);
+
+    (customHttpService.post as jest.Mock).mockResolvedValue(
+      mockPostTransferResponse,
+    );
+  });
+
+  describe('transfer', () => {
```
diderikvw, line 53: Is this "happy/typical path" that would be covered in Integration Test as well, and therefore can be left out of unit testing, depending on effort vs. added value?

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      // Act
+      const transferResult = await safaricomApiService.transfer(transferInput);
+
+      // Assert
+      expect(customHttpService.get).not.toHaveBeenCalled();
+      expect(customHttpService.post).toHaveBeenCalled();
+      expect(transferResult).toBeDefined();
+    });
+
+    it('should return success response if all API-calls succeed', async () => {
```
diderikvw, line 70: I do not understand from the description how this test is different from the first one. It should succeed and authenticate if tokenSet not valid? Also a happy/typical path covered in Integration Test? See comment above.

The Integration Test should also cover incorrect credentials scenario (maybe it does, have not checked it at this moment).

diderikvw (changes requested):

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.spec.ts
jannisvisser: change order + update 'it'. this happy path is also integration-tested. leave it out?

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.spec.ts
jannisvisser: proposal is to remove these unit test and instead create extra integration test which calls payment endpoint twice, where checking that the 2nd time the outgoing GET authenticate call is not made. Such an assertion is new in our api-test portfolio, check with others if that is feasible first.

services/121-service/src/utils/token/token-validation.service.ts
jannisvisser: I opted for option 3, which seems quite feasible in the end. That way this parameter can be dropped completely.

services/121-service/src/utils/token/token-validation.service.ts
jannisvisser: comment added

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.ts
jannisvisser: - I separated this to api helper service and wrote unit test
- it is not similar enough to the visa method to refactor into shared code
- given that I created the api helper service I also moved 'createTransferPayload' into there, which seems to make sense. Although you could then argue it's confusing why that public method has no unit test now.

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.spec.ts
jannisvisser: talked about this with @RubenGeo and he agrees that api test does not serve this purpose well, as it's oriented at input and output of our api endpoints, and not on what happens inside. Instead look at options again to unit test this.

aberonni: I am missing some context to give a proper opinion on this, and am struggling to garner it just from reading the code. But perhaps it's ok if I don't share my opinion on this one and avoid "too many cooks spoiling the broth". If you really feel like my input could be valuable, feel free to put something in my calendar so maybe we can talk through it.

jannisvisser: @diderikvw as already mentioned above, I (and @RubenGeo) really don't think we should switch this to an API-test. Our API tests are really meant to assert based on input and out put of (a flow of) our API-endpoints, and not on what happens inside the black box. 

Conversely, a unit test serves this purpose really well. It's some specific logic (if tokenSet present skip, otherwise authenticate) that you can easily simulate to test different scenarios quickly, and you can easily assert whatever you want on in-between steps (in this case, the outgoing GET call being made or not). I think it's worth the small "hack" I implemented now to make this work. Namely, I had to make the tokenSet private instead of public. To cover for that I created public method setTokenSetForUnitTest(). I'm not overly happy with it, but I think it's fine. 

What do you think?

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.helper.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    originatorConversationId: string,
+  ): string | undefined {
+    if (!transferResponse || !transferResponse.data) {
+      return `Error: No response data from Safaricom API`;
+    } else if (transferResponse.data.errorCode) {
+      if (transferResponse.data.errorCode === '500.002.1001') {
+        // This happens only in case of unintended Redis job re-attempt, and only if the API-request already went through the first time
+        // Return custom error, as it should be handled differently than other errors
+        const duplicateOriginatorConversationIdErrorMessage = `Error: ${transferResponse.data.errorMessage} for originatorConversationId ${originatorConversationId}`;
+        throw new DuplicateOriginatorConversationIdError(
```
diderikvw, line 46: I do not like how a helper function that should just format the text of an error message is throwing an error sometimes. But since this is quite deep inside the Safaricom module and encapsulated, let's leave it for now.

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.ts
```diff
@@ -4,67 +4,59 @@ import { TokenSet } from 'openid-client';
...
 
-    // All the checks above mean that at this stage transferResponse.data.ResponseCode === '0'
     return {
       mpesaConversationId: transferResponse.data.ConversationID,
     };
   }
 
   private async authenticate(): Promise<void> {
-    if (this.isTokenValid(this.tokenSet)) {
+    if (this.tokenValidationService.isTokenValid(this.tokenSet)) {
```
diderikvw, line 59: Could we instead call the transfer function twice in the unit test? So the first time to make sure the token is valid and the second time to do the real test? I am sure you considered this, so maybe I am missing a caveat here.

(this is also more like what happens in production: first time token created, then not anymore until expired)

services/121-service/src/payments/fsp-integration/intersolve-visa/intersolve-visa.api.service.ts
```diff
@@ -72,24 +76,14 @@ export class IntersolveVisaApiService {
     const tokenSet = await client.grant({
       grant_type: 'client_credentials',
     });
+    if (tokenSet.expires_at) {
+      tokenSet.expires_at *= 1000; // Convert to milliseconds to align with standards in tokenValidationService.isTokenValid()
```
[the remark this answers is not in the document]
jannisvisser: I'm not sure, but it is what's returned by this client.grant() method. It is beyond the scope of unit test spike to look at this I think, but feel free to add a TODO comment to look into this further.

services/121-service/src/payments/fsp-integration/safaricom/services/safaricom.api.service.ts
jannisvisser: For me this feels not how a unit test should work. A unit test should test a specific method with some input and some expected output. It should not initiate a flow. This relates back to [this unresolved comment ](https://github.com/global-121/121-platform/pull/6303#discussion_r1901876289) so let's continue the convo there and resolve this one.

jannisvisser: @RubenGeo see above discussion. After my last comment, @diderikvw and I discussed in a call that if we frame it such that the first call of the transfer method is part of the Arrange step, and the second is part of the Act step, then it might be OK. I coded it now as such. See [this commit](https://github.com/global-121/121-platform/pull/6303/commits/c024c3a7a037b53a8f32ec0b69b7f87b04b422cd). We thought it be good to get your opinin as well though.

RubenGeo: @jannisvisser Sounds like a good solution. Definitely better than covering this with an API test

PR #6304, opened 2024-12-18
Change payment card content

tijsziere:
[AB#32349](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/32349) 

## Describe your changes

As per codesign in Kenya


aberonni:
@tijsziere converting this to draft while the PR checklist in the description is addressed.

Some tips:
- You need to create a relevant task somewhere on the board so we can keep track of where this came from if we need to. Then link that task in the PR description.
- You need to export new localization strings. Hopefully there should be documentation for this in the portalicious codebase. If not, let me know (also in a DM is fine).

The above should make the CI checks pass. Then you can move to ready for review and share the task in review status with the team.

I haven't actually looked at the code yet, given that the PR isn't technically ready for review.

(I'm being pedantic, but in the interest of setting you up for smoother contributions in the future!)

aberonni:
FYI for anyone finding this: PR is parked while Tijs waits for input from design

PR #6312, opened 2024-12-19
portalicious: fix transfer value in payment creation AB#32309

arsforza:
[AB#32309](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/32309)

## Describe your changes
Contrary to what is stated in the DevOps item, the total amount for 1024 PAs is correct, since the single transfer amount is 17.50 EUR and not 18.
Nevertheless, the total amount was formatted without decimal digits so, in case of a payment for an odd number of PAs, it was rounded up to the nearest integer, causing confusion.
I've added decimal digits only if the total is not an integer.


interfaces/Portalicious/src/app/pages/project-payments/components/create-payment/create-payment.component.ts
```diff
@@ -286,6 +286,11 @@ export class CreatePaymentComponent {
       return [];
     }
 
+    const totalPaymentAmount =
+      this.paymentAmount() * dryRunResult.sumPaymentAmountMultiplier;
+    const totalPaymentAmountFormat =
+      totalPaymentAmount % 1 === 0 ? '1.0-0' : '1.2-2';
```
aberonni, line 292: I actually think it's a bug that we're not showing decimal digits - bit of a miss from us. This should probably be the default in most places, except for wherever we show "aggregates" / big numbers (like on the payment metrics on the all payments page).

My recommendation would be:
 - use `1.2-2` always here
 - look for `1.0-0` elsewhere and check whether we are using it in a place that could lead to similar confusion

arsforza: I thought of using `1.2-2` always but I also thought 1.0-0 would look better for design reasons.
I'll change it.

elwinschmitz, line 292: @arsforza Would it also work with just the format: `1.0-2`

So "minimally 0 decimal-digits, maximum 2", then it would be doing the "check if necessary to show `00` for you?

aberonni: For "precise money values" which is a term I just came up with, I actually think it would be wise to always show two decimal digits regardless (to avoid any confusion for the user).

elwinschmitz: Yeah, I would also be all for "more explicit, less wrong" :D

But I hope these things where already considered in the design and the "shortest number possible with least confusing zeros" was the goal here.

arsforza: @elwinschmitz `1.0-2` shows `.5` instead of `.50`

aberonni: > But I hope these things where already considered in the design

Maybe wishful thinking, maybe design was ignored (at least I am not sure that I ever took the design as leading when copy-pasting the currency pipe code 🙈 )

aberonni (approved):

PR #6331, opened 2024-12-23
Api tests for update contact info

RubenGeo:
[AB#32342](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/32342) 

### DIFF
--- a/services/121-service/test/visa-card/contact-information-visa-card.test.ts
+++ b/services/121-service/test/visa-card/contact-information-visa-card.test.ts
@@ -15,7 +15,7 @@ import {
   resetDB,
 } from '@121-service/test/helpers/utility.helper';
 
-describe('Should update contact information', () => {
+describe('Update registration data of registration with visa customer', () => {
   let accessToken: string;
 
   beforeEach(async () => {
@@ -24,7 +24,7 @@ describe('Should update contact information', () => {
     await waitFor(2_000);
   });
 
-  it('should update contact information', async () => {
+  it('should update phone number before and after changing FSP configuration from Visa', async () => {
     // Arrange
     await seedPaidRegistrations([registrationVisa], programIdVisa);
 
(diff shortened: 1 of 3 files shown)


services/121-service/test/visa-card/contact-information-visa-card.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      {
+        phoneNumber: '987654321',
+      },
+      'test',
+      accessToken,
+    );
+
+    // Assert
+    expect(responseUpdateBeforeFspChange.status).toBe(HttpStatus.OK);
+    expect(responseUpdateAfterFspChange.status).toBe(HttpStatus.OK);
```
jannisvisser, line 65: @RubenGeo Can you explain how this setup and assertions test what you want to test? I don't really get it.

RubenGeo: I updated the labels of describe and it because I found the previous ones inadequate.

This API test was created in response to a bug where we posted:
image
to Intersolve after a user updated their registration data.

In this PR, I modified our mock service to throw errors when the update contact info endpoints are called with undefined or null. This ensures that the assertions will produce a status code other than 200 if we attempt to call the Intersolve endpoint with undefined values or the string "undefined".

This approach is somewhat suboptimal because it primarily tests whether the PATCH registration endpoint does not return an error, rather than directly testing whether an API call is made to Intersolve. Unfortunately, testing this in an API test is challenging (and, to be fair, we also don’t test this in our do-payment API tests).

I also considered creating a unit test. However, testing this as a “unit” is complicated due to various dependencies, such as whether a registration has an Intersolve customer, what registration data is available, and the FSP configuration associated with the registration. For these reasons, I chose this solution instead. @jannisvisser

jannisvisser (approved):

PR #6390, opened 2025-01-16
Change error checks in K6 tests

Piotrk39:
[AB#32606](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/32606) 

## Describe your changes

Due to the bug that caused the message from K6 runs to be always "passed" the tests additionally need better error handling where more checks are displayed and validated through the run of load/ performance tests.


.github/workflows/test_k6_cronjob.yml
```diff
@@ -49,6 +49,10 @@ jobs:
         if: always()
         uses: jwalton/gh-docker-logs@v2
 
+      - name: Print job status
```
RubenGeo, line 52: This part is left over testing part that came from my branch. Something probably went wrong with a rebase later. This part can be removed

k6/tests/getProgramWithManyAttributes.js
```diff
@@ -18,22 +19,33 @@ const programId = 2;
...
   },
   vus: 1,
-  duration: '20m',
+  duration: '30s',
   iterations: 1,
 };
 
+const failedChecks = new Counter('failed_checks');
+
+function checkAndFail(response, checks) {
```
RubenGeo, line 31: Are you sure this checkAndFail function is needed?  In the run-all-tests.sh we have a line that looks at the checks that failed:
```
if [ $(jq '.metrics.checks.fails' summary.json) -gt 0 ]; then
```
so I think a `check()` that fails may already be sufficient to trigger the teams message. And the fail() is not needed

RubenGeo (changes requested):

k6/tests/getProgramWithManyAttributes.js
Piotrk39: I will update it and test.

Piotrk39:
@RubenGeo  All tests should be passing now as expected. There is one more issue connected with the K6 tests and it is that they do not fail after threshold of requests is breached e.g if we have threshold for failed_requests set to 30 percent after this is breached then we still can see pass in the outcome of the test. For now we should only rely on the failed checks.

RubenGeo (dismissed):

RubenGeo (approved):

PR #6400, opened 2025-01-20
portalicious: store selected columns

PeterSmallenbroek:
AB#32583 


interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -146,6 +146,11 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
   readonly contextMenu = viewChild<Menu>('contextMenu');
   readonly extraOptionsMenu = viewChild<Menu>('extraOptionsMenu');
 
+  selectedColumnsStateKey = computed(() => {
+    const key = this.localStorageKey();
+    return key ? `${key}-selected-columns` : 'selected-columns';
```
aberonni, line 151:
```suggestion
    return key ? `${key}-selected-columns` : undefined;
```

I think if we don't have a key to save the filters, then we shouldn't save the selected columns either.

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,10 +489,31 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
...
+      this.selectedColumnsStateKey(),
     );
+    if (storedSelectedColumns) {
+      const interpretedColumns = JSON.parse(
+        storedSelectedColumns,
+      ) as QueryTableColumn<TData>[];
+      console.log(
+        '🚀 ~ QueryTableComponent<TData ~ resetColumnVisibility ~ interpretedColumns:',
+        interpretedColumns,
+      );
```
aberonni, line 507:
```suggestion
```

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,10 +489,31 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
...
+      const matchedColumns = interpretedColumns
+        .map((column) => this.columns().find((c) => c.field === column.field))
+        .filter(Boolean) as QueryTableColumn<TData>[];
```
aberonni, line 510:
```suggestion
        .filter((column) => !!this.columns().find((c) => c.field === column.field))  as QueryTableColumn<TData>[];
```

I don't think we need a double loop here

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,10 +489,31 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
    */
   visibleColumns = model<QueryTableColumn<TData>[]>([]);
 
-  resetColumnVisibility() {
-    this.visibleColumns.set(
-      this.columns().filter((column) => !column.defaultHidden),
+  resetColumnVisibility(removeFromLocalStorage = false) {
```
aberonni, line 492: I'm a bit confused about `removeFromLocalStorage` and when it is used vs when it is not used. I could look at the code to understand better, but maybe this is a sign that the variable could be named slightly better instead. I imagine this `removeFromLocalStorage` is something like `hard reset` vs `soft reset` or `reset to saved preferences` vs `reset to default`.

In general, I would expect the reset button to always reset to default, and never to reset to what was saved in local storage - but I imagine that is the current behaviour, even though the variable name makes me question that.

aberonni (changes requested):

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
PeterSmallenbroek: I agree with the naming. And yes, it already does the behaviour you mention. I'll explain it a little more: While testing I found that the effect is being run on the loading of the component. And (of course) when the 'Revert to default' button is clicked. The problem is in the first run; Which then removes the localstorage entry (if you didn't have that flag).
I think we can rename it to something like 'revertToDefault'

aberonni: Or maybe it should be two different functions? `resetColumnVisibilityToDefault` vs `restoreColumnVisibilityFromLocalStorage`

PeterSmallenbroek: I've refactored this part of the code to make it more readable and single responsibility'ish. If you still really want to split it in 2 functions then let's figure that out together because I couldn't on my own.

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,9 +489,43 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
...
+
+    return JSON.parse(storedColumns) as QueryTableColumn<TData>[];
+  }
+
+  private getMatchingColumns(
+    storedColumns: QueryTableColumn<TData>[],
+  ): QueryTableColumn<TData>[] {
+    return storedColumns
+      .map((column) => this.columns().find((c) => c.field === column.field))
+      .filter(Boolean) as QueryTableColumn<TData>[];
```
aberonni, line 504:
```suggestion
.filter((column) => !!this.columns().find((c) => c.field === column.field));
```

The double loop made its way back again :D

Also I don't think the ` as ` is necessary here.

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,9 +489,43 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
...
+  }
+
+  private getDefaultColumns(): QueryTableColumn<TData>[] {
```
aberonni, line 507:
```suggestion
  private defaultColumns = computed<QueryTableColumn<TData>[]>(() => {
```

Not a necessary change but this feels like it would be more in line with how we normally name / treat computed properties like this one.

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
```diff
@@ -484,9 +489,43 @@ export class QueryTableComponent<TData extends { id: PropertyKey }, TContext> {
...
+    return this.columns().filter((column) => !column.defaultHidden);
+  }
+
+  resetColumnVisibility(revertToDefault = false): void {
```
aberonni, line 511: Yeah this function is way better indeed. I think the only thing is maybe I would rename it now
```suggestion
  updateColumnVisibility(revertToDefault = false): void {
```

aberonni (changes requested):
Much nicer!

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
aberonni: Ahhh okok but then maybe?

```suggestion
      .map((column) => this.columns().find((c) => c.field === column.field))
      .filter(column => column !== undefined)
```

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
PeterSmallenbroek: I think a computed only has benefits when it's used in the html. So that's why I chose it to be a regular method. Or am I missing something here?

aberonni: Yeah, this wouldn't necessarily have _benefits_ from a technical / performance aspect or anything similar. It's more about avoiding introducing a "getter" pattern for something that can be a `computed`, and this being picked up elsewhere

aberonni (dismissed):

interfaces/Portalicious/src/app/components/query-table/query-table.component.ts
aberonni: I'll let you decide if you wanna change or if you think I'm overthinking things (which I might be 😄 )

aberonni (approved):

PR #6430, opened 2025-01-28
Fix allow empty string values for nonRequired dropdows

RubenGeo:
[AB#33084](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33084) [AB#33082](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33082) 

## Describe your changes

See github comments in the code 

### DIFF
--- a/services/121-service/src/registration/validators/registrations-input-validator.ts
+++ b/services/121-service/src/registration/validators/registrations-input-validator.ts
@@ -260,7 +260,7 @@ export class RegistrationsInputValidator {
             if (row[att.name] === undefined) {
               return;
             }
-            if (row[att.name] == null) {
+            if (row[att.name] == null || row[att.name] === '') {
               validatedRegistrationInput.data[att.name] = null;
               return;
             }
--- a/services/121-service/test/registrations/bulk-update-registration.test.ts
+++ b/services/121-service/test/registrations/bulk-update-registration.test.ts
@@ -1,6 +1,5 @@
 import { SeedScript } from '@121-service/src/scripts/enum/seed-script.enum';
 import { waitFor } from '@121-service/src/utils/waitFor.helper';
-import { assertRegistrationBulkUpdate } from '@121-service/test/helpers/assert.helper';
 import { patchProgram } from '@121-service/test/helpers/program.helper';
 import {
   bulkUpdateRegistrationsCSV,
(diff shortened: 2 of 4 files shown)


services/121-service/src/registration/validators/registrations-input-validator.ts
```diff
@@ -260,7 +260,7 @@ export class RegistrationsInputValidator {
             if (row[att.name] === undefined) {
               return;
             }
-            if (row[att.name] == null) {
+            if (row[att.name] == null || row[att.name] === '') {
```
RubenGeo, line 263: Fix explanation:

When importing a CSV, all empty cells in a column are interpreted as empty strings (''). Before this fix, the code would attempt to check if an empty string was a valid dropdown option when processing a CSV with empty cells for non-required dropdown questions.

This fix ensures that empty strings and null values are treated the same way. If a cell is empty (i.e., an empty string), no registrationAttributeData will be stored.

Additionally, this affects attributes of type text, where empty strings will no longer be stored. If an empty string is posted for such attributes, no data will be saved.

Note:
This does not impact existing data. Our code already handles cases where registrationAttributeData contains empty strings or does not exist in a similar manner. So I don't think it really matters to some registrationAttributeData that store an empty string will persist

services/121-service/test/registrations/bulk-update-registration.test.ts
```diff
@@ -96,16 +111,25 @@ describe('Update attribute of multiple PAs via Bulk update', () => {
...
+
+    // Explicit assertions for pa2 using patch data
+    expect(pa2AfterPatch).toMatchObject(registrationDataThatWillChangePa2);
+
+    // Ensure unchanged fields remain the same
+    const dataThatStaysTheSamePa1 = filterUnchangedProperties(
       pa1BeforePatch,
+      registrationDataThatWillChangePa1,
     );
-    assertRegistrationBulkUpdate(
```
RubenGeo, line 104: I tried to make the assertions in these tests a bit clearer by removing assertRegistrationBulkUpdate and using toMatchObject instead. The advantages of this approach are:
- You can more easily see the line where an assertion went wrong on a failed test. As it does not point to the assertion helper.
- You can see the key for which the assertion went wrong. In the attached screenshot it's hard to tell if the assertion went wrong for 'phonenumber', 'name' or whatever other property.

image

services/121-service/test/registrations/bulk-update-registration.test.ts
```diff
@@ -12,6 +11,22 @@ import {
   resetDB,
 } from '@121-service/test/helpers/utility.helper';
 
+function filterUnchangedProperties(
```
RubenGeo, line 14: I added this function at the top of the test file. I am not sure if it should move to one of the helper files, or what our practice should be here. I expect that this helper function will only ever be use in this files, which argues for keeping it here, but we usually do not have seperate fuctions inside .test.ts files

PeterSmallenbroek (approved):

PR #6464, opened 2025-02-04
Events: small changes and refactor todo's

diderikvw:
[AB#33268](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33268) 

## Describe your changes

While reviewing code I saw some things in the EventsService public interface that could be cleaner.

### DIFF
--- a/services/121-service/src/events/events.controller.ts
+++ b/services/121-service/src/events/events.controller.ts
@@ -27,7 +27,6 @@ import { ExportFileFormat } from '@121-service/src/metrics/enum/export-file-form
 import { ScopedUserRequest } from '@121-service/src/shared/scoped-user-request';
 import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
 import { UserService } from '@121-service/src/user/user.service';
-import { RequestHelper } from '@121-service/src/utils/request-helper/request-helper.helper';
 import { sendXlsxReponse } from '@121-service/src/utils/send-xlsx-response';
 
 @UseGuards(AuthenticatedUserGuard)
@@ -62,7 +61,7 @@ export class EventsController {
     required: false,
     enum: ExportFileFormat,
     description:
-      'Format to return the data in. Options are "json" and "xlsx". Defaults to "json" if not specified. If "xlsx" is selected, the response will be a file download in which the data is slightly differently formatted for portal users in ',
+      'Format to return the data in. Options are "json" and "xlsx". Defaults to "json" if not specified. If "xlsx" is selected, the response will be a file download in which the data is slightly differently formatted for portal users.',
   })
   @Get('programs/:programId/events')
   public async getEvents(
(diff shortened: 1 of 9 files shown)


services/121-service/src/events/events.controller.ts
```diff
@@ -91,19 +91,20 @@ export class EventsController {
...
         searchOptions,
-      );
+      });
       if (result.length === 0) {
         throw new HttpException({ errors: errorNoData }, HttpStatus.NOT_FOUND);
       }
       return sendXlsxReponse(result, filename, res);
     }
-    const result = await this.eventService.getEventsJsonDto(
+    // ##TODO: Why only check for format xlsx if the user has access to this program and not for json as well? If json is not used at the moment: easiest to just remove it here?
```
RubenGeo, line 103: It's used in the API tests (getEvents in registration helper). Mainly because it's more convienent to test with JSON than csv

services/121-service/src/events/events.service.ts
```diff
@@ -20,8 +20,11 @@ import { ScopedUserRequest } from '@121-service/src/shared/scoped-user-request';
 import { UserService } from '@121-service/src/user/user.service';
 import { UserType } from '@121-service/src/user/user-type-enum';
 
+// TODO: REFACTOR: Turn this into an interface instead of a type, for simplicity, and think of a better name than "LogEntity".
 type LogEntity = Partial<RegistrationViewEntity> & {
+  // TODO: REFACTOR: Would the registrationId not always be in the RegistrationViewEntity? If not, can we make it so (enforce it)? If not, just call it registrationId to be more clear?
```
RubenGeo, line 25: Because you can select which 'registration view' fields you want to return when using the paginate package. See for example 'select' here in the url of example on the paginate package [readme](https://github.com/ppetzold/nestjs-paginate?tab=readme-ov-file#endpoint). So there is guarantee which fields return

services/121-service/src/events/events.service.ts
```diff
@@ -60,6 +69,8 @@ export class EventsService {
     );
   }
 
+  // TODO: REFACTOR: put the parameters into an interface
+  // TODO: REFACTOR: why not call the input parameters oldRegistrationViews and newRegistrationsViews and always let them be arrays for simplicity?
```
RubenGeo, line 73: I was inspired here by the typeorm .save() method which also takes in an array or a single entity. Which I always thought was pretty convenient as the consumer of that package, but it does make the implementation of the log function a bit more complex.

services/121-service/src/events/events.controller.ts
RubenGeo: Permission check can be removed as it is already done with the decorator   @AuthenticatedUser({
    permissions: [PermissionEnum.RegistrationPersonalEXPORT],
  })

RubenGeo (approved):

PR #6467, opened 2025-02-04
Integration test for running all cron jobs

diderikvw:
[AB#33265](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33265) 

## Describe your changes

Simple integration test to see if the cron jobs can actually find the endpoints they are supposed to call. No testing of what actually happens in the endpoint called, that is part of already existing tests.

### DIFF
--- a/services/121-service/src/cronjob/cronjob.service.ts
+++ b/services/121-service/src/cronjob/cronjob.service.ts
@@ -126,12 +126,21 @@ export class CronjobService {
   @Cron(CronExpression.EVERY_DAY_AT_4AM, {
     disabled: !shouldBeEnabled(process.env.CRON_NEDBANK_VOUCHERS),
   })
-  public async cronDoNedbankReconciliation(): Promise<void> {
+  public async cronDoNedbankReconciliation(): Promise<{
+    url: string;
+    responseStatus: number;
+  }> {
     // Calling via API/HTTP instead of directly the Service so scope-functionality works, which needs a HTTP request to work which a cronjob does not have
     const accessToken = await this.axiosCallsService.getAccessToken();
     const url = `${this.axiosCallsService.getBaseUrl()}/financial-service-providers/nedbank`;
     const headers = this.axiosCallsService.accesTokenToHeaders(accessToken);
-    await this.httpService.patch(url, {}, headers);
+    const response: AxiosResponse = await this.httpService.patch(
+      url,
+      {},
+      headers,
+    );
+
+    return { url, responseStatus: response.status }; // Only used for testing purposes; this method is then called from the controller
   }
 
   @Cron(CronExpression.EVERY_DAY_AT_6AM, {
@@ -152,7 +161,11 @@ export class CronjobService {
     return { url, responseStatus: response.status }; // Only used for testing purposes; this method is then called from the controller
   }
 
-  public getAllMethods(): { methodNames: string[]; responseStatus: number } {
+  public getAllMethods(): {
+    methodNames: string[];
+    url: string;
+    responseStatus: number;
+  } {
     const prototype = Object.getPrototypeOf(this);
     const methodNames = Object.getOwnPropertyNames(prototype).filter((name) => {
       const descriptor = Object.getOwnPropertyDescriptor(prototype, name);
@@ -162,6 +175,10 @@ export class CronjobService {
         name !== 'constructor'
       );
     });
-    return { methodNames, responseStatus: 200 }; // TODO: REFACTOR: Faking the responseStatus for testing purposes, since the integration test expects it for all methods. A cleaner way would be that this function only returns methods with the @Cron decorator.
+    return {
+      methodNames,
+      url: `/get-all-methods-dummy-url`,
+      responseStatus: 200,
+    }; // TODO: REFACTOR: Faking the url and responseStatus for testing purposes, since the integration test expects it for all methods. A cleaner way would be that this function only returns methods with the @Cron decorator.
   }
 }
--- a/services/121-service/src/cronjob/cronjob.service.ts
+++ b/services/121-service/src/cronjob/cronjob.service.ts
@@ -10,18 +10,6 @@ function shouldBeEnabled(envVariable: string | undefined): boolean {
   return !!envVariable && envVariable.toLowerCase() === 'true';
 }
 
-// ##TODO: See if we can change this function to getCronFunctions and just get the public functions with the @Cron decorator
-function getAllMethods(instance: any): string[] {
-  const prototype = Object.getPrototypeOf(instance);
-  return Object.getOwnPropertyNames(prototype).filter((name) => {
-    const descriptor = Object.getOwnPropertyDescriptor(prototype, name);
-    return (
-      descriptor &&
-      typeof descriptor.value === 'function' &&
-      name !== 'constructor'
-    );
-  });
-}
 @Injectable()
 export class CronjobService {
   private httpService = new CustomHttpService(new HttpService());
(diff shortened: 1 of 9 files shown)


services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -9,6 +10,18 @@ function shouldBeEnabled(envVariable: string | undefined): boolean {
   return !!envVariable && envVariable.toLowerCase() === 'true';
 }
 
+// ##TODO: See if we can change this function to getCronFunctions and just get the public functions with the @Cron decorator
+function getAllMethods(instance: any): string[] {
```
RubenGeo, line 14: Why did you chose to use a seperate function here? Could this logic not be in  public getAllMethods()

services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -25,7 +38,16 @@ export class CronjobService {
...
     const headers = this.axiosCallsService.accesTokenToHeaders(accessToken);
-    await this.httpService.post(url, {}, headers);
+    const response: AxiosResponse = await this.httpService.post(
+      url,
+      {},
+      headers,
+    );
+    if (response.status < 200 || response.status >= 300) {
+      throw new Error(
+        `Failed to run Cron Job cronCancelByRefposIntersolve. Status code: ${response.status}`,
```
RubenGeo, line 48: You could consider making a method that does this. Something like:

`throwIfResonceCodeNotSucces(responseCode, cronjobName: string)`

services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -25,7 +38,16 @@ export class CronjobService {
...
+      );
+    }
```
RubenGeo, line 50: Some of these cronjobs take a long time to process (expecially the visa one). So in that case a 'timout' could occur I think which has response code '0' 

If this happens this error would throw in production, while the cronjob is actually running succesfull. 

Maybe instead of this throw error response could you have the cronjob controller return an object like this in which CronjobName is a new Enum:

```
[
 {
  "cronjobName": CronjobName.validateCommercialBankEthiopiaAccountEnquiries
statusCode: 200 
},
... other cronjobs object
]
```
The api test can than validate this object. The enum cronjob names can also be used in console.log etc

Other solutions are probably also possible.

services/121-service/src/payments/reconciliation/intersolve-voucher-reconciliation/intersolve-voucher-reconciliation.controller.ts
```diff
@@ -59,6 +59,7 @@ export class IntersolveVoucherReconciliationController {
   })
   @Patch('/financial-service-providers/intersolve-voucher/unused-vouchers')
   public async cronRetrieveAndUpdatedUnusedIntersolveVouchers(): Promise<void> {
+    // ##TODO: What is the reason again we cannot log the console message in the CronjobService itself?
```
RubenGeo, line 62: Because cronjob api requests can have a timeout

services/121-service/test/cronjob/cronjobs.test.ts
```diff
@@ -0,0 +1,26 @@
...
+} from '@121-service/test/helpers/utility.helper';
+
+describe('Cron jobs', () => {
+  let accessToken: string;
+
+  beforeEach(async () => {
+    accessToken = await getAccessToken();
+  });
+
+  it('run all cron jobs', async () => {
```
RubenGeo, line 15: Maybe: 
```suggestion
  it('should initiate all cronjobs succesfully', async () => {
```

RubenGeo (changes requested):

services/121-service/src/cronjob/cronjob.service.ts
diderikvw: I could not get that to work before, I will try again.

services/121-service/src/payments/reconciliation/intersolve-voucher-reconciliation/intersolve-voucher-reconciliation.controller.ts
diderikvw: I will remove this comment.

services/121-service/src/cronjob/cronjob.service.ts
diderikvw: Try to filter out the private method, or put it outside of the class.

services/121-service/src/cronjob/cronjob.service.ts
diderikvw: Solved.

services/121-service/src/cronjob/cronjob.service.ts
diderikvw: Done.

services/121-service/test/cronjob/cronjobs.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    // Act
+    const response = await getServer()
+      .patch(`/cronjob`)
+      .set('Cookie', [accessToken]);
+
+    // Assert
+    expect(response.status).toBe(HttpStatus.OK);
+    for (const cronjob of response.body) {
+      if (cronjob.responseStatus < 200 || cronjob.responseStatus >= 300) {
+        throw new Error(
```
RubenGeo, line 27: So we normally do not throw Errors in api tests. I think it makes sense to use one of the following approaches:

1:

`expect(response.body).toMatchSnapshot()`

However than we need to update the snapshot here anytime we have a new cronjob.

2: Or you can do something like:

```
    response.body.forEach((cronjob: { methodName: string; responseStatus: number }) => {
      expect(cronjob.responseStatus).toBeGreaterThanOrEqual(200);
      expect(cronjob.responseStatus).toBeLessThan(300);
    });
```
but than we do not see the name of the method that failed in a failed test run which is a bit annoying..

3: Or you could do something like (which I got from chatgpt):
```
    // Assert
    expect(response.status).toBe(HttpStatus.OK);

    response.body.forEach((cronjob: { methodName: string; responseStatus: number }) => {
      const expectedObject = {
        methodName: cronjob.methodName,
        responseStatus: expect.toBeGreaterThanOrEqual(200) && expect.toBeLessThan(300), // not sure if this part works
      };

      expect(cronjob).toMatchObject(expectedObject);
    });
```

RubenGeo (changes requested):

services/121-service/src/cronjob/dtos/run-cronjobs-response.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class RunCronjobsResponseDto {
+  @ApiProperty({ example: 'cronGetDailyExchangeRates' })
+  public readonly methodName: string;
+  @ApiProperty({ example: 'http://localhost:3000/api/exchange-rates' })
+  public readonly url: string;
```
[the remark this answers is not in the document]
diderikvw: It was handy for me to test during development, when calling the endpoint from Swagger :)
But if you think that it will not be handy in the future, I will remove it.

services/121-service/test/cronjob/cronjobs.test.ts
diderikvw: How about now?

RubenGeo (dismissed):

RubenGeo:
I approved it but the tests are failing @diderikvw

PR #6516, opened 2025-02-13
Set focus on the first form-control, but preventing "too soon" validation-errors

elwinschmitz:
[AB#33223](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33223)
[AB#33599](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33599)

https://lively-river-04adce503-6516.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
+++ b/interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
@@ -10,6 +10,7 @@
 >
   <ng-template pTemplate="headless">
     <form
+      #sideBarForm
       class="flex max-h-full min-h-full w-full flex-col overflow-y-auto p-4 pb-20"
       [formGroup]="formGroup()"
       (ngSubmit)="onFormSubmit()"
@@ -17,7 +18,6 @@
       (keydown.meta.enter)="onFormSubmit()"
       pFocusTrap
       [pFocusTrapDisabled]="!visible()"
-      [pAutoFocus]="visible()"
     >
       <div class="flex grow flex-col gap-2">
         <h2 class="mb-5">
--- a/interfaces/Portalicious/src/app/components/form/form-sidebar.component.ts
+++ b/interfaces/Portalicious/src/app/components/form/form-sidebar.component.ts
@@ -1,12 +1,13 @@
 import {
   ChangeDetectionStrategy,
   Component,
+  ElementRef,
   input,
   model,
+  viewChild,
 } from '@angular/core';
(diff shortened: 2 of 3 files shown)


interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
```diff
@@ -18,6 +19,7 @@
       pFocusTrap
       [pFocusTrapDisabled]="!visible()"
       [pAutoFocus]="visible()"
```
aberonni, line 21: This `pAutoFocus` doesn't seem to be working. It's not the biggest deal (I couldn't get it to work in confirmation popups either) but maybe we can remove it then.

interfaces/Portalicious/src/app/components/form/form-sidebar.component.ts
```diff
@@ -34,6 +36,7 @@ export class FormSidebarComponent<
   readonly visible = model<boolean>(false);
   readonly formTitle = input.required<string>();
   readonly modal = model<boolean>(true);
+  readonly sideBarForm = viewChild<ElementRef<HTMLFormElement>>('sideBarForm');
```
aberonni, line 39:
```suggestion
  readonly sideBarForm = viewChild.required<ElementRef<HTMLFormElement>>('sideBarForm');
```

aberonni (approved):
Just a couple of nitpicks

interfaces/Portalicious/src/app/components/form/form-sidebar.component.ts
elwinschmitz: I was doubting between this and "only the question-mark" below. ;)
And that won out because its only 1 character. :P

interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
elwinschmitz: Yeah, I think the whole feature of Prime-Autofocus is not compatible with the requirement "the first form element should not show an error"...

But I'll check if it isn't a crucial part of my fix...

elwinschmitz: In the end it turns out my fix is a "manual recreation" of that `pAutofocus`-directive...

So I wonder witch edge-case we're not covering here now... But it seems to work... ;)

aberonni (approved):

PR #6535, opened 2025-02-19
Include Matomo-client for tracking default metrics

elwinschmitz:
[AB#33700](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33700)

## Describe your changes

- Load the Matomo-client only when necessary
- Include all the necessary deployment-configuration
- Limit all features that would require cookies/privacy-notices, etc.

## Preview deployment

https://lively-river-04adce503-6535.westeurope.5.azurestaticapps.net


aberonni (approved):
There are some small things that I'd like to refactor - but I'll do that in a separate PR because the implementation here looks very solid and thorough (I especially like the attention to detail in the .mjs files). So I'm merging as-is so hopefully we can tracking data and in the meantime do the refactor separately. (Why refactor? This question will be answered in the new PR).

PR #6536, opened 2025-02-20
show voucher button for paper voucher

tijsziere:
[AB#33861](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33861) 

https://lively-river-04adce503-6536.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/pages/project-registration-activity-log/components/table-cell-overview/table-cell-overview.component.ts
+++ b/interfaces/Portalicious/src/app/pages/project-registration-activity-log/components/table-cell-overview/table-cell-overview.component.ts
@@ -26,8 +26,8 @@ import {
 } from '~/components/colored-chip/colored-chip.helper';
 import { TableCellComponent } from '~/components/query-table/components/table-cell/table-cell.component';
 import { MESSAGE_CONTENT_TYPE_LABELS } from '~/domains/message/message.helper';
-import { REGISTRATION_STATUS_LABELS } from '~/domains/registration/registration.helper';
 import { FSPS_WITH_VOUCHER_SUPPORT } from '~/domains/payment/payment.helpers';
+import { REGISTRATION_STATUS_LABELS } from '~/domains/registration/registration.helper';
 import { Activity } from '~/domains/registration/registration.model';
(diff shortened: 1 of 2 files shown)


interfaces/Portalicious/src/app/pages/project-registration-activity-log/components/table-cell-overview/table-cell-overview.component.ts
```diff
@@ -146,8 +146,10 @@ export class TableCellOverviewComponent
 
     if (
       item.type !== ActivityTypeEnum.Transaction ||
-      item.attributes.financialServiceProviderName !==
-        FinancialServiceProviders.intersolveVoucherWhatsapp
+      (item.attributes.financialServiceProviderName !==
+        FinancialServiceProviders.intersolveVoucherWhatsapp &&
+        item.attributes.financialServiceProviderName !==
+          FinancialServiceProviders.intersolveVoucherPaper)
```
aberonni, line 152: Should use `FSPS_WITH_VOUCHER_SUPPORT` 

```suggestion
      !FSPS_WITH_VOUCHER_SUPPORT.includes(item.attributes.financialServiceProviderName)
```

Which makes me wonder also if the `isIntersolveVoucher` computed property in the same file has the same issue as the one you're fixing. We can look at this together if it helps.

aberonni (changes requested):

interfaces/Portalicious/src/app/pages/project-registration-activity-log/components/table-cell-overview/table-cell-overview.component.ts
tijsziere: I think you're right, we also need to change that computation. Do I understand correctly that you are suggesting to have an enum with voucher_fsps somewhere? Maybe best if we briefly talk indeed!

aberonni: That enum already exists! You just need to import and update this and the other computed property. And then test to make sure both work with paper vouchers ofc.

Slip in my DMs if you wanna look at it together. Or if you'd like me to take over that is also fine!

PR #6538, opened 2025-02-21
Refactor move logic to registration action menu service

RubenGeo:
[AB#33898](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33898) 

Moved logic from to project-registrations.page to seperate service. Refactor to prepare for the introcution of an actions menu on the profile page

https://lively-river-04adce503-6538.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
+++ b/interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
@@ -61,7 +61,7 @@ export class ProjectRegistrationsPageComponent {
   private router = inject(Router);
   private projectApiService = inject(ProjectApiService);
   private toastService = inject(ToastService);
-  private registrationMenuService = inject(RegistrationActionMenuService);
+  readonly registrationMenuService = inject(RegistrationActionMenuService);
 
   readonly registrationsTable =
     viewChild.required<RegistrationsTableComponent>('registrationsTable');
@@ -76,23 +76,18 @@ export class ProjectRegistrationsPageComponent {
 
   project = injectQuery(this.projectApiService.getProject(this.projectId));
 
-  readonly canChangeStatus = computed(() => {
-    const project = this.project.data();
-    if (!project) {
-      return () => false;
-    }
-
-    return (status: RegistrationStatusChangeTarget) =>
-      this.registrationMenuService.canChangeStatus()({
+  readonly canChangeStatus = computed(
+    () => (status: RegistrationStatusChangeTarget) =>
+      this.registrationMenuService.canChangeStatus({
         status,
         projectId: this.projectId(),
-        hasValidation: !!project.validation,
-      });
-  });
+        hasValidation: !!this.project.data()?.validation,
+      }),
+  );
 
   readonly canSendMessage = computed(
     () => () =>
-      this.registrationMenuService.canSendMessage()({
+      this.registrationMenuService.canSendMessage({
         projectId: this.projectId(),
       }),
   );
@@ -157,6 +152,7 @@ export class ProjectRegistrationsPageComponent {
       RegistrationStatusEnum.deleted,
     ),
   ]);
+
   sendMessage({
     triggeredFromContextMenu = false,
   }: {
@@ -201,7 +197,7 @@ export class ProjectRegistrationsPageComponent {
     return this.registrationMenuService.createContextItemForRegistrationStatus({
       status,
       projectId: this.projectId(),
-      hasValidation: this.project.data()?.validation ?? false,
+      hasValidation: !!this.project.data()?.validation,
       command: () => {
         this.changeStatus({
           status,
--- a/interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
+++ b/interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
@@ -25,11 +25,13 @@ import {
   REGISTRATION_STATUS_VERB,
   registrationLink,
 } from '~/domains/registration/registration.helper';
+import { RegistrationStatusChangeTarget } from '~/domains/registration/registration.model';
 import { ChangeStatusDialogComponent } from '~/pages/project-registrations/components/change-status-dialog/change-status-dialog.component';
 import { ExportRegistrationsComponent } from '~/pages/project-registrations/components/export-registrations/export-registrations.component';
 import { ImportRegistrationsComponent } from '~/pages/project-registrations/components/import-registrations/import-registrations.component';
 import { SendMessageDialogComponent } from '~/pages/project-registrations/components/send-message-dialog/send-message-dialog.component';
 import { AuthService } from '~/services/auth.service';
+import { RegistrationActionMenuService } from '~/services/registration-action-menu.service';
 import { ToastService } from '~/services/toast.service';
 import { getOriginUrl } from '~/utils/url-helper';
 
@@ -59,6 +61,7 @@ export class ProjectRegistrationsPageComponent {
   private router = inject(Router);
   private projectApiService = inject(ProjectApiService);
   private toastService = inject(ToastService);
+  private registrationMenuService = inject(RegistrationActionMenuService);
 
   readonly registrationsTable =
     viewChild.required<RegistrationsTableComponent>('registrationsTable');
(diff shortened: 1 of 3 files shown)


interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
```diff
@@ -73,45 +75,29 @@ export class ProjectRegistrationsPageComponent {
...
-        }
+  readonly canChangeStatus = computed(() => {
+    const project = this.project.data();
+    if (!project) {
+      return () => false;
+    }
+
+    return (
+      status:
+        | RegistrationStatusEnum.declined
```
RubenGeo, line 86: Would we benefit from creating a interface/enum which is this subset of registration statusses. As we use it  in multiple places? 
@aberonni

aberonni (changes requested):
image

In it's current state, tbh, I don't think this is an improvement over having the same code duplicated in two places. The added complexity and confusing syntax doesn't seem like a worthy tradeoff.

However, it might be because of the way it's implemented. I'll leave it up to you how to address it (of course, feel free to ask for help if I can).

interfaces/Portalicious/src/app/services/registration-action-menu.service.ts
```diff
@@ -0,0 +1,95 @@
...
+      ({
+        status,
+        projectId,
+        hasValidation,
+      }: {
+        status: RegistrationStatusChangeTarget;
+        projectId: string;
+        hasValidation: boolean;
+      }) => {
+        if (status === RegistrationStatusEnum.validated && !hasValidation) {
```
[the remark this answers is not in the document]
RubenGeo: I agree, this is also the reason I  left the `createContextMenuItemForRegistrationStatus` in the project-registrations.page.ts. To take care of duplicate logic in what that page fetches 

- So in the backend our services do not have any context of which projectId/programId is used and they always need to get it from a param. There is no state of `public programId` at the top of a service file
- So I sort of assumed that was not possible/good practice and that only pages.ts have context of projectId in url. Is there an example how we do this in another service in Portalicious?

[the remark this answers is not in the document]
aberonni: In the spirit of "teach a man to fish" 😄 :

image

Just in case it isn't clear: because of the way [tanstack-query](https://tanstack.com/query/v5/docs/framework/angular/overview) works, just because you "refetch" an endpoint in a service, does not mean that you refetch that endpoint from the backend - that only happens if there is no cached version of the response.

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
```diff
@@ -73,45 +76,25 @@ export class ProjectRegistrationsPageComponent {
...
-          | RegistrationStatusEnum.paused
-          | RegistrationStatusEnum.validated,
-      ) => {
-        if (
-          status === RegistrationStatusEnum.validated &&
-          !this.project.data()?.validation
-        ) {
-          return false;
-        }
+  readonly canChangeStatus = computed(() => {
```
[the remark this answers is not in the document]
RubenGeo: I think you could call `registrationMenuService.changeStatus` directly in the template if we change `private registrationMenuService` to `readonly registrationMenuService` but than that would only make sense I guess if we are able to get the project in the registrationMenuService. Relates to this comment: https://github.com/global-121/121-platform/pull/6538#discussion_r1969521875

interfaces/Portalicious/src/app/services/registration-action-menu.service.ts
```diff
@@ -0,0 +1,95 @@
...
+import { RegistrationStatusChangeTarget } from '~/domains/registration/registration.model';
+import { AuthService } from '~/services/auth.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class RegistrationActionMenuService {
+  private readonly authService = inject(AuthService);
+
+  readonly canChangeStatus = computed(
```
aberonni, line 19: Come to think of it, the fact that this is computed makes no sense because it doesn't recompute 😄 Maybe we need to have a conversation about what 'computed' means?

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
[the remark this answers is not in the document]
aberonni: True. Or if that isn't possible, at least a simplification like this (but this would mean `.registrationMenuService.canChangeStatus` not being computed anymore, which it doesn't need to be the way it is currently written).

```suggestion
  readonly canChangeStatus = computed(() => this.registrationMenuService.canChangeStatus({
        status,
        projectId: this.projectId(),
        hasValidation: !!project.data()?.validation,
      }));
```

RubenGeo:
> image
> 
> In it's current state, tbh, I don't think this is an improvement over having the same code duplicated in two places. The added complexity and confusing syntax doesn't seem like a worthy tradeoff.
> 
> However, it might be because of the way it's implemented. I'll leave it up to you how to address it (of course, feel free to ask for help if I can).

I see your point. 👍 I do have to say that it felt easy to implement the Action Menu on the profile page: https://github.com/global-121/121-platform/pull/6541 after this refactor. An advantage of having a seperate re-used service even if it is more verbose is that business logic is stored in one place. So you have less change of forgetting to update the other place if one thing changes

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
```diff
@@ -73,45 +76,25 @@ export class ProjectRegistrationsPageComponent {
...
-    }),
+    return (status: RegistrationStatusChangeTarget) =>
+      this.registrationMenuService.canChangeStatus()({
+        status,
+        projectId: this.projectId(),
+        hasValidation: !!project.validation,
+      });
+  });
+
+  readonly canSendMessage = computed(
```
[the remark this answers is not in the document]
RubenGeo: Since RegistrationActionMenuService.canSendMessage is not computed anymore. It makes sense now that this is still here right?

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
```diff
@@ -74,44 +77,19 @@ export class ProjectRegistrationsPageComponent {
...
+      }),
   );
-  readonly canSendMessage = computed(() =>
-    this.authService.hasPermission({
-      projectId: this.projectId(),
-      requiredPermission: PermissionEnum.RegistrationNotificationCREATE,
-    }),
+
+  readonly canSendMessage = computed(
+    () => () =>
```
aberonni, line 89:
```suggestion
    () =>
```

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
```diff
@@ -150,19 +128,14 @@ export class ProjectRegistrationsPageComponent {
...
+      projectId: this.projectId(),
       command: () => {
         this.sendMessage({
           triggeredFromContextMenu: true,
         });
       },
-    },
-    {
-      separator: true,
-    },
```
aberonni, line 165: Is there a reason the separator was removed?

interfaces/Portalicious/src/app/services/registration-action-menu.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  public canSendMessage({ projectId }: { projectId: string }): boolean {
+    return this.authService.hasPermission({
+      projectId,
+      requiredPermission: PermissionEnum.RegistrationNotificationCREATE,
+    });
+  }
+
+  public createContextItemForRegistrationStatus({
```
aberonni, line 57:
```suggestion
  public createContextItemForRegistrationStatusChange({
```

aberonni (changes requested):
Minor stuff but otherwise looking good!

interfaces/Portalicious/src/app/pages/project-registrations/project-registrations.page.ts
RubenGeo: Nope :)

aberonni (approved):

PR #6539, opened 2025-02-21
chore: add tracking service to portalicious

aberonni:
[AB#33897](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33897)

## Describe your changes

Following up on the work done by @elwinschmitz in #6535 - I decided to refactor some of the conditions into a separate service, similarly to what was done with the authservice, so that the application as a whole is unaware (and does not care) whether Matomo is enabled / configured for the current instance.

https://lively-river-04adce503-6539.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/pages/privacy/privacy.page.html
+++ b/interfaces/Portalicious/src/app/pages/privacy/privacy.page.html
@@ -71,31 +71,7 @@
     The data of the 121 Platform is hosted in a Microsoft Azure-environment of
     The Netherlands Red Cross.
   </p>
-  @if (isMatomoEnabled) {
-    <p i18n>
-      We use the external "<em>Matomo Cloud</em>"-service by InnoCraft Ltd. to
-      measure the usage of the 121 Platform.
-    </p>
-    <p i18n>
-      We do not share any personal information with Matomo/InnoCraft Ltd. Our
-      use of their service is covered by the
-      <a
-        href="https://matomo.org/matomo-cloud-privacy-policy/"
-        target="_blank"
-        >Matomo Cloud Privacy Policy</a
-      >
-      and the
-      <a
-        href="https://matomo.org/matomo-cloud-dpa/"
-        target="_blank"
-        >Matomo Cloud Data Processing Agreement</a
-      >.
-    </p>
-  } @else {
-    <p i18n>
-      We do not share any personal information with other third-parties.
-    </p>
-  }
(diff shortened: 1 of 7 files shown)


interfaces/Portalicious/src/app/pages/privacy/privacy.page.html
```diff
@@ -71,31 +71,7 @@ <h2 i18n>Who we share information with</h2>
...
-        target="_blank"
-        >Matomo Cloud Data Processing Agreement</a
-      >.
-    </p>
-  } @else {
-    <p i18n>
-      We do not share any personal information with other third-parties.
-    </p>
-  }
+  <ng-container *ngComponentOutlet="TrackingCopyComponent"></ng-container>
```
elwinschmitz, line 74: The downside of this "splitting up in parts" is that the full text-body of the page is harder to read/write...

But then again, the whole approach of a mix of "conditional text blocks" for a Privacy-notice is "not ideal", I'd say...

Also.. there's the issue of "naming things"...
In this case its not really an issue of "to track" or "not to track" (because there is no cookie set to "keep track of" visitors/users... there is just "measure a thing" or not...

And the 'legal' implication is that whether we are involving a separate third-party or not...

elwinschmitz: Speaking about that 'legal' part...

The "last changed"-date should be updated with this change as well...
(on Line 4: https://github.com/global-121/121-platform/pull/6539/files#diff-648071ea4639a1f3c384299c6418583836f4d7a70ab7df4d0db0135867d00040L4)

But that isn't really tenable now... with the whole 'conditional content'-magic...

At least there is now only 1 "last updated"-date... and multiple possible "places that can have changes"... :/

elwinschmitz (commented):
Maybe we can discuss this topic further after tomorrow's Matomo-focused meeting...

interfaces/Portalicious/src/app/pages/privacy/privacy.page.html
aberonni: I hear you. I still prefer the tracking service to decide what to show here. Do you have any desired actions / outcomes here?

elwinschmitz: I don't think the actual date really, _really_ matters in this case..

I've added a (more visible) note in the upcoming backlog-item(s) that will touch on this in the future...

PR #6522, opened 2025-02-14
feat: duplicates in portalicious

aberonni:
[AB#33536](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33536) [AB#33535](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33535)

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have added tests wherever relevant
- [x] I have made sure that all automated checks pass before requesting a review
- [x] I do not need any deviation from our PR guidelines

## Portalicious preview deployment

https://lively-river-04adce503-6522.westeurope.5.azurestaticapps.net

PR #6541, opened 2025-02-25
Feat.action menu profile page

RubenGeo:
[AB#33693](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33693) 

## Describe your changes

Add action menu to profile page

https://lively-river-04adce503-6541.westeurope.5.azurestaticapps.net

### DIFF
--- a/e2e/portalicious/pages/BasePage.ts
+++ b/e2e/portalicious/pages/BasePage.ts
@@ -110,8 +110,8 @@ class BasePage {
   }
 
   async goto(path: string) {
-    const defaulLanguage = 'en-GB';
-    path = `${defaulLanguage}${path}`;
+    const defaultLanguage = 'en-GB';
+    path = `${defaultLanguage}${path}`;
     await this.page.goto(path);
   }
 }
--- a/e2e/portalicious/pages/BasePage.ts
+++ b/e2e/portalicious/pages/BasePage.ts
@@ -108,6 +108,12 @@ class BasePage {
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(filePath);
   }
+
+  async goto(path: string) {
+    const defaulLanguage = 'en-GB';
+    path = `${defaulLanguage}${path}`;
+    await this.page.goto(path);
+  }
 }
 
 export default BasePage;
--- a/e2e/portalicious/pages/RegistrationActivityLogPage.ts
+++ b/e2e/portalicious/pages/RegistrationActivityLogPage.ts
@@ -26,15 +26,6 @@ class RegistrationActivityLogPage extends RegistrationBasePage {
     expect(sentMessageText).toBe(message);
     await dropdownButton.click();
   }
-
-  async clickActionDropdown() {
-    await this.page.getByRole('button', { name: 'Actions' }).click();
-  }
-
-  async inititateAction(action: string) {
-    await this.clickActionDropdown();
-    await this.page.getByRole('menuitem', { name: action }).click();
-  }
 }
 
 export default RegistrationActivityLogPage;
--- a/e2e/portalicious/pages/RegistrationActivityLogPage.ts
+++ b/e2e/portalicious/pages/RegistrationActivityLogPage.ts
@@ -26,6 +26,15 @@ class RegistrationActivityLogPage extends RegistrationBasePage {
     expect(sentMessageText).toBe(message);
     await dropdownButton.click();
   }
+
+  async clickActionDropdown() {
+    await this.page.getByRole('button', { name: 'Actions' }).click();
+  }
+
+  async inititateAction(action: string) {
+    await this.clickActionDropdown();
+    await this.page.getByRole('menuitem', { name: action }).click();
+  }
 }
 
 export default RegistrationActivityLogPage;
--- a/e2e/portalicious/tests/RegistrationPage/InitiateActions.spec.ts
+++ b/e2e/portalicious/tests/RegistrationPage/InitiateActions.spec.ts
@@ -1,0 +1,124 @@
+import { expect, test } from '@playwright/test';
+
+import { SeedScript } from '@121-service/src/scripts/enum/seed-script.enum';
+import {
+  getRegistrationIdByReferenceId,
+  seedIncludedRegistrations,
+} from '@121-service/test/helpers/registration.helper';
+import {
+  getAccessToken,
+  resetDB,
+} from '@121-service/test/helpers/utility.helper';
+import {
+  programIdPV,
+  registrationPV5,
+} from '@121-service/test/registrations/pagination/pagination-data';
+
+import LoginPage from '@121-e2e/portalicious/pages/LoginPage';
+import RegistrationActivityLogPage from '@121-e2e/portalicious/pages/RegistrationActivityLogPage';
+
+const projectId = 2;
+let registrationId: number;
+
+test.beforeEach(async ({ page }) => {
+  await resetDB(SeedScript.nlrcMultiple);
+
+  const accessToken = await getAccessToken();
+  await seedIncludedRegistrations([registrationPV5], programIdPV, accessToken);
+  registrationId = await getRegistrationIdByReferenceId({
+    programId: programIdPV,
+    referenceId: registrationPV5.referenceId,
+    accessToken,
+  });
+
+  // Login
+  const loginPage = new LoginPage(page);
+  await page.goto(`/`);
+  await loginPage.login(
+    process.env.USERCONFIG_121_SERVICE_EMAIL_ADMIN,
+    process.env.USERCONFIG_121_SERVICE_PASSWORD_ADMIN,
+  );
+});
+
+test('User can initiate registration status changes from registration page', async ({
+  page,
+}) => {
+  const activityLogPage = new RegistrationActivityLogPage(page);
+
+  const statusChangeActions = ['Validate', 'Include', 'Decline', 'Delete'];
+
+  for (const actionName of statusChangeActions) {
+    await test.step('Navigate to registration', async () => {
+      await activityLogPage.goto(
+        `/project/${projectId}/registrations/${registrationId}`, // TODO should we not hardcode the url here?
+      );
+    });
+
+    await test.step(`Initiate action: ${actionName}`, async () => {
+      await activityLogPage.inititateAction(actionName);
+
+      const dialog = page.getByRole('dialog');
+      await expect(dialog).toBeVisible();
+
+      const title = dialog.getByText(`${actionName} registration(s)`);
+      await expect(title).toBeVisible();
+
+      // Check if the action is only for 1 registration
+      const paragraph = dialog.getByText(/You're about to/);
+      await expect(paragraph).toContainText(
+        `You're about to ${actionName.toLowerCase()} 1 registrations.`,
+      );
+    });
+  }
+});
+
+test('User can open add note sidebar from action menu', async ({ page }) => {
+  const activityLogPage = new RegistrationActivityLogPage(page);
+
+  const actionName = 'Add note';
+
+  await test.step('Navigate to registration', async () => {
+    await activityLogPage.goto(
+      `/project/${projectId}/registrations/${registrationId}`,
+    );
+  });
+
+  await test.step(`Initiate action: ${actionName}`, async () => {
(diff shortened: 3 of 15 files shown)


interfaces/Portalicious/src/app/services/paginate-query.service.ts
```diff
@@ -248,4 +249,20 @@ export class PaginateQueryService {
       previewItem: selection[0],
     };
   }
+
+  public getActionDataForRegistration(
```
RubenGeo, line 253: @aberonni
I’m not entirely sure about this implementation.
	•	You need this object to perform a bulk action, as it uses our paginate queries, and both ChangeStatusDialogComponent and SendMessageDialogComponent expect it.
	•	I think we definitely need a method to generate the missing parts of .query.filter, but I’m not sure if we need all the other properties to make the correct API call.
	•	One option is to add a method in src/app/domains/registration/registration.api.service.ts, e.g., changeStatusOnRegistration, that generates the appropriate query filter.
	•	However, this would require refactoring ChangeStatusDialogComponent and SendMessageDialogComponent to call that method instead.

Additionally, is this the right place for this function?
	•	Should PaginateQueryService be aware of the registration context, or would it be better placed elsewhere?
	•	Alternatively, should we generalize this using a T type, even though it’s currently only used for registrations? That feels like over-engineering, but I’d like your thoughts on it.

RubenGeo:
@aberonni Could you take a look at this before I/we implement the e2e tests

interfaces/Portalicious/src/app/components/registration-page-layout/registration-page-layout.component.ts
```diff
@@ -137,6 +200,32 @@ export class RegistrationPageLayoutComponent {
...
+    const registration = this.registration.data();
+    if (!registration) {
+      return;
+    }
+    const actionData =
+      this.paginateQueryService.getActionDataForRegistration(registration);
+
+    this.changeStatusDialog().triggerAction(actionData, status);
+  }
+
```
aberonni, line 228: This is how I would refactor your code, based on your comments. I don't think it needs to be in the paginate query service (for now) and I don't think we need to refactor the triggerAction flow. While it's true that we don't need the extra object info, it's also true that an "ActionData" object is a known entity that we can work with.

```suggestion
  private getActionDataForSingleRegistration() {
    const registration = this.registration.data();
    if (!registration) {
      return;
    }
    
    return {
      query: {
        filter: {
          referenceId: `${FilterOperator.EQ}:${registration.referenceId}`,
        },
      },
      count: 1,
      selection: [],
      selectAll: false,
      previewItem: registration,
    };
  }
  
  sendMessage() {
    const actionData =
      this.getActionDataForSingleRegistration();
    if (!actionData) {
      return;
    }
    this.sendMessageDialog().triggerAction(actionData);
  }

  changeStatus({
    status,
  }: {
    status: RegistrationStatusEnum;
    triggeredFromContextMenu?: boolean;
  }) {
    const actionData =
      this.getActionDataForSingleRegistration();
    if (!actionData) {
      return;
    }
    this.changeStatusDialog().triggerAction(actionData, status);
  }

```

FWIW I think you could've seen a similar solution, had you concentrated on the existing duplication in your functions. Both of your functions had the same 5 lines of code:

```ts
    const registration = this.registration.data();
    if (!registration) {
      return;
    }
    const actionData =
      this.paginateQueryService.getActionDataForRegistration(registration);
```

FYI I only looked at this code and not at the whole PR.

e2e/portalicious/pages/BasePage.ts
```diff
@@ -104,6 +104,12 @@ class BasePage {
     const fileChooser = await fileChooserPromise;
     await fileChooser.setFiles(filePath);
   }
+
+  async goto(path: string) {
+    const defaulLanguage = 'en-GB';
```
RubenGeo, line 109: Should we get this from somewhere? We cannot access NG_DEFAULT_LOCALE in playwright at this moment

e2e/portalicious/tests/RegistrationProfilePage/InitiateActions.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  page,
+}) => {
+  const activityLogPage = new RegistrationActivityLogPage(page);
+  await test.step('Validate status change dialog opened', async () => {
+    const statusChangeActions = ['Validate', 'Include', 'Decline', 'Delete'];
+
+    for (const actionName of statusChangeActions) {
+      await test.step('Navigate to registration', async () => {
+        await activityLogPage.goto(
+          `/project/${projectId}/registrations/${registrationId}`, // TODO should we not hardcode the url here?
```
RubenGeo, line 53: Should we not hardcode the url here but get it dynamically?

interfaces/Portalicious/src/app/services/paginate-query.service.ts
RubenGeo: @aberonni 
After our discussion I decided to keep this method in this service:
- Good to seperate the concern of creating pagination queries out of the componenents
- This service seems like the most logical place.

Furthermore I decided not to implement the T type for this. The query that is created is to specific as it uses referenceId

interfaces/Portalicious/src/app/components/button-menu/button-menu.component.ts
```diff
@@ -27,4 +28,17 @@ export class ButtonMenuComponent {
   readonly size = input<'large' | 'small'>();
 
   readonly menuOpen = model(false);
+
+  // Filter out items without command or visible children as these than look like a subheader without any action or children or routerlink
```
RubenGeo, line 32: Filtering out these 'sub header' items without children seems like something generic you would always want. But I am also a bit afraid that the logic in this component is a bit too hidden. 

What do you think?

e2e/portalicious/pages/BasePage.ts
aberonni:
```suggestion
    const defaultLanguage = 'en-GB';
```

Besides the typo, I think this is fine. We have it hardcoded in plenty of places in the E2E tests already - and if it breaks we will hear about it in CI.

e2e/portalicious/pages/RegistrationActivityLogPage.ts
```diff
@@ -27,6 +27,15 @@ class RegistrationActivityLogPage extends BasePage {
     expect(sentMessageText).toBe(message);
     await dropdownButton.click();
   }
+
+  async clickActionDropdown() {
```
aberonni, line 31: I've made quite some changes in #6522 that add a "RegistrationBasePage" because these actions will be relevant for all registration pages. Maybe worth seeing how bring these new functions into there.

interfaces/Portalicious/src/locale/messages.xlf
```diff
@@ -1540,6 +1540,15 @@
       <trans-unit id="9022161403115733621" datatype="html">
         <source>We do not share any personal information with other third-parties.</source>
       </trans-unit>
+      <trans-unit id="8433732438274024544" datatype="html">
+        <source>Add note</source>
```
aberonni, line 1544: I'm surprised this was added, and nothing was removed. We already had this label somewhere. Maybe it's missing an id?

interfaces/Portalicious/src/app/services/paginate-query.service.ts
```diff
@@ -248,4 +249,20 @@ export class PaginateQueryService {
...
+        filter: {
+          referenceId: `${FilterOperator.EQ}:${registration.referenceId}`,
+        },
+      },
+      count: 1,
+      selection: [],
+      selectAll: false,
+      previewItem: registration,
+    };
+  }
```
aberonni, line 267: I do think this could be made to be generic, and to reuse the function we already have.

```suggestion
  public singleItemToActionData<TData>({
    item,
    fieldForFilter,
  }: {
    item: TData;
    fieldForFilter: keyof TData & string;
  }) {
    return this.selectionEventToActionData({
      selection: [item],
      fieldForFilter,
      totalCount: 1,
      previewItemForSelectAll: item,
    });
  }
```

interfaces/Portalicious/src/app/components/registration-page-layout/registration-page-layout.component.ts
```diff
@@ -64,6 +85,47 @@ export class RegistrationPageLayoutComponent {
...
+          visible: this.canUpdatePersonalData(),
+        },
+        this.registrationMenuService.createContextItemForMessage({
+          projectId: this.projectId(),
+          command: () => {
+            this.sendMessage();
+          },
+        }),
+      ],
+    },
```
aberonni, line 106:
```suggestion
        {
          label: $localize`Add note`,
          icon: 'pi pi-pen-to-square',
          command: () => {
            this.addNoteFormVisible.set(true);
          },
          visible: this.canUpdatePersonalData(),
        },
        this.registrationMenuService.createContextItemForMessage({
          projectId: this.projectId(),
          command: () => {
            this.sendMessage();
          },
        }),
```

I see extra whitespace in the menu, and I think it's because of this extra level of nesting that shouldn't be necessary.

e2e/portalicious/tests/RegistrationPage/InitiateActions.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    await activityLogPage.goto(
+      `/project/${projectId}/registrations/${registrationId}`,
+    );
+  });
+
+  await test.step(`Initiate action: ${actionName}`, async () => {
+    await activityLogPage.inititateAction(actionName);
+    await page.waitForLoadState('domcontentloaded');
+    await page.waitForLoadState('networkidle');
+    const sidebarDescription = page.getByTestId('add-note-sidebar-subtitle');
```
aberonni, line 90: Below you did something similar with

```ts
const paragraph = dialog.getByText(/You're about to/);
```

which I think is better because it means we don't have to add a testid, and define a new element in the page.

aberonni (changes requested):
Good stuff!

interfaces/Portalicious/src/app/components/button-menu/button-menu.component.ts
aberonni: I think it's a good idea, and it makes sense that the component takes care of this logic ✌️

e2e/portalicious/pages/RegistrationActivityLogPage.ts
RubenGeo: Feels like we need to merge that one first than I can rebase and see :)

interfaces/Portalicious/src/app/services/paginate-query.service.ts
aberonni: >  I don't see such a pattern very often in Portalicious. Is there are reason for that?

Not a conscious one. I suppose we just have done more "guarding" rather than "throwing".

That said, I think you can remove the `| undefined` in the return type of `selectionEventToActionData`, and it should solve this problem. Tbh I don't know why it's there to begin with.

interfaces/Portalicious/src/app/components/registration-page-layout/registration-page-layout.component.ts
RubenGeo: If you remove that nesting it sort of converts Add Note and Send Message to the subheader style
image
I am not too annyed about this extra white space, but maybe it can be removed with some custom css?

aberonni: The extra white space is really ugly 😄  And I wouldn't introduce extra CSS unless necessary - that feels like a hacky workaround.

I see that the issue you're encountering is a somewhat known bug https://github.com/primefaces/primeng/issues/17131

Maybe we can indeed used [TieredMenu](https://primeng.org/tieredmenu#popup) instead? And put the status update group under a tier. If that doesn't work, then I would rather
- remove the "status update" category
- just add a separator between the add note & message actions, and the other actions (like we have in the registrations table)
- add a comment about what we would've like to do, but can't, and link to the bug above

RubenGeo: We chose to add a label to the first 2 nested items called 'General'

PR #6544, opened 2025-02-26
Add e2e tests for status changes

Piotrk39:
[AB#33859](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33859) 

## Describe your changes
Adds multiple tests for testing status changes of registrations.


e2e/helpers/interfaces.ts
aberonni, line 1: I don't think the solution is overengineered, but I do question why this file is necessary. Maybe I'm missing something though.

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromIncludedToCompleted.ts
```diff
@@ -0,0 +1,81 @@
...
+} from '@121-service/test/helpers/registration.helper';
+import { getAccessToken } from '@121-service/test/helpers/utility.helper';
+import {
+  programIdPV,
+  registrationPvMaxPayment,
+} from '@121-service/test/registrations/pagination/pagination-data';
+
+import { Components, Pages } from '../../../helpers/interfaces';
+
+export default (pages: Partial<Pages>, components: Partial<Components>) => {
```
aberonni, line 17: Why is this needed? Can't you just pass in `page` and everything else becomes business as usual?

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromIncludedToCompleted.ts
```diff
@@ -0,0 +1,81 @@
...
+  test('[31211] Move PA(s) from status "Included" to "Completed"', async () => {
+    const accessToken = await getAccessToken();
+    const paymentReferenceId = [registrationPvMaxPayment.referenceId];
+    const { basePage, registrations } = pages;
+    const { tableComponent } = components;
+
+    if (!basePage || !registrations || !tableComponent) {
+      throw new Error('pages and components not found');
+    }
```
aberonni, line 26:
```suggestion
    const homePage = new HomePage(page);
    const registrations = new RegistrationsPage(page);
    const tableComponent = new TableComponent(page);
```

Following up on my comment above, I don't understand the need for a different paradigm here.

aberonni (changes requested):

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromIncludedToCompleted.ts
Piotrk39: In general the idea was to keep the same browser session for all the tests it would be even more beneficial if those tests shared same db setup and were sequential. Unfortunately the more tests I wrote the less it made sense. Especially for validated status that was just becoming included and other statuses becoming harder and harder to follow. Now the setup is doing a reset each time new test is executed so keeping the same browser context does not make sense anymore. If it was different situation and test were sequential and could benefit from same browser context then Playwright needs to pass the same page object for each tests meaning they need to be shared with a helper.

Saying that the benefits are close to none so I'll stick to the current implementation.

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromIncludedToCompleted.ts
Piotrk39: I reviewed my approach and I will stick to usual pattern

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+test('[31214] Move PA(s) from status "Completed" to "Included"', async ({
+  page,
+}) => {
+  const accessToken = await getAccessToken();
+  const registrations = new RegistrationsPage(page);
+  const tableComponent = new TableComponent(page);
+
+  await test.step('Change status of all registrations to "Included"', async () => {
```
nielsbom, line 52: We're doing the test with a **single** registration, right? Maybe good to communicate that here, instead of the more vague "all".

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    });
+  });
+
+  await test.step('Validate the status of the registration', async () => {
+    await registrations.validateStatusOfFirstRegistration({
+      status: 'Included',
+    });
+  });
+
+  await test.step('Change status of registratios to "Completed" with doing a payment', async () => {
```
nielsbom, line 73: Typo

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  await test.step('Change status of all registrations to "Included"', async () => {
+    await changeBulkRegistrationStatus({
+      programId: 2,
+      status: RegistrationStatusEnum.included,
+      accessToken,
+    });
+  });
+
+  await test.step('Search for the registration with status "Included"', async () => {
```
nielsbom, line 60: Do we need to filter if we already know there's only a single registration?

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      registrationPvMaxPayment.referenceId,
+      {
+        maxPayments: '2',
+      },
+      'automated test',
+      accessToken,
+    );
+  });
+
+  await test.step('Search for the registration with status "Completed"', async () => {
```
nielsbom, line 95: No filtering needed here right?

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    await tableComponent.clearAllFilters();
+    await tableComponent.filterColumnByDropDownSelection({
+      columnName: 'Registration Status',
+      selection: 'Completed',
+    });
+  });
+
+  await test.step('Change status of registratios to "Included"', async () => {
```
nielsbom, line 103: Typo

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  registrationPvMaxPayment,
+} from '@121-service/test/registrations/pagination/pagination-data';
+
+import TableComponent from '@121-e2e/portalicious/components/TableComponent';
+import BasePage from '@121-e2e/portalicious/pages/BasePage';
+import LoginPage from '@121-e2e/portalicious/pages/LoginPage';
+import RegistrationsPage from '@121-e2e/portalicious/pages/RegistrationsPage';
+
+const projectTitle = 'NLRC Direct Digital Aid Program (PV)';
+const toastMessage =
```
nielsbom, line 26: Why a variable? Used only once.

nielsbom (commented):
I see a lot of use of `resetDuplicateRegistrations`. Why? Do we need more than one registration? I don't see any asserts on the other registrations _not_ being changed? Can we remove that?

nielsbom (commented):
I general I feel like there's a lot of unnecessary stuff going on:
- registrations are created that are not needed
- filtering for registrations when we only have one or the top one is the one we care and we select

I think we can throw a lot of the setup away.

e2e/portalicious/pages/RegistrationsPage.ts
```diff
@@ -180,6 +180,17 @@ class RegistrationsPage extends BasePage {
     return fullNameText;
   }
 
+  async validateStatusOfFirstRegistration({ status }: { status: string }) {
+    await this.page.waitForTimeout(200);
```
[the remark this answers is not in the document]
Piotrk39: In general we should not use hardcoded timeouts instead we should wait for loaders and other elements in DOM to be loaded. Unfortunately sometimes we are faced with a situation when everything is already loaded in the page and neither there are network changes nor any front-end elements. That is why we discourage it. In Portalicious when we operate on the table new elements seem to be loaded rarely thus occasional use of waitForTimeout.

Let me see if using this will change it: await this.table.waitForLoaded();

e2e/portalicious/pages/RegistrationsPage.ts
```diff
@@ -180,6 +180,17 @@ class RegistrationsPage extends BasePage {
...
+    await this.page.waitForSelector('table tbody tr td');
```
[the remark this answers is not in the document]
Piotrk39: Let me try your approach and I'll see if it works. Testing the table is generally more complicated then other parts of portalicious.

e2e/portalicious/pages/RegistrationsPage.ts
```diff
@@ -180,6 +180,17 @@ class RegistrationsPage extends BasePage {
...
+    const registrationStatus = await this.table.getCell(0, 3);
+    const statusText = (await registrationStatus.textContent())?.trim();
+    if (!statusText) {
+      throw new Error('Could not find full name in the table');
```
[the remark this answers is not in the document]
Piotrk39: Exactly

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Good point

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: More of a sanity check. I can change it if it is preferable.

Piotrk39:
> I see a lot of use of `resetDuplicateRegistrations`. Why? Do we need more than one registration? I don't see any asserts on the other registrations _not_ being changed? Can we remove that?

For me personally it is a sanity check but I can change this approach.

Piotrk39:
> For some tests the arranging/setup part feels long compared to the interesting part of the test (act, assert).
> 
> We already have a `seedIncludedRegistrations` helper. I think we can make these tests more focused if we make a `seedRegistrationsWithStatus` helper that takes a status and creates 1..n registrations with that status. The arrange part of each test then becomes just a call to that function.

Sure good idea!

Piotrk39:
> Is there a reason why we don't use the Arrange, Act, Assert pattern of comments in the E2E tests?

As by adding comments like:

//Arrange

//Act

//Assert

?

I can start adding them tho act and assert sometimes are present multiple times in tests and it would have to be repeated multiple times. But if this is a correct way o doing things I can start adding them.

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Probably good idea!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
nielsbom: I think the rule is: each test should test one thing. If we also want to test whether the table component can apply the same state change to multiple registrations that should be a _single_ and _separate_ test.

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
nielsbom: I haven't gone through all test files btw, I'm assuming the other test files follow the same pattern and should be updated similarly.

nielsbom:
> > Is there a reason why we don't use the Arrange, Act, Assert pattern of comments in the E2E tests?
> 
> As by adding comments like:
> 
> //Arrange
> 
> //Act
> 
> //Assert
> 
> ?
> 
> @nielsbom I can start adding them tho act and assert sometimes are present multiple times in tests and it would have to be repeated multiple times. But if this is a correct way o doing things I can start adding them.

I think a good reason for having "arrange", "act", "assert" in a test is to help the people writing the test to make sure the test is focused on a single action and not mixing multiple things. Would you agree? This then also makes it easier for people reading/updating the test to understand it more quickly.

e2e/portalicious/pages/RegistrationsPage.ts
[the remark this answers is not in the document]
Piotrk39: It is a new method so I was not aware of it, works like a charm I will use this one instead of a timeout. Thanks!

e2e/portalicious/pages/RegistrationsPage.ts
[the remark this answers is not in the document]
Piotrk39: So much better and cleaner!

e2e/portalicious/components/TableComponent.ts
```diff
@@ -83,7 +87,7 @@ class TableComponent {
   }
 
   async clearAllFilters() {
-    await this.clearAllFiltersButton.click();
+    await this.clearAllFiltersButton.first().click();
```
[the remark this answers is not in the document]
Piotrk39: Coment added!

e2e/portalicious/pages/RegistrationsPage.ts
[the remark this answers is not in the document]
Piotrk39: Done!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Done!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Done!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Not at all. Thanks!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Done!

e2e/portalicious/tests/ChangeRegistrationStatus/MovePasFromCompletedToIncluded.spec.ts
Piotrk39: Fixed it!

Piotrk39:
@nielsbom 

1. I took some liberty in clearing more than one status that as a final result was a duplicate.
2.  With the exception of ```Completed``` status that need more complex setup everything else was simplified with ```seedRegistrationsWithStatus``` or deleted. 
3. //Arrange //Act //Assert were also added.
4. I added ```validateToastMessageAndWait```function to validate the toast message and wait for the table to be active but without waiting for the full cycle of a toast to disappear which takes 6 seconds.

Piotrk39:
Moved to: https://github.com/global-121/121-platform/pull/6610

PR #6548, opened 2025-02-26
Deploy all clients & deploy all portals

tijsziere:
[AB#28101](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/28101) 

### DIFF
--- a/.github/workflows/deploy_all_clients.yml
+++ b/.github/workflows/deploy_all_clients.yml
@@ -1,0 +1,88 @@
+# See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
+
+name: 'Deploy Portal & Service to all clients'
+
+permissions:
+  actions: read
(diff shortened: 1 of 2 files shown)


RubenGeo (changes requested):
- We should add to the tech doc that you also need to update these ymls when adding a new client. 
- Is it somehow possible to put an array somewhere of all clients and loop over that? So it's easier to add new clients?

tijsziere:
> * We should add to the tech doc that you also need to update these ymls when adding a new client.
> * Is it somehow possible to put an array somewhere of all clients and loop over that? So it's easier to add new clients?

* will do
* I looked into this, there is no easy way of doing it. (There are some options around "matrix", but that does not cover our needs entirely) I'll propose to keep the item on the backlog for Elwin to do some improvements

PR #6550, opened 2025-02-27
Upgrade primeNg to 19.08

RubenGeo:
AB#XXXX 

Upgrading PrimeNG to 19.08. This is needed because we need what is in this [PR](https://github.com/primefaces/primeng/pull/17441) which is only released in version 19.06. This should allow the visible flag to work in nested items


RubenGeo:
Closed this PR because upgrading to 19.0.6 or 19.0.8 seems to break the expanding of rows in the activity log. Hopefully this is fixed in a later version or PrimeNG

PR #6557, opened 2025-02-27
feat: various UI improvements for RTL languages

aberonni:
AB#34003


arsforza (changes requested):

arsforza (approved):

interfaces/Portalicious/src/app/components/form/form-sidebar.component.html
```diff
@@ -1,7 +1,7 @@
 <p-drawer
   #sideBar
   [(visible)]="visible"
-  position="right"
+  [position]="rtlHelper.createPosition('right')()"
```
elwinschmitz, line 4: We could have the Helper-service act as a sort-of polyfill, so that we provide it with "start" and that it outputs "left" or "right", depending on the `isRtl`-state...

But maybe its better to leave the "buggy PrimeNG-implementation" in-place, to highlight the need for: https://github.com/orgs/primefaces/discussions/3649
(Or even create a PR for that issue ourselves...)

interfaces/Portalicious/src/app/components/page-layout/page-layout.component.html
```diff
@@ -40,7 +40,13 @@
                 } @else {
                   <span class="font-normal">{{ parentTitle }}</span>
                 }
-                <i class="pi pi-chevron-right mx-2 text-sm"></i>
+                <i
+                  class="pi mx-2 text-sm"
+                  [ngClass]="{
+                    'pi-chevron-right': !rtlHelper.isRtl(),
+                    'pi-chevron-left': rtlHelper.isRtl(),
```
elwinschmitz, line 47: Similar to the "polyfill"-approach mentioned above:

We could add a Tailwind-plugin/alias/class/thing like: `pi-chevron-start`
And have it output "the same as `pi-chevron-left`/`*-right`" depending on the `dir`-state.

That way, we don't have to "think about it"(and handle the RtlService, etc..) in our day-to-day-code...

elwinschmitz: We could also create a "Direction-agnostic-chevron"-component...
So we can contain this logic... and don't have to repeat it...

I'll move this into: https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/34095

PR #6578, opened 2025-03-03
chore: remove duplicate db-resets from integration tests

nielsbom:
[AB#34087](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/34087)

## Describe your changes

Some integration tests would reset the db twice before starting. Removed to cut down on test run time.


elwinschmitz (commented):
Hmm.. It is indeed better to remove the duplication/useless-db-actions...

But its now not 100% ensured that we're creating "independent tests"... i.e. theoretically you could write a test-scenario(an `it(...)`, that passes, because its dependent on data seeded for some other scenario...
The chance is pretty small/slim, but still...

I also don't really see an obvious alternative. ;)

RubenGeo (approved):

nielsbom:
> But its now not 100% ensured that we're creating "independent tests"... i.e. theoretically you could write a test-scenario(an it(...), that passes, because its dependent on data seeded for some other scenario...
> The chance is pretty small/slim, but still...
> 
> I also don't really see an obvious alternative. ;)

Having an `afterEach` that removes (tears down) any data we might be leaning on would prevent inter-test dependencies right? Let's discuss that somewhere else :-)

PR #6512, opened 2025-02-12
Add E2E test for Error in missing columnToMatch Configuration

Piotrk39:
[AB#33353](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/33353) 

## Describe your changes

Adds a test and new mock configuration for a program with missing columnToMatch Configuration in Excel FSP.

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [x] I have added tests wherever relevant
- [x] I have made sure that all automated checks pass before requesting a review
- [x] I do not need any deviation from our PR guidelines

## Portalicious preview deployment

This PR does not have any preview deployments yet.

PR #6582, opened 2025-03-04
Remove excel fail seed and use delete program fsp config property instead

RubenGeo:
[AB#34075](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/34075) 

## Describe your changes

Refactor of #6512 see [comment](https://github.com/global-121/121-platform/pull/6512#issuecomment-2697394949) in that PR 


services/121-service/src/scripts/enum/seed-script.enum.ts
aberonni, line 1: @RubenGeo could you add a comment to this enum elaborating on when we should / shouldn't add a new seed script? Maybe this would help avoid this in the future.

aberonni (changes requested):
Small addition but otherwise looks good and makes sense to me

aberonni (approved):

PR #6584, opened 2025-03-05
chore: refactor rtl-friendly chevron icons

aberonni:
[AB#34095](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/34095)

https://lively-river-04adce503-6584.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/services/rtl-helper.service.ts
+++ b/interfaces/Portalicious/src/app/services/rtl-helper.service.ts
@@ -20,12 +20,27 @@ const LOGICAL_VALUE_TO_PHYSICAL_VALUE_MAP: Record<
   },
 };
 
+const LTR_VALUE_TO_RTL_VALUE_MAP: Record<
+  Direction,
+  Record<PhysicalPosition, PhysicalPosition>
+> = {
+  ltr: {
+    left: 'left',
+    right: 'right',
+  },
+  rtl: {
+    left: 'right',
+    right: 'left',
+  },
+};
+
 @Injectable({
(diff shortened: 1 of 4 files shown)


interfaces/Portalicious/src/app/services/rtl-helper.service.ts
```diff
@@ -20,12 +20,27 @@ const LOGICAL_VALUE_TO_PHYSICAL_VALUE_MAP: Record<
...
+  Record<PhysicalPosition, PhysicalPosition>
+> = {
+  ltr: {
+    left: 'left',
+    right: 'right',
+  },
+  rtl: {
+    left: 'right',
+    right: 'left',
+  },
```
elwinschmitz, line 34:
```suggestion
  ltr: {
    backward: 'left',
    forward: 'right',
  },
  rtl: {
    backward: 'right',
    forward: 'left',
  },
```

To fully get rid of literal using "left" or "right" from our thinking, I'd like to use something like this... (But `start`/`end` would also be fine with me... but there's more room for confusion, I'd think?)

elwinschmitz (dismissed):
I could live with the current "left"/"right" usage... But I like an alternative better. :D

aberonni:
@elwinschmitz very cheeky self-approve here!! Tbh I am not a fan of forward/backward - I think it's more confusing from a dx perspective than "left/right". But hey, my LOGAF on this is not very high, so I'm fine with it. But very cheeky nonetheless!

PR #6638, opened 2025-03-21
chore: check for dead code in portalicious with knip

aberonni:
[AB#34711](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/34711)

I wanted to add [knip](https://knip.dev/) to avoid dead code such as this: https://github.com/global-121/121-platform/pull/6583/files#r2007873570

Turns out, there isn't a tool capable of detecting unused public class properties in typescript: https://github.com/microsoft/TypeScript/issues/29293

So it doesn't solve the use case I had for it. But still, we have it enabled on the 121-service, so I figured why not enable it here too and get rid of some dead code.

- [x] I have asked the design team to review these changes, or the changes do not touch the UI/UX

https://lively-river-04adce503-6638.westeurope.5.azurestaticapps.net


interfaces/Portalicious/src/app/domains/project/project.api.service.ts
```diff
@@ -27,7 +27,7 @@ import { Dto } from '~/utils/dto-type';
 
 const BASE_ENDPOINT = 'programs';
 
-export const filterableAttributesToIgnore = [
+const filterableAttributesToIgnore = [
```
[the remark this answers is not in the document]
aberonni: It wasn't anywhere else you silly robot

RubenGeo (approved):

interfaces/Portalicious/package-lock.json
```diff
@@ -3703,346 +3637,6 @@
...
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "os": [
-        "darwin"
-      ],
-      "engines": {
-        "node": ">=18"
-      }
-    },
```
elwinschmitz, line 3722: This will/should turn up locally for people with a mac, I'd think... And the win32-x64 one for people on Windows... and the linux-one for the use on GitHub-Actions...

I don't fully understand how these platform-specific-optional-dependencies are supposed to work, but it would be a waste not having a speedy build that is optimised for the platforms we use...

interfaces/Portalicious/package-lock.json
```diff
@@ -4917,85 +4511,15 @@
...
       "cpu": [
-        "x64"
+        "arm64"
       ],
       "dev": true,
       "license": "MIT",
       "optional": true,
       "os": [
-        "win32"
+        "darwin"
```
elwinschmitz, line 4522: Isn't this package-lock-file now only 'optimised' for use on a Mac?

interfaces/Portalicious/package-lock.json
```diff
@@ -13012,6 +11798,117 @@
...
+        "minimist": "^1.2.8",
+        "picocolors": "^1.1.0",
+        "picomatch": "^4.0.1",
+        "pretty-ms": "^9.0.0",
+        "smol-toml": "^1.3.1",
+        "strip-json-comments": "5.0.1",
+        "summary": "2.1.0",
+        "zod": "^3.22.4",
+        "zod-validation-error": "^3.0.3"
+      },
```
elwinschmitz, line 11838: Nice that a tool that's purpose is to "remove unnecessary dependencies", has _this many_ dependencies... 🧐

interfaces/Portalicious/src/app/services/registrations-table-column.service.ts
```diff
@@ -30,15 +30,15 @@ import {
 } from '~/services/registration-attribute.service';
 import { TranslatableStringService } from '~/services/translatable-string.service';
 
-export const FILTERABLE_ATTRIBUTES_LABELS: Record<string, string> = {
+const FILTERABLE_ATTRIBUTES_LABELS: Record<string, string> = {
```
elwinschmitz, line 33: I thought some ESLint-rule was already taking care of these "unused variables"...

Good to know it doesn't. ;)

elwinschmitz (commented):
Looking at the output of `knip`:
https://github.com/global-121/121-platform/actions/runs/14054795277/job/39351804331#step:7:1

```
Unused dependencies (2)
@angular/platform-browser-dynamic  package.json:40:6
primeicons                         package.json:56:6

Unused devDependencies (2)
autoprefixer  package.json:78:6
postcss       package.json:98:6

Unlisted dependencies (2)
@angular-eslint/builder  angular.json    
@eslint/js               eslint.config.js
```

I wonder how accurate this is? And/or how much attention we need to pay to these warnings...

As I assume `autoprefixer` is used by `postcss` and that is in turn used by Tailwind, somehow.. (at least thats what the config says..)

Also, if we don't use the `primeicons`, can we just remove the dependency?

interfaces/Portalicious/package-lock.json
aberonni: I'm just running npm i and letting node figure it out. Do we really need to be reviewing npm's code?

interfaces/Portalicious/package-lock.json
aberonni: I doubt it. To be on the safe side, I re-ran the npm install scripts from scratch rather than tweaking what was there before. Maybe this has made a difference. But tbh, I'm not sure this is something we should be worrying about (I might be wrong).

interfaces/Portalicious/package-lock.json
elwinschmitz: Well, my assumption was that `knip` was making changes to it. (Given that it was removing all kinds of dependencies marked `optional`.)

But my main concern is that we get into a constant back-and-forth with changes to these dependencies/lock-files.

In the past, I've tried to adjust the `install`-commands in several places to reduce the number of optional-and-unused dependencies; But it takes only 1 other developer that runs `npm i some-fancy-tool-thingy` and doesn't check that the changes in the lock-file are not _only-the-things-they-want-to-add_ to get back to square 1...

And also, some things will cause "local changes" by "just switching branches" for some people(=npm configs/VSCode extensions/muscle-memory-of-running-npm-i)... which is the kind of confusion these lock-files are supposed to relieve us from...

interfaces/Portalicious/package-lock.json
elwinschmitz: Now the lock-file looks like there are only "additions from the new tool", so that is good in one sense...

But I thought that `knip` also applied some 'optimisations' to our dependency-tree/lock-file... so those are now hereby reverted... So I hope it won't complain about that the next time it runs... (see my other comment above)

elwinschmitz (approved):
If it turns out to be noisy, we can always tweak some of the rules in the config-file later...

interfaces/Portalicious/package-lock.json
aberonni: > Well, my assumption was that knip was making changes to it. (Given that it was removing all kinds of dependencies marked optional.)

Maybe it did, because under the hood it didn't run `npm install` but something else during the setup. But I wouldn't say we need to worry about it.

interfaces/Portalicious/package-lock.json
aberonni: > But I thought that knip also applied some 'optimisations' to our dependency-tree/lock-file

Nope. Or at least, it's not supposed to, or I don't expect it to.

PR #6689, opened 2025-04-01
Improve error for cronjob authentication failure

nielsbom:
[AB#34950](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/34950) 

## Describe your changes

Improve error for cronjob authentication failure.

I added some unit tests and an integration test around cronjobs. I had to restructure some code to allow for easier testing.

Also some refactoring mostly renaming.

- [x] I have asked the design team to review these changes, or the changes do not touch the UI/UX

### DIFF
--- a/services/121-service/src/cronjob/cronjob.service.ts
+++ b/services/121-service/src/cronjob/cronjob.service.ts
@@ -159,7 +159,8 @@ export class CronjobService {
     return { url, responseStatus: response.status }; // Only used for testing purposes; this method is then called from the controller
   }
 
-  public getAllMethodNames(): string[] {
+  // Used for testing and triggering through Swagger UI.
+  public getAllCronJobMethodNames(): string[] {
     const prototype = Object.getPrototypeOf(this);
     const methodNames = Object.getOwnPropertyNames(prototype)
       .filter((name) => {
@@ -170,8 +171,7 @@ export class CronjobService {
           name !== 'constructor'
         );
       })
-      // Filter out this method.
-      .filter((name) => name !== 'getAllMethodNames');
+      .filter((name) => name.startsWith('cron'));
     return methodNames;
   }
 }
--- a/services/121-service/src/cronjob/cronjob.service.ts
+++ b/services/121-service/src/cronjob/cronjob.service.ts
@@ -33,7 +33,8 @@ export class CronjobService {
     ),
   })
   public async cronCancelByRefposIntersolve(cronJobMethodName): cronReturn {
-    const { baseUrl, headers } = await this.initService(cronJobMethodName);
+    const { baseUrl, headers } =
+      await this.prepareCronJobRun(cronJobMethodName);
     // This function periodically checks if some of the IssueCard calls failed and tries to cancel them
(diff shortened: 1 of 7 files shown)


services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -189,8 +140,51 @@ export class CronjobService {
...
+  }
+
+  // Separate function: easier to test.
+  private async getHeaders(): Promise<Header[]> {
+    const accessToken = await this.axiosCallsService.getAccessToken();
+    return this.axiosCallsService.accesTokenToHeaders(accessToken);
+  }
+
+  // Separate function: easier to test.
+  private async callEndpoint(
```
RubenGeo, line 169: This is a nice impovement 👍

services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -1,184 +1,135 @@
...
+    }
+  | false
+>;
+
 @Injectable()
 export class CronjobService {
   private httpService = new CustomHttpService(new HttpService());
   private axiosCallsService = new AxiosCallsService();
+  private headers: Header[];
+  private baseUrl: string;
```
RubenGeo, line 29: I am not sure if headers and baseUrl should be a class attribute if they are set again every time a cronjob is ran. So for example if there was like a `getHeaders` function that would get the headers from the class attribute and re-authenticates if they are expired it would make sense to me to store it as a class attribute as than the 'headers' are shared between cronjob function. However if they are re-initiated every-time it think it's clearer if they are the return of a function instead of a class attribute.

I think the baseurl could be class attribute but than I think it makes more sense to set it 'onModuleInit'. But I think it's also fine to just call the getBaseUrl method everytime.

What do you think?

services/121-service/src/cronjob/cronjob.service.ts
```diff
@@ -189,8 +140,51 @@ export class CronjobService {
...
         );
       })
-      // Filter out this method.
-      .filter((name) => name !== 'getAllMethodNames');
+      .filter((name) => name.startsWith('cron'));
     return methodNames;
   }
+
+  // If not initialized correctly no use running the cronjobs.
+  private async initService(cronjobName: string): Promise<void> {
```
RubenGeo, line 148: I would expect a function called 'initService' to run when the app starts up. Maybe it could be named differently or keep the headers and baseurl functionality in 2 functions?

RubenGeo (changes requested):
Great improvements 👍

services/121-service/src/cronjob/cronjob.service.ts
nielsbom: Agreed, I'll push some changes to remove these class properties. I've had to change some tests too to be able to do this.

services/121-service/src/cronjob/cronjob.service.ts
nielsbom: Renamed it to `prepareCronJobRun(cronjobName)`.

RubenGeo (approved):
🚀

PR #6694, opened 2025-04-04
Get registration by numeric of date range in portalicious

RubenGeo:
[AB#35058](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/35058) 

- [ ] I have asked the design team to review these changes, or the changes do not touch the UI/UX

https://lively-river-04adce503-6694.westeurope.5.azurestaticapps.net

### DIFF
--- a/interfaces/Portalicious/src/app/domains/project/project.api.service.ts
+++ b/interfaces/Portalicious/src/app/domains/project/project.api.service.ts
@@ -35,6 +35,7 @@ const filterableAttributesToIgnore = [
   // Below attributes are already hardcoded in the table columns service
   'programFinancialServiceProviderConfigurationName',
   'fullName',
+  'created',
   'registrationCreatedDate',
 ];
 
--- a/interfaces/Portalicious/src/app/services/paginate-query.service.spec.ts
+++ b/interfaces/Portalicious/src/app/services/paginate-query.service.spec.ts
@@ -113,10 +113,24 @@ describe('PaginateQueryService', () => {
       const result =
         // @ts-expect-error accessing a private method for unit testing purposes
         service.convertPrimeNGFilterMetadataToValueAndOperator(filterMetadata);
-      expect(result).toEqual({
-        value: '2025-04-03T00:00:00.000Z,2025-04-03T23:59:59.999Z',
-        operator: FilterOperator.BTW,
-      });
+
+      expect(result).toBeDefined();
+
+      // Early return if result is undefined (won't happen due to the expect above, but makes TypeScript happy)
+      if (!result) return;
+
+      // Check operator is correct
+      expect(result.operator).toBe(FilterOperator.BTW);
+
+      // Check value is comma-separated string containing two valid ISO dates
+      const dateStrings = result.value.split(',');
+      expect(dateStrings.length).toBe(2);
+
+      // Validate both are valid ISO date strings
+      const startDate = new Date(dateStrings[0]);
+      const endDate = new Date(dateStrings[1]);
+      expect(startDate.toISOString()).toBe(dateStrings[0]);
+      expect(endDate.toISOString()).toBe(dateStrings[1]);
     });
   });
 
@@ -269,10 +283,10 @@ describe('PaginateQueryService', () => {
       const startOfDayLocal = new Date(date.setHours(0, 0, 0, 0));
       const endOfDayLocal = new Date(date.setHours(23, 59, 59, 999));
       const expectedStartOfDayUtc = new Date(
-        startOfDayLocal.getTime() - clientTimeZoneOffset,
+        startOfDayLocal.getTime() + clientTimeZoneOffset,
       ).toISOString();
       const expectedEndOfDayUtc = new Date(
-        endOfDayLocal.getTime() - clientTimeZoneOffset,
+        endOfDayLocal.getTime() + clientTimeZoneOffset,
       ).toISOString();
 
       // @ts-expect-error accessing a private method for unit testing purposes
@@ -288,7 +302,7 @@ describe('PaginateQueryService', () => {
 
       const endOfDayLocal = new Date(date.setHours(23, 59, 59, 999));
       const expectedEndOfDayUtc = new Date(
-        endOfDayLocal.getTime() - clientTimeZoneOffset,
+        endOfDayLocal.getTime() + clientTimeZoneOffset,
       ).toISOString();
 
       // @ts-expect-error accessing a private method for unit testing purposes
@@ -307,7 +321,7 @@ describe('PaginateQueryService', () => {
 
       const startOfDayLocal = new Date(date.setHours(0, 0, 0, 0));
       const expectedStartOfDayUtc = new Date(
-        startOfDayLocal.getTime() - clientTimeZoneOffset,
+        startOfDayLocal.getTime() + clientTimeZoneOffset,
       ).toISOString();
 
       // @ts-expect-error accessing a private method for unit testing purposes
--- a/interfaces/Portalicious/src/app/services/paginate-query.service.spec.ts
+++ b/interfaces/Portalicious/src/app/services/paginate-query.service.spec.ts
@@ -1,6 +1,7 @@
 import { HttpParams } from '@angular/common/http';
 import { TestBed } from '@angular/core/testing';
 
+import { endOfDay, startOfDay } from 'date-fns';
 import { FilterMatchMode, FilterMetadata } from 'primeng/api';
 import { TableLazyLoadEvent } from 'primeng/table';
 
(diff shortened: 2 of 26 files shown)


interfaces/Portalicious/src/app/services/paginate-query.service.ts
```diff
@@ -100,6 +113,35 @@ export class PaginateQueryService {
...
+    endOfDay: Date;
+  } {
+    const startOfDay = new Date(date);
+    startOfDay.setHours(0, 0, 0, 0);
+    startOfDay.setMinutes(
+      startOfDay.getMinutes() - startOfDay.getTimezoneOffset(),
+    );
+    const endOfDay = new Date(date);
+    endOfDay.setHours(23, 59, 59, 999);
+    endOfDay.setMinutes(endOfDay.getMinutes() - endOfDay.getTimezoneOffset());
```
elwinschmitz, line 141: We're only handling UTC-timestamps, right?

So maybe we should be using the `setUTCHours`-methods, so we don't have to juggle with timezones...?

Or could be just use `T00:00:00:000Z` (as start) and `T23:59:59:999Z` (as end) after the `date`?

services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -111,21 +116,27 @@ export class RegistrationsPaginationService {
...
-      .map((r) => r.name)
-      .filter((r) => r !== DefaultRegistrationDataAttributeNames.phoneNumber); // Phonenumber is already in the registration table so we do not need to filter on it twice
+    const relationsWithoutPhoneNumber =
+      programRegistrationAttributeRelations.filter(
+        (r) => r.name !== DefaultRegistrationDataAttributeNames.phoneNumber,
+      );
+    const relationNamesWithoutPhonenumber = relationsWithoutPhoneNumber.map(
+      (r) => r.name,
+    );
+    // Phonenumber is already in the registration table so we do not need to filter on it twice
```
elwinschmitz, line 128: I think some lines have been auto-formatted to the wrong position...

```suggestion
    const relationsWithoutPhoneNumber =
      programRegistrationAttributeRelations.filter(
        // <comment-should-go-here?>
        (r) => r.name !== DefaultRegistrationDataAttributeNames.phoneNumber,
      );
    // <or comment-should-go-here?>
    const relationNamesWithoutPhonenumber = relationsWithoutPhoneNumber.map(
      (r) => r.name,
    );
```

services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -370,58 +381,68 @@ export class RegistrationsPaginationService {
...
   ): Record<string, FilterOperator[] | true> {
     const filterObject = {};
-    for (const name of registrationDataNamesProgram) {
-      filterObject[name] = allowedFilterOperators;
+    for (const r of attributeRelations) {
+      if (r.type === RegistrationAttributeTypes.numeric) {
+        filterObject[r.name] = AllowedFilterOperatorsNumber;
+      } else {
+        filterObject[r.name] = AllowedFilterOperatorsString;
+      }
```
elwinschmitz, line 407:
```suggestion
    for (const relation of attributeRelations) {
      if (relation.type === RegistrationAttributeTypes.numeric) {
        filterObject[relation.name] = AllowedFilterOperatorsNumber;
      } else {
        filterObject[relation.name] = AllowedFilterOperatorsString;
      }
```
?
I thought we'd have a ESLint-rule for "one letter variables"...

services/121-service/src/registration/services/registrations-pagination.service.ts
```diff
@@ -462,6 +488,24 @@ export class RegistrationsPaginationService {
...
+        return queryBuilder.andWhere(
+          `${uniqueJoinId}.${columnName}::numeric > :value${uniqueJoinId}`,
+          { [`value${uniqueJoinId}`]: value },
+        );
+      case 'lessThan':
+        return queryBuilder.andWhere(
+          `${uniqueJoinId}.${columnName}::numeric < :value${uniqueJoinId}`,
+          { [`value${uniqueJoinId}`]: value },
+        );
+      case 'between':
```
elwinschmitz, line 501: Would it be possible for some value to "fall in between" these scenario's?

So for timestamps of `<date>T00:00:00:000Z` exactly?

So wouldn't it be necessary to have `>=`/`<=` instead of only `>`/`<`?

And/Or is the below `BETWEEN`-command/operator already including "the edges"?

services/121-service/test/helpers/registration.helper.ts
```diff
@@ -13,6 +15,56 @@ import {
...
+    whatsappPhoneNumber:
+      whatsappPhoneNumber !== undefined ? whatsappPhoneNumber : '14155236666',
+    addressStreet:
+      addressStreet !== undefined ? addressStreet : 'Default Street',
+    addressHouseNumber:
+      addressHouseNumber !== undefined ? addressHouseNumber : 1,
+    addressHouseNumberAddition: '',
+    addressPostalCode:
+      addressPostalCode !== undefined ? addressPostalCode : '1234AB',
+    addressCity: addressCity !== undefined ? addressCity : 'Default City',
```
elwinschmitz, line 64:
```suggestion
    referenceId: referenceId ?? `ref-${Math.random().toString(36).substring(2, 10)}`,
    preferredLanguage: LanguageEnum.en,
    paymentAmountMultiplier: paymentAmountMultiplier ?? 1,
    fullName: fullName ?? 'Default Name',
    phoneNumber: phoneNumber ?? '14155236666',
    programFinancialServiceProviderConfigurationName: programFinancialServiceProviderConfigurationName ?? FinancialServiceProviders.intersolveVisa,
    whatsappPhoneNumber: whatsappPhoneNumber ?? '14155236666',
    addressStreet: addressStreet ?? 'Default Street',
    addressHouseNumber: addressHouseNumber ?? 1,
    addressHouseNumberAddition: '',
    addressPostalCode: addressPostalCode ?? '1234AB',
    addressCity: addressCity ?? 'Default City',
```

Curious; would the above have the same effect?

elwinschmitz (changes requested):
Looked at "only the front-end bits...

NOT at the migration...

interfaces/Portalicious/src/app/services/paginate-query.service.ts
elwinschmitz: On second thought...

In the `e2e/`-folder/tests, we use the package `date-fns`: https://date-fns.org/

Maybe we should use that everywhere?

interfaces/Portalicious/src/app/domains/project/project.api.service.ts
```diff
@@ -35,6 +35,7 @@ const filterableAttributesToIgnore = [
   // Below attributes are already hardcoded in the table columns service
   'programFinancialServiceProviderConfigurationName',
   'fullName',
+  'created',
   'registrationCreatedDate',
```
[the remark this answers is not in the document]
RubenGeo: yes, I put a TODO comment for that in the registrationViewEntity in the backend

interfaces/Portalicious/src/app/services/paginate-query.service.spec.ts
```diff
@@ -213,4 +255,44 @@ describe('PaginateQueryService', () => {
...
   });
+
+  describe('getDateFilterValue', () => {
+    it('should return start and end of the day for EQUALS match mode', () => {
+      const date = new Date('2025-04-03T12:00:00Z');
+      // @ts-expect-error accessing a private method for unit testing purposes
+      const result = service.getDateFilterValue(date, FilterMatchMode.EQUALS);
+      expect(result).toBe('2025-04-03T00:00:00.000Z,2025-04-03T23:59:59.999Z');
+    });
+
```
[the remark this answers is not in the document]
RubenGeo: Good one!

services/121-service/src/registration/services/registrations-pagination.service.ts
RubenGeo: But this situation is still good right? The filter is supposed to work for greater than and not greater than or equal. Maybe we should call about this if we feel like we are talking past each other

services/121-service/test/helpers/registration.helper.ts
RubenGeo: Almost, however I want to also explicitly be able to set variables to null. So that the default value is not used. For example import addressHouseNumber as null instead of 1.

RubenGeo: This file is also a bit out of scope for this PR. So resolving it for now

services/121-service/src/registration/services/registrations-pagination.service.ts
elwinschmitz: In _this specific_ comment, I was talking about the position of the `// comment` only;

Not the variable names..
But, yes, I agree/understand your point about those. ;)

services/121-service/test/helpers/registration.helper.ts
elwinschmitz: Yeah, the "explicitly `null`" scenario cannot be done with this short `??` syntax no.. too bad..

interfaces/Portalicious/src/app/utils/local-time-to-utc-time.ts
```diff
@@ -0,0 +1,10 @@
+export const localTimeToUtcTime = (localDate: Date): Date =>
+  new Date(
+    localDate.getUTCFullYear(),
+    localDate.getUTCMonth(),
+    localDate.getUTCDate(),
+    localDate.getUTCHours(),
+    localDate.getUTCMinutes(),
+    localDate.getUTCSeconds(),
+    localDate.getUTCMilliseconds(),
+  );
```
elwinschmitz, line 10: Ah. not that I've checked it separately, this is better then the "shorter" version:

`new Date(localDate.toUTCString()).toISOString();`

As that 👆 doesn't take the miliseconds into account.

elwinschmitz (approved):

aberonni:
@RubenGeo we noticed during standup that this was ready for review by design. But then:

image

This shouldn't have been merged without review from design; instead this should've been reviewed by design on the test URL. No need to revert or anything, but good to keep in mind for next time. The reason is because we want to avoid UI/UX changes to end up in production without a review from design, as much as possible. 

(If there is a reason to circumvent the guidelines, or a reason for this to be urgent and ignore the checklist in the description, I would expect that to be explained in the PR description itself.)

PR #6700, opened 2025-04-07
E2E test for advanced filtering by date

Piotrk39:
[AB#35094](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/35094) 

## Describe your changes

Adds additional test steps for existing e2e tests "Filter by date" to allocate advanced filtering: Equals, less then, more then etc.

- [ ] I have asked the design team to review these changes, or the changes do not touch the UI/UX

https://lively-river-04adce503-6700.westeurope.5.azurestaticapps.net

### DIFF
--- a/e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
+++ b/e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
@@ -47,19 +47,22 @@ test('[34947] Filter registrations by Date selection', async ({ page }) => {
   const registrations = new RegistrationsPage(page);
   const tableComponent = new TableComponent(page);
   // Act & Assert
-  await test.step('Filter registrations columns by date', async () => {
+  await test.step('Filter registration created column by "Equals" date', async () => {
     // Filter Registration column by date
     await tableComponent.filterColumnByDate({
       columnName: 'Registration created',
       day,
+      range: 'Equals',
     });
     await tableComponent.validateAllRecordsCount(4);
+    await tableComponent.clearAllFilters();
     // Filter by different date
     const diffDay = currentDate.getDate() === 1 ? 2 : 1;
     if (diffDay !== undefined) {
       await tableComponent.filterColumnByDate({
         columnName: 'Registration created',
         day: diffDay,
+        range: 'Equals',
       });
     }
     await registrations.validateRegistrationIsNotPresent();
--- a/e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
+++ b/e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
@@ -14,7 +14,6 @@ import {
 
 import TableComponent from '@121-e2e/portalicious/components/TableComponent';
 import LoginPage from '@121-e2e/portalicious/pages/LoginPage';
-import RegistrationsPage from '@121-e2e/portalicious/pages/RegistrationsPage';
 
 // Get current date information
 const currentDate = new Date();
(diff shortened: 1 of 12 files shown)


e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
```diff
@@ -47,21 +47,46 @@ test('[34947] Filter registrations by Date selection', async ({ page }) => {
   const registrations = new RegistrationsPage(page);
```
RubenGeo, line 47: Maybe it makes sense to also rename registrations here to registrationPage

e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
```diff
@@ -47,21 +47,46 @@ test('[34947] Filter registrations by Date selection', async ({ page }) => {
...
+  });
+
+  await test.step('Filter registration created column by "Greater than" date', async () => {
+    // Filter Registration column by date
+    await tableComponent.filterColumnByDate({
+      columnName: 'Registration created',
+      day,
+      range: 'Greater than',
+    });
+    await registrations.validateRegistrationIsNotPresent();
```
RubenGeo, line 90: `validateRegistrationIsNotPresent` actually check whether the string `No results found` is found on the page. In our frontend code this is the default behaviour of any empty query table. So I was thinking it may make more sense to move this function to the `TableComponent` file. 

Also maybe we should have a different name for that function? Maybe assertEmptyTableState() or validateEmptyTableMessage()

e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
```diff
@@ -47,21 +47,46 @@ test('[34947] Filter registrations by Date selection', async ({ page }) => {
...
         columnName: 'Registration created',
         day: diffDay,
+        range: 'Equals',
       });
     }
     await registrations.validateRegistrationIsNotPresent();
+    await tableComponent.clearAllFilters();
+  });
+
+  await test.step('Filter registration created column by "Less than" date', async () => {
```
RubenGeo, line 72: Maybe it makes sense to have a step where we test a filter which does returns registartions. For example test for an equal of the current day. 

```
await tableComponent.filterColumnByDate({
        columnName: 'Registration created',
        day: currentDay,
        range: 'Equals',
      });
``

RubenGeo (changes requested):
I love that so little changes are needed to the e2e to make the old date filters work for the new setup. I think it says something about our code quality

In the PR I would also expect some changes to the numeric filter file `FilterRegistrationsByNumberInput.spec`. As numeric columns now also have range filtering, which I think we should also cover in a test. But some reason those tests did not fail in the PR run not sure why?

e2e/portalicious/tests/ViewFilterAndSortRegistrations/FilterRegistrationsByDate.spec.ts
Piotrk39: Done!

Piotrk39:
> I love that so little changes are needed to the e2e to make the old date filters work for the new setup. I think it says something about our code quality
> 
> In the PR I would also expect some changes to the numeric filter file `FilterRegistrationsByNumberInput.spec`. As numeric columns now also have range filtering, which I think we should also cover in a test. But some reason those tests did not fail in the PR run not sure why?

@RubenGeo I'll refactor those e2e tests in a separate PR. The reason for it not failing is that we did not change the locator for the number input and that default filter is 'Equals' therefore the behaviour of the filter remained unchanged.

Piotrk39:
Moved to: https://github.com/global-121/121-platform/pull/6711

PR #6896, opened 2025-05-26
feat: Use t3-env to validate ENV-vars (in Mock-Service)

elwinschmitz:
AB#35885

## Describe your changes

During start-up of the Mock-Service, the currently set ENV-vars will be checked/validated if they match the configured schema/definition. (see `services/mock-service/src/env.ts`)

This will probably not cause any issue with your current state of `services/.env`, but otherwise it should throw specific errors and solutions how to fix it.

## Portal preview-deployment


services/mock-service/.eslintrc.js
diderikvw, line 1: Why would or should Linter configuration be (very) different between the 121-service and mock-service as it seems to be now?

services/mock-service/package-lock.json
diderikvw, line 1: That is a lot more dependencies. We probably all need them, otherwise you would not have added them :) I would like to have a brief explainer.

And I am thinking a debrief/show&tell of "the new and improved way to do env variables" for the entire dev team.

And a refactor item to update all existing code that gets env variables? Or is that included in this item?

services/mock-service/src/config.ts
```diff
@@ -38,8 +40,4 @@ export const API_PATHS = {
     'financial-service-providers/safaricom/timeout-callback',
 };
 
-export const EXTERNAL_API_ROOT = `${
-  DEVELOPMENT
-    ? `http://121-service:${process.env.PORT_121_SERVICE}/`
-    : process.env.EXTERNAL_121_SERVICE_URL
-}api`;
+export const EXTERNAL_API_ROOT = `${env.EXTERNAL_121_SERVICE_URL}api`;
```
diderikvw, line 43: Needs a heads-up to devs that they possibly need to update their .env locally?

services/mock-service/src/env.ts
diderikvw, line 1: So here we define the env variables that the service actually needs from .env?

services/mock-service/src/fsp-integration/intersolve-visa/intersolve-visa.mock.service.ts
```diff
@@ -255,8 +256,7 @@ export class IntersolveVisaMockService {
       res.status = HttpStatus.NOT_FOUND;
       res.statusText = 'NOT_FOUND';
     }
-    // must match \"([+]){1}([1-9]){1}([0-9]){5,14}\
-    const mobileNumberRegex = new RegExp('^([+]){1}([1-9]){1}([0-9]){5,14}$');
+    const mobileNumberRegex = /^\+[1-9]\d{5,14}$/;
```
diderikvw, line 259: A new way to write regular expressions?

diderikvw (approved):

services/mock-service/.eslintrc.js
elwinschmitz: Well... because of reasons:

- History!  
  - A lot of these plugins where added on a "once you need it"-basis. So for example, as the codebase might not use any RegExps, there's no real need to add the plugin. But once you use the 1st, we'd install the plugin.

  - As the code for each service evolved with different lint-rules, they cannot comply at once to a "combined set of rules".  
  But it does make sense to have the same level of quality for both the 121-Service AND the Mock-Service.

- The 'scope' of these separate services/interfaces/tests in our platform is all quite different, so they require different tools to work with them.  
  Some ESLint-plugins, like `eslint-plugin-n` only make sense in a NodeJs-backend context, so no need to check/fix issues in our frontend code.

services/mock-service/src/config.ts
elwinschmitz: Theoretically, the heads-up will be given by the `t3-env`-tool on startup of the Mock-Service...

That should throw the error about a possibly faulty value of `EXTERNAL_121_SERVICE_URL`, and thus pointing to the 'requirements' defined in `src/env.ts`.

So, the idea is, that no "reading the documentation first" would be required.

---

Also, technically the values `http://localhost:3000` and `http://121-service:3000` should work the same. (But in practice, as seen on this PRs' test-runs, not for all environments/tools. 😢 )
In this specific refactoring, I tried to minimise the number of:
- ENV-variables that need to be set/handled. 
- "variants of some URL of the 121-service".
- "arbitrary switches based on `NODE_ENV`".

services/mock-service/.eslintrc.js
diderikvw: Ok, understood. Will you align lint plug-ins and settings between 121-service, mock service and portal wherever possible? Or if not, should we create an item to do that "later" and/or add some TODO comments?

diderikvw (approved):

k6/run-all-tests.sh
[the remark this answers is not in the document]
elwinschmitz: The changes are only Bash-related; So no real functional changes related to `k6`.

services/mock-service/.eslintrc.js
elwinschmitz: There is already: AB#27407 so that'll cover it.

elwinschmitz:
@diderikvw I've removed/spun-off some changes and simplified it a bit; So less to go wrong now. ;)

diderikvw (approved):

PR #6969, opened 2025-06-26
Strip unknown value on dto in controller

RubenGeo:
[AB#36875](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/36875) 

## Describe your changes

Update to validation of dtos of incoming requests. This change ensures unknown values are stripped from the body of incoming requests. 

Post and Patch registrations is excluded as these can contain dynamic properties. As well as any callback url as we do not want to throw errors if external parties change their callback

[Discussion in teams](https://teams.microsoft.com/l/message/19:ce634fd37f34477d9f98391f70e1f0ef@thread.skype/1751362483571?tenantId=d3ab9790-6ae2-4bd8-aa5e-02864483e7c7&groupId=55f2fa49-7d9b-4c63-b616-b986f3136c3a&parentMessageId=1751362483571&teamName=510%20-%20CVA&channelName=121%20Development&createdTime=1751362483571) about whether to strip unknown values or throw upon receiving unknown values. Decision has been made to strip them. 

## Portal preview-deployment


RubenGeo:
@reviewer

Based on the [NestJS documentation](https://docs.nestjs.com/techniques/validation#using-the-built-in-validationpipe):

	•	whitelist: If set to true, the validator will strip any properties from the input object that don’t have corresponding validation decorators.
	•	forbidNonWhitelisted: If set to true, instead of silently stripping unknown properties, the validator will throw an exception.

We could consider enabling whitelist: true for now. This provides a safer baseline by automatically removing unexpected properties from incoming requests without breaking them outright. It’s less disruptive and allows us to identify unvalidated fields more gradually.

That said, enabling forbidNonWhitelisted: true would be stricter and more explicit, surfacing potential issues earlier by rejecting requests with unknown fields. We could consider that as a follow-up step once we’re confident in the DTO definitions and want to enforce stricter validation.

My gut feel is to immediately go with forbidNonWhitelisted, than we immediately bite through the sour apple.

RubenGeo:
We could also consider turning below code snippet into a custom decorator
```
  @UsePipes(
    new ValidationPipe({ whitelist: false, forbidNonWhitelisted: false }),
)
```

But I first want to hear opinions on this before I put the effort in that

RubenGeo:
These changes also throw an error if a dto is DTO is missing the any class-validator decorators for a property. [See](https://github.com/global-121/121-platform/pull/6969/commits/f6c0b9d942c8e12ffc8852068f0c2448facb7693) that commit of how to fix that. I still need to go over all dto's and apply that. But before I put effort into that this change still warants some discussion

services/121-service/src/payments/reconciliation/safaricom-reconciliation/safaricom-reconciliation.controller.ts
```diff
@@ -31,6 +33,9 @@ export class SafaricomReconciliationController {
     status: HttpStatus.CREATED,
     description: 'Notified transfer status',
   })
+  @UsePipes(
+    new ValidationPipe({ whitelist: false, forbidNonWhitelisted: false }),
+  ) // Registrations have dynamic attributes, so we cannot use whitelist
```
jannisvisser, line 38: The comment does not fit here. (and in more places I think)

I'll make a task to also do this for onafriq.

jannisvisser (changes requested):
One small comment.

For the rest, I would also say to immediately go for forbidNonWhiteListed: true. We should be prepared for production errors in the time after release then probably.

I actually played around with these values also related to the onafriq callback as touched upon [here](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/36710#14929182). I somehow did not get this to work. In the sense that when I removed the validation decorators on the dto properties (IsNotEmpty), the callback started fialing because of unknown properties. And even after globally and/or locally setting whitelist: false and forbidNonWhitelisted: false, it kept giving that error. 

Anyway, that scenario is not the same as this one, as we're now actually setting both to true, but it does tell me that I don't completely understand the behaviour, which makes me a little more uncertain on what will happen after this change.

services/121-service/src/registration/validators/any-valid-body.validator.ts
```diff
@@ -0,0 +1,28 @@
...
+} from '@nestjs/common';
+
+const RawBody = createParamDecorator(
+  (_data: unknown, ctx: ExecutionContext): any => {
+    const request = ctx.switchToHttp().getRequest();
+    return request.body;
+  },
+);
+
+// This custom decorator is needed to allow external callbacks with unknown or extra properties
```
jannisvisser, line 14: it's not only used for external callbacks

services/121-service/src/registration/validators/any-valid-body.validator.ts
```diff
@@ -0,0 +1,28 @@
...
+// It applies a ValidationPipe with relaxed settings, so the global whitelist does not block valid payloads
```
jannisvisser, line 15: whiteList = true both here and globally. So what is the difference? In the previous version we were passing whiteList: false to the exception endpoints.

services/121-service/src/registration/registrations.controller.ts
```diff
@@ -360,12 +361,11 @@ export class RegistrationsController {
   })
   @ApiParam({ name: 'programId', required: true, type: 'integer' })
   @ApiParam({ name: 'referenceId', required: true, type: 'string' })
-  //Note: this endpoint must be placed below /programs/:programId/registrations/status to avoid conflict
```
jannisvisser, line 363: is this no longer true?

services/121-service/src/registration/validators/any-valid-body.validator.ts
```diff
@@ -0,0 +1,28 @@
...
+export const AnyValidBody = (...pipes: unknown[]) =>
+  RawBody(
+    new ValidationPipe({
+      validateCustomDecorators: true,
+      whitelist: true,
+      skipMissingProperties: true,
+      forbidUnknownValues: false,
+      transform: true,
+    }),
+    ...pipes,
```
jannisvisser, line 27: Did you test this functionally? And does it make sense to add a test that asserts scenarios for both correct / uncorrect payloads with both global/relaxed endpoint?

services/121-service/src/payments/reconciliation/onafriq-reconciliation/onafriq-reconciliation.controller.ts
```diff
@@ -30,12 +23,9 @@ export class OnafriqReconciliationController {
     description: 'Callback processed successfully.',
   })
   @HttpCode(HttpStatus.OK) // NOTE: Onafriq internally labels the callback as success on status 200
-  @UsePipes(
-    new ValidationPipe({ whitelist: false, forbidNonWhitelisted: false }),
-  )
   @Post('callback')
   public async processTransactionCallback(
-    @Body() onafriqTransactionCallback: OnafriqTransactionCallbackDto,
+    @AnyValidBody() onafriqTransactionCallback: OnafriqTransactionCallbackDto,
```
jannisvisser, line 28: use same comment as elsewhere?

jannisvisser (changes requested):

services/121-service/test/middlewares/any-body-decorator.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+class TestDto {
+  @IsString()
+  known: string;
+}
+
+@Controller('test')
+class TestController {
+  @Post('any-valid-body')
+  testAnyValidBody(@AnyValidBody() body: TestDto) {
```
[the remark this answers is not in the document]
RubenGeo: The onafriq endpoint and safaricom are at the moment only logging the extra properties and not storing it in a transaction. So it was very hard to use those for a test

services/121-service/src/payments/reconciliation/safaricom-reconciliation/dtos/safaricom-timeout-callback.dto.ts
```diff
@@ -1,6 +1,7 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { IsNotEmpty } from 'class-validator';
 
+// Does not need validation decorators on all properties, because we want to accept any payload structure from Safaricom, as we cannot control it.
```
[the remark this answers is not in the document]
RubenGeo: I think keep the @IsOptional is better because it's explicit. So added that to this and changed OriginatorConversationID to also be optional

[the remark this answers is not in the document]
RubenGeo: and throw an error if it does not exist

jannisvisser (approved):

PR #6971, opened 2025-06-26
Test flakiness: wait for events instead of changed registrations

RubenGeo:
[AB#36898](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/36898) 

In the 121-service docker logs of the [failed api tests](https://github.com/global-121/121-platform/actions/runs/15773000731/job/44461262205) I found the followig event related errors: 
image

My assumption is, is the we reset the database while our 121-service is still trying to store events. The registrations in the database are than already gone due to the reset. Which is why that error is thrown in the screenshot and the server crashes.

To fix this I changed the `waitForStatusChangeToComplete` to wait for the registration status change events to be created instead of waiting for the moment that registration.status changed. Because the events are created after the registrations status change.

Not sure if all these assumptions are correct or if this PR will solve the issues. I also don't think it will hurt to try it and conceptually I think this is an improvement anyway.

Edit: 

image

In this[ api test](https://github.com/global-121/121-platform/actions/runs/15632769116/job/44040722982) run I found an error on the insertion of event. I think this support my earlier made hypothesis that the db resets before events are inserted

## Portal preview-deployment

### DIFF
--- a/services/121-service/test/helpers/registration.helper.ts
+++ b/services/121-service/test/helpers/registration.helper.ts
@@ -146,18 +146,20 @@ export async function waitForDeleteRegistrations({
   maxWaitTimeMs?: number;
 }) {
   const startTime = Date.now();
+  const accessToken = await getAccessToken();
   while (Date.now() - startTime < maxWaitTimeMs) {
     // Get payment transactions
     let totalRegistrationSuccesfullyDeleted = 0;
+
     for (const referenceId of referenceIds) {
-      const getEventResponse = await getEvents({
+      const getEventsResponse = await getEvents({
         programId,
         fromDate: undefined,
         toDate: undefined,
         referenceId,
-        accessToken: await getAccessToken(),
+        accessToken,
       });
-      const deleteEvent = getEventResponse.body.find(
+      const deleteEvent = getEventsResponse.body.find(
         (event) =>
           event.type === EventEnum.registrationStatusChange &&
           event.attributes?.newValue === RegistrationStatusEnum.deleted,
--- a/services/121-service/test/helpers/registration.helper.ts
+++ b/services/121-service/test/helpers/registration.helper.ts
@@ -150,12 +150,13 @@ export async function waitForDeleteRegistrations({
     // Get payment transactions
     let totalRegistrationSuccesfullyDeleted = 0;
     for (const referenceId of referenceIds) {
-      const getEventResponse = await getEvents(
+      const getEventResponse = await getEvents({
         programId,
-        undefined,
-        undefined,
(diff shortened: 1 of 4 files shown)


services/121-service/test/helpers/registration.helper.ts
```diff
@@ -150,12 +150,13 @@ export async function waitForDeleteRegistrations({
...
-      const getEventResponse = await getEvents(
+      const getEventResponse = await getEvents({
         programId,
-        undefined,
-        undefined,
+        fromDate: undefined,
+        toDate: undefined,
         referenceId,
-      );
+        accessToken: await getAccessToken(),
```
elwinschmitz, line 158: This will (still) do a login-request for EVERY reference-id in the `referenceIds`-array! 😮

We could do this once (at line 148, for example) only...

services/121-service/test/helpers/registration.helper.ts
```diff
@@ -267,37 +268,6 @@ export function getRegistrations({
     .send();
 }
 
-export async function changeBulkRegistrationStatus({
```
elwinschmitz, line 270: Was this not used anymore anywhere? 🎉 :)

services/121-service/test/helpers/registration.helper.ts
```diff
@@ -150,12 +150,13 @@ export async function waitForDeleteRegistrations({
     // Get payment transactions
     let totalRegistrationSuccesfullyDeleted = 0;
     for (const referenceId of referenceIds) {
-      const getEventResponse = await getEvents(
+      const getEventResponse = await getEvents({
```
elwinschmitz, line 153: If we're nitpicking on "naming variables" anyway... Then lets also go for `getEventsResponse` :D

services/121-service/test/helpers/registration.helper.ts
```diff
@@ -391,17 +361,14 @@ export async function waitForStatusChangeToComplete(
   const startTime = Date.now();
   while (Date.now() - startTime < maxWaitTimeMs) {
     // Get payment transactions
```
elwinschmitz, line 363:
```suggestion
```

Not getting _that_ anymore. :)

elwinschmitz (changes requested):
Only small things. :)

services/121-service/test/helpers/registration.helper.ts
RubenGeo: yup

elwinschmitz (approved):

PR #6976, opened 2025-06-27
chore: send coverage data to QLTY

aberonni:
[AB#36889](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/36889)

I landed on QLTY as the tool of choice after following the trail from the related item's parent to spike [AB#23125](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/23125) "What would be a good tool to improve code quality?". This spike listed a few code quality tools, and on top of that I did some googling and comparison.

The first tool in the description (CodeClimate quality, which has since been renamed to QLTY) came out on top of the comparison:
- free for OSS
- can receive custom code coverage reports
- can notify devs on dip in coverage diff

Given that it's Good Enough For Now™ I went ahead and added some integrations for it. It should be easy to switch to something else if we ever decide to, now that we have the coverage reports generated and combined (which was the harder part).

**The coverage report that is being sent to QLTY can be seen [here](https://qlty.sh/gh/global-121/projects/121-platform/settings/coverage)**. If this PR is merged to main, then I expect the coverage integration to do More Things™ but we will only really see them after a merge has happened.

For the sake of debugging, it is also possible to download the reports manually as an artifact from a job run. For example, [here](https://github.com/global-121/121-platform/actions/runs/16022306494)

**Note**: one downside of this is that I've had to move the unit test job to the same workflow as the integration tests, because otherwise combining the two reports became more complex - with having to share artifacts between jobs etc..

The alternative to this would've been to use [server-side merging](https://docs.qlty.sh/coverage/merging#server-side-merging) (which I could still be persuaded to use) but the downside here is that it would be harder to replicate QLTY's information locally.

## Portal preview-deployment

https://happy-rock-0411d2003-6976.westeurope.3.azurestaticapps.net


.github/workflows/test_interface_portal.yml
```diff
@@ -54,11 +60,26 @@ jobs:
...
+          files: ${{ env.interfacePath }}/coverage/lcov.info
 
       - name: 'Check translations output-file'
         working-directory: ${{ env.interfacePath }}
         run: |
           npm run set-env-variables -- env=development
           npm run extract-i18n
           git diff --exit-code src/locale/messages.xlf
+
+      - name: Save coverage report
```
elwinschmitz, line 79: Why not directly AFTER the run of the `npm run test:coverage`-step?

.github/workflows/test_service_api.yml
```diff
@@ -43,15 +51,58 @@ jobs:
...
+        working-directory: ${{ env.workingDirectory }}
         run: |
-          docker compose exec 121-service npm run test:integration:all
+          npm install
+          npm run coverage
+
+      - uses: qltysh/qlty-action/coverage@v1
+        with:
+          oidc: true
+          add-prefix: services/121-service/
```
elwinschmitz, line 94:
```suggestion
          add-prefix: ${{ env.workingDirectory }}
```

services/121-service/package.json
```diff
@@ -29,6 +29,11 @@
     "test:integration:debug": "node --inspect-brk=0.0.0.0:9231 --require=tsconfig-paths/register --expose-gc node_modules/.bin/jest --config=jest.integration.config.js --runInBand --detectOpenHandles --logHeapUsage",
     "test:integration:watch": "npm run test:integration:all -- --watchAll",
     "test:integration:coverage": "npm run test:integration:all -- --coverage --coverageDirectory=coverage/integration",
+    "coverage:fix-paths": "find coverage -type f -name \"coverage-final.json\" -exec sed -i '' \"s|/home/node/app|$(pwd)|g\" {} \\;",
+    "coverage:merge": "istanbul-merge --out coverage/combined/coverage.json coverage/integration/coverage-final.json coverage/unit/coverage-final.json",
+    "coverage:report": "nyc report -t coverage/combined --report-dir coverage/combined --reporter=lcov --reporter=text",
+    "coverage": "npm run coverage:fix-paths && npm run coverage:merge && npm run coverage:report",
+    "open-coverage-report": "open coverage/combined/lcov-report/index.html",
```
elwinschmitz, line 36:
```suggestion
    "coverage:open-report": "open coverage/combined/lcov-report/index.html",
```

I like namespacing things. ;)

.github/workflows/test_service_static_checks.yml
```diff
@@ -1,27 +1,26 @@
 # See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions
 
-name: 'Test Service: Code'
+name: 'Test Service: Static Checks'
```
elwinschmitz, line 3: Actually nicer to have these separate indeed.

Maybe one day, we could set these to run "more often, quicker" and the others "only when PR is set to `ready-for-review` or something)...

.github/workflows/test_service_api.yml
```diff
@@ -43,15 +51,58 @@ jobs:
...
+          NODE_ENV: development
+          TWILIO_SID: AC
+          TWILIO_AUTHTOKEN: test
+          TWILIO_WHATSAPP_NUMBER: 15005550006
+          MOCK_TWILIO: TRUE
+          MOCK_INTERSOLVE: TRUE
+          MOCK_COMMERCIAL_BANK_ETHIOPIA: TRUE
+          MOCK_NEDBANK: TRUE
+          NEDBANK_CERTIFICATE_PATH: test
+
```
elwinschmitz, line 70: 😱 I some merge-conflict/rebase-magic in one of our futures... 🔮

.github/workflows/test_interface_portal.yml
aberonni: I figured that a failure in the "check translations" step is more likely than in the "save coverage report" step - so we'd like to make that happen sooner. And also the "save coverage report" step feels like an "end of the workflow" kind of thing (like the saving of artifacts). But I'm happy to move it if you think I should.

elwinschmitz: Yeah, I had the same idea about the "end-of-the-run"ness of the step...
But then I thought that "waiting till the end" wouldn't make sense if it was marked `always()` anyway..

But now I think I don't really mind either way. :D

interfaces/portal/karma.conf.js
```diff
@@ -0,0 +1,36 @@
+// Karma configuration file, see link for more information
+// https://karma-runner.github.io/1.0/config/configuration-file.html
```
elwinschmitz, line 2: Less wrong, more specific:
```suggestion
// https://karma-runner.github.io/6.4/config/configuration-file.html
```

Might be wrong, more realistic, less to change/update:
```suggestion
// https://karma-runner.github.io/latest/config/configuration-file.html
```

elwinschmitz (approved):

PR #7021, opened 2025-07-08
test: Replace arbitrary waitFor with actual message-counting

elwinschmitz:
[AB#36863](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/36863)

## Describe your changes

...

- [x] _Adding_ tests is irrelevant

## Portal preview-deployment

### DIFF
--- a/services/121-service/test/visa-card/block-visa-card.test.ts
+++ b/services/121-service/test/visa-card/block-visa-card.test.ts
@@ -7,7 +7,7 @@ import {
   programIdVisa,
   registrationVisa,
 } from '@121-service/src/seed-data/mock/visa-card.data';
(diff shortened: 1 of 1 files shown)


RubenGeo:
Would be nice in the future to also make a function that waits for a specific message test instead of just a number of messages. But for now this is better

RubenGeo (approved):

elwinschmitz:
@RubenGeo : 
> Would be nice in the future to also make a function that waits for a specific message text instead of just a number of messages. But for now this is better than what it was @elwinschmitz

That would also be nice;
But in practice its also nice that we now test for the 'extra' "there is a message for you..."-messages. (but you could of course also take that into account...)

PR #7038, opened 2025-07-11
feat: update default "export" for program registration attributes

aberonni:
[AB#37155](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37155) [AB#37160](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37160)

## Notes for release

This needs to be communicated to our stakeholders:
- export type `all-registrations` has become `registrations`
  - This means that the endpoint `/api/programs/${programId}/metrics/export-list/all-registrations` has become `/api/programs/${programId}/metrics/export-list/registrations`
- export type `included` does not exist anymore
  - This means that the endpoint `/api/programs/${programId}/metrics/export-list/included` does not exist anymore
- export type `payment` should be used in most instances where `included` was used before (because this was the only place where the export type `included` was being actually taken into account)
- When no `select` is provided to `export-list/registrations`, then a default hardcoded list of attributes will be returned, instead of the `export` property on program registration attributes being taken into account.
  - in other words, the `all-registrations`/`registrations` value in the `export` property of program registration attributes is not supported anymore

## Describe your changes

- Rename `ExportType.allRegistrations` to `ExportType.registrations`
- Remove the unused `ExportType.included`
- Change the default `export` property on `ProgramRegistrationAttributeEntity` to `[ExportType.payment]` (which seems to be the only remaining export type that uses this field...)

## Portal preview-deployment

https://happy-rock-0411d2003-7038.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/test/metrics/export-list.test.ts
+++ b/services/121-service/test/metrics/export-list.test.ts
@@ -4,6 +4,7 @@ import { RegistrationStatusEnum } from '@121-service/src/registration/enum/regis
 import { DebugScope } from '@121-service/src/scripts/enum/debug-scope.enum';
 import { SeedScript } from '@121-service/src/scripts/enum/seed-script.enum';
 import {
+  registrationScopedGoesPv,
   registrationScopedMiddelburgPv,
   registrationsPV,
 } from '@121-service/test/fixtures/scoped-registrations';
@@ -79,15 +80,6 @@ describe('Metric export list', () => {
     expect(data.map((r) => r.referenceId).sort()).toEqual(
       expectedReferenceIds.sort(),
     );
-
-    for (const registration of registrationsOCW.slice(1)) {
-      const exportRegistrationFound = data.find(
-        (r) => r.referenceId === registration.referenceId,
-      );
-      expect(exportRegistrationFound).toMatchSnapshot({
-        created: expect.any(String),
-      });
-    }
   });
 
   it('should return all filtered registrations from 1 program using a filter for included and a scoped user', async () => {
@@ -114,10 +106,10 @@ describe('Metric export list', () => {
     expect(data.length).toBe(1);
 
     const exportRegistration = data[0];
-
-    expect(exportRegistration).toMatchSnapshot({
-      created: expect.any(String),
-    });
+    expect(exportRegistration.referenceId).toBe(
+      registrationScopedGoesPv.referenceId,
+    );
+    expect(exportRegistration.status).toBe('new');
   });
 
   it('should return all filtered registrations from 1 program using a filter and search query', async () => {
@@ -144,10 +136,31 @@ describe('Metric export list', () => {
     expect(data.length).toBe(1);
 
     const exportRegistration = data[0];
+    expect(exportRegistration.referenceId).toBe(
+      registrationScopedGoesPv.referenceId,
+    );
+  });
 
-    expect(exportRegistration).toMatchSnapshot({
-      created: expect.any(String),
-    });
+  it('should export all registration attributes when no "select" is provided', async () => {
+    // Act
+    const getRegistrationsResponse = await getServer()
+      .get(`/programs/${OcwProgramId}/metrics/export-list/registrations`)
+      .set('Cookie', [accessToken])
+      .send();
+
+    // Assert
+    const data = getRegistrationsResponse.body.data;
+    expect(getRegistrationsResponse.status).toBe(HttpStatus.OK);
+    expect(data.length).toBe(5);
+
+    for (const registration of registrationsOCW.slice(1)) {
+      const exportRegistrationFound = data.find(
+        (r) => r.referenceId === registration.referenceId,
+      );
+      expect(exportRegistrationFound).toMatchSnapshot({
+        created: expect.any(String),
+      });
+    }
   });
 
   it('should support using "select" to retrieve a specific subset of columns', async () => {
--- a/services/121-service/test/metrics/__snapshots__/export-list.test.ts.snap
+++ b/services/121-service/test/metrics/__snapshots__/export-list.test.ts.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Metric export list should export all registrations of a single program regardless of status 1`] = `
+exports[`Metric export list should export all registration attributes when no "select" is provided 1`] = `
 {
   "addressCity": "AnotherCity",
   "addressHouseNumber": "2",
@@ -33,7 +33,7 @@ exports[`Metric export list should export all registrations of a single program 
 }
 `;
 
-exports[`Metric export list should export all registrations of a single program regardless of status 2`] = `
+exports[`Metric export list should export all registration attributes when no "select" is provided 2`] = `
 {
   "addressCity": "DifferentCity",
   "addressHouseNumber": "3",
@@ -66,7 +66,7 @@ exports[`Metric export list should export all registrations of a single program 
 }
 `;
 
-exports[`Metric export list should export all registrations of a single program regardless of status 3`] = `
+exports[`Metric export list should export all registration attributes when no "select" is provided 3`] = `
 {
   "addressCity": "AnotherCity",
   "addressHouseNumber": "4",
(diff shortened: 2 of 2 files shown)


services/121-service/src/metrics/metrics.controller.ts
```diff
@@ -50,7 +50,7 @@ export class MetricsController {
     permissions: [PermissionEnum.RegistrationPersonalEXPORT],
   })
   @ApiOperation({
-    summary: `[SCOPED] Retrieve data for export. Filters only work for export type ${ExportType.allRegistrations}. If no "select" is provided, then the default export columns will be returned.`,
+    summary: `[SCOPED] Retrieve data for export. Filters only work for export type ${ExportType.registrations}. If no "select" is provided, then the hardcoded default select for exporting registrations will be returned.`,
```
aberonni, line 53: This is the new behaviour according to my research. @RubenGeo it will use the function `getDefaultSelectForExportRegistrations` whenever a select is not provided. This seems to be intended behaviour but I'm not sure.

services/121-service/test/metrics/export-list.test.ts
```diff
@@ -78,10 +61,10 @@ describe('Metric export list', () => {
     });
   });
 
-  it('should export all people affected of a single program regardless of status', async () => {
```
RubenGeo, line 81: I think it does not make sense to use snapShots for all the tests as each test is testing more than it should and it's harder to read the test code. Another advantage of not using a snapshot is that you can be more granular in you assertion for example using 

```
id: expect.any(Number)
created: expect.any(Date)
status: expect.any(String)
```

I think a setup of the tests could be:



- should return all filtered registrations from 1 program using a filter and search query
  - Only assert the referenceIds here. Because this test should assert only whether the filter works and not which columns are returned
- should return all filtered registrations from 1 program using a filter for included and a scoped user
  - Only assert the referenceIds here. Because this test should assert only whether the filter/scope works and not which columns are returned
- should export all registrations of a single program regardless of status
  - Assert the referenceIds here. And maybe also the statusses here to sort of document that this endpoint explicitely also returns registration with status deleted.
- should support using "select" to retrieve a specific subset of columns 
  - I would check these using a toStrictEqual because it's more readable than a seperate snap shot file
```
.toStrictEqual({
      referenceId: registrationScopedMiddelburgPv.referenceId,
      fullName: registrationScopedMiddelburgPv.fullName,
      phoneNumber: registrationScopedMiddelburgPv.phoneNumber,
    });
```
- should retrieve all columns when not sending a select **(new test)**
  - Here I think it can makes sense to use a snapshot, however this has a drawback that than the created date will break the test. However using something like toStrictEqual and building the object to validate using the also has advantages because than you can do something like this:
```
      id: expect.any(Number),
      created: expect.any(Date),
      status: expect.any(String),
```
    - Which is nice because than the test is less likely to break in the future. However it may not be worth the effort because there are many columns. You could build to the object to validate using PaginateConfigRegistrationView the programRegistrationAttributes of the PV program and the original registration
  - 
-  should export in excel format

- should export failed when user doesn't have enough permission


What are your thoughts on this?

RubenGeo (changes requested):

services/121-service/test/metrics/export-list.test.ts
aberonni: Compared to what was there before, IMHO a snapshot does the same thing but better. I find it more significant to look at a diff in a snapshot file to see what is changing in the I/O of a function, than to read test file, and it gives me a very quick read as a developer to see what has changed by simply updating the snapshots.

The specifics that you list seem to be an improvement compared to the tests that were already there. On those specifics:

> Only assert the referenceIds here. Because this test should assert only whether the filter works and not which columns are returned

Agreed, will do.

> I would check these using a toStrictEqual because it's more readable than a seperate snap shot file

I think this is a matter of preference. Personally, I find the snapshot more readable.

>  the created date will break the test. 

This has been fixed since you first looked at it.

--

So essentially, I'm happy to update the tests that only need to check the referenceIds to only check referenceIds. On the other things, I need to be convinced to change them because I don't think they're improvements. And if you're asking for tests to be added, I think I need more details on what should be there vs what is already there.

aberonni: I updated the tests according to my last comment.

RubenGeo: > I find it more significant to look at a diff in a snapshot file to see what is changing in the I/O of a function, than to read test file

This is a good point for a dif view, which I did not consider. 

And I also have to say that I did not know you could do this:
```
      expect(exportRegistrationFound).toMatchSnapshot({
        created: expect.any(String),
      });
```
which is really nice.

I gave it some thought and I think I think we should be careful to not over use snapshots for the following reasons:

1. A (non) snapshot tests forces us as developers to more explicitly think about what we want to assert when writing a test and what we are updating when a test breaks. And I think it's good to not assert more to make it easier for tests to each have a seperate concern
2. In a snapshot you are storing duplicate (non typed) data. For example this commit would not have been needed if snapshot testing was not used: [`86f178d` (#7038)](https://github.com/global-121/121-platform/pull/7038/commits/86f178d694ccde1b96https://github.com/global-121/121-platform/pull/7038/commits/86f178d694ccde1b96c8a5d62c441bc76daafccdc8a5d62c441bc76daafccd). I feel that you can more often catch errors due to typing instead of having to wait on CI to tell you that a (less) relevant part of snapshot is outdated
3. And this is more personal but: I think they are harder to read. Switching between two files without being able to cmd+click to find which snapshot relates to which tests (especially when the snapshots contain a lot of unneeded data) is less convienent. It's also harder to understand the logic from reading a snapshot. E.g. You would understand from reading this logic that this code expects the registration to come back from the endpoint to be the same as the one you seeded in the beforeAll instead of having sort back track what the values are you see in a snapshot:
```
.toStrictEqual({
      referenceId: registrationScopedMiddelburgPv.referenceId,
      fullName: registrationScopedMiddelburgPv.fullName,
      phoneNumber: registrationScopedMiddelburgPv.phoneNumber,
    });
```

For this specific case for the 'select' related tests I still think we are storing too much irrelevant info in the snapshots. Mainly that the check happens for all 4 registrations. It would actually be nice if we would assert just the colums/keys for one registration. We could either do this by moving the select to a separate file and only seed one registartion there. Or we could keep them in the current file and use a filter by referenceId in the API call (but than we of course sort of mix the filter with the select functionality). 

And I personally would use a to toStrictEqual for the test `should support using "select" to retrieve a specific subset of columns` as it is easy to write and easier to read in my opinion. For the `should export all registration attributes when no "select" is provided` I guess we can take the shortcut of the snapshot as dynamically building the object is harder there. But maybe we can add more of this cool logic to this like you did here  
```
      expect(exportRegistrationFound).toMatchSnapshot({
        created: expect.any(String),
      });
```

Also for id and programFspConfigurationId for example

services/121-service/test/metrics/export-list.test.ts
```diff
@@ -144,10 +136,31 @@ describe('Metric export list', () => {
...
+      .get(`/programs/${OcwProgramId}/metrics/export-list/registrations`)
+      .set('Cookie', [accessToken])
+      .send();
+
+    // Assert
+    const data = getRegistrationsResponse.body.data;
+    expect(getRegistrationsResponse.status).toBe(HttpStatus.OK);
+    expect(data.length).toBe(5);
+
+    for (const registration of registrationsOCW.slice(1)) {
```
RubenGeo, line 156: Is saw that this was also in the previous version of the tests so maybe out of scope for this PR. But do you know why this slice is needed?

RubenGeo (approved):
I wrote a long comment about my vendetta against snapshot testing. If you think it's nonsense I don't want to hold this PR hostage so you can resolve the convo and merge this PR

services/121-service/test/metrics/export-list.test.ts
aberonni: Resolving as per your other comment, but happy to discuss further elsewhere if it helps (I feel this conversation keeps coming back)

services/121-service/test/metrics/export-list.test.ts
aberonni: Nope - I just carried it over from the previous test. I didn't want to make the scope of this any bigger given that it seems to be urgent.

PR #7050, opened 2025-07-15
ci: Make the path-filter external-action only run on PRs, to prevent confusing errors

elwinschmitz:
[AB#37303](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37303)

## Portal preview-deployment

### DIFF
--- a/.github/workflows/test_e2e_portal.yml
+++ b/.github/workflows/test_e2e_portal.yml
@@ -139,8 +139,13 @@ jobs:
     needs: [test-shard]
     if: ${{ always() }}
     steps:
-      - name: Required check failed
+      - name: Check required job results
         if: ${{ contains(needs.*.result, 'failure') }}
-        run: exit 1
-      - name: Dummy step
-        run: echo "All required checks have successfully passed."
+        run: |
+          echo "Some checks have failed." >> $GITHUB_STEP_SUMMARY
+          exit 1
+      - name: Check required job results
+        if: ${{ contains(needs.*.result, 'failure') != 'true' || contains(needs.*.result, 'cancelled') != 'true' || contains(needs.*.result, 'skipped') != 'true' }}
+        run: |
+          echo "All required checks have successfully passed." >> $GITHUB_STEP_SUMMARY
+          exit 0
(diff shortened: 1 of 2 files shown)


jannisvisser:
@elwinschmitz is this intentionally still on draft?

.github/workflows/test_e2e_portal.yml
```diff
@@ -136,8 +140,13 @@ jobs:
...
+      - name: Check required job results
         if: ${{ contains(needs.*.result, 'failure') }}
-        run: exit 1
-      - name: Dummy step
-        run: echo "All required checks have successfully passed."
+        run: |
+          echo "Some checks have failed." >> $GITHUB_STEP_SUMMARY
+          exit 1
+      - name: Check required job results
+        if: ${{ contains(needs.*.result, 'failure') != 'true' || contains(needs.*.result, 'cancelled') != 'true' || contains(needs.*.result, 'skipped') != 'true' }}
```
aberonni, line 149:
```suggestion
        if: ${{ contains(needs.*.result, 'failure') != 'true' || contains(needs.*.result, 'cancelled') != 'true' }}
```

I think `skipped` happens when the path-filter fails, no?

.github/workflows/test_service_api.yml
```diff
@@ -171,8 +171,13 @@ jobs:
...
+      - name: Check required job results
         if: ${{ contains(needs.*.result, 'failure') }}
-        run: exit 1
-      - name: Dummy step
-        run: echo "All required checks have successfully passed."
+        run: |
+          echo "Some checks have failed." >> $GITHUB_STEP_SUMMARY
+          exit 1
+      - name: Check required job results
+        if: ${{ contains(needs.*.result, 'failure') != 'true' || contains(needs.*.result, 'cancelled') != 'true' || contains(needs.*.result, 'skipped') != 'true' }}
```
aberonni, line 180:
```suggestion
        if: ${{ contains(needs.*.result, 'failure') != 'true' || contains(needs.*.result, 'cancelled') != 'true' }}
```

aberonni (changes requested):
Same as above

aberonni:
@elwinschmitz does the comment on devops mean that this doesn't actually work as intended? If so, I would put it back on draft. However, if cancelling a job is the only edge case, I'm not sure we need to treat it as urgent - I would simply advise against cancelling jobs.

PR #7060, opened 2025-07-18
Refactor E2E tests

Piotrk39:
[AB#37375](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37375) 

## Describe your changes

Refactors E2E tests to use one reset instance as much as possible and removes duplicated code.

## Portal preview-deployment

https://happy-rock-0411d2003-7060.westeurope.3.azurestaticapps.net


e2e/portal/tests/ChangeRegistrationStatus/DeleteRegistrations.spec.ts
aberonni, line 1: @Piotrk39 I guess we already kind discussed this a bit via chat last week, but this kind of refactor doesn't make sense to me. The minor speed gain we might be getting here ("might", because with sharding this might even make things slower) doesn't seem to justify cramming everything into one file.

I could understand this if you were to look at a way to reduce repetition. All of these tests seem to do _almost_ the same thing. Then each test could just be a function call with some parameters? That would make sense to me. But this version of this refactor isn't something that I would like to endorse.

Again, I think it's worth talking about this with all the relevant stakeholders, or being clearer about the goal here. I'd suggest doing that also before beginning on other refactors, to avoid wasted effort.

Piotrk39:
Closing as it is not relevant for time being

PR #7131, opened 2025-08-21
test: indirectRelationConfig exhaustiveness

jannisvisser:
[AB#37677](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37677)

## Describe your changes

- Adds a test that checks for exhaustiveness of the indirectRelationConfig object
  - which should include all entities with an indirect relation to Registration entity, 
  - except when the path includes a one-to-many relationship (direction: one entity X to many registrations). This means that e.g. transactions (many-to-one with registration) are kept and therefore scoped, while e.g. 'payments' (not yet in main, but one-to-many with transactions) is filtered out and therefore not scoped.
- The initial local run of the test identified 3 missing entities, which are added to the object. 
- See [AB#37677](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37677) for more details
- FYI: I needed a small tweak to make the knip test work. Not sure if this is ideal..

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/app.module.ts
+++ b/services/121-service/src/app.module.ts
@@ -1,6 +1,5 @@
 import { BullModule } from '@nestjs/bull';
 import { Module, OnApplicationBootstrap } from '@nestjs/common';
-import { Type } from '@nestjs/common';
 import { APP_GUARD, APP_INTERCEPTOR } from '@nestjs/core';
 import { MulterModule } from '@nestjs/platform-express';
 import { ScheduleModule } from '@nestjs/schedule';
@@ -38,17 +37,7 @@ import { ProgramExistenceInterceptor } from '@121-service/src/shared/interceptor
 import { TransactionJobsModule } from '@121-service/src/transaction-jobs/transaction-jobs.module';
 import { TransactionQueuesModule } from '@121-service/src/transaction-queues/transaction-queues.module';
 import { TypeOrmModule } from '@121-service/src/typeorm.module';
-import { DevelopmentController } from '@121-service/src/utils/test-helpers/test.controller';
-
-const getControllers = (): Type<unknown>[] => {
-  const baseControllers: Type<unknown>[] = [AppController];
-
-  if (env.NODE_ENV === 'development') {
-    baseControllers.push(DevelopmentController);
-  }
-
-  return baseControllers;
-};
+import { TestController } from '@121-service/src/utils/test-helpers/test.controller';
 
 @Module({
   // Note: no need to import just any (new) Module in ApplicationModule, when another Module already imports it
@@ -103,7 +92,7 @@ const getControllers = (): Type<unknown>[] => {
     SafaricomReconciliationModule,
     OnafriqReconciliationModule,
   ],
-  controllers: getControllers(),
+  controllers: [AppController, TestController],
   providers: [
     {
       provide: APP_GUARD,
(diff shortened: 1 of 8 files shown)


RubenGeo:
@jannisvisser It does not matter for this PR but the branch name is supposed to start with your git name and I think most people uses kebab instead of cammelcase for the descriptive name part of the branch

services/121-service/src/scoped.repository.spec.ts
```diff
@@ -19,4 +25,91 @@ describe('ScopedRepository', () => {
     // An actual relevant test to still add is to check if .where is correctly prohibited
     expect(scopedRepository).toBeDefined();
   });
+
+  describe('indirectRelationConfig exhaustiveness', () => {
```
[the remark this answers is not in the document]
jannisvisser: added in new version

services/121-service/src/app.module.ts
```diff
@@ -37,6 +38,17 @@ import { ProgramExistenceInterceptor } from '@121-service/src/shared/interceptor
...
+
+const getControllers = (): Type<unknown>[] => {
+  const baseControllers: Type<unknown>[] = [AppController];
+
+  if (env.NODE_ENV === 'development') {
+    baseControllers.push(DevelopmentController);
+  }
+
+  return baseControllers;
+};
```
RubenGeo, line 51:
```
  @ApiOperation({
    summary:
      'WARNING: Kills 121-service. Only works in DEBUG-mode. Only used for testing purposes.',
  })
  @ApiExcludeEndpoint(!IS_DEVELOPMENT)
  @Post('kill-service')
  killService(@Body() body: SecretDto, @Res() res): void {
    if (body.secret !== env.RESET_SECRET) {
      return res.status(HttpStatus.FORBIDDEN).send('Not allowed');
    }
    if (!IS_DEVELOPMENT) {
      return;
    }

    console.log('Service is being killed...');
    // eslint-disable-next-line n/no-process-exit -- Exiting the app is the literal purpose of this method/endpoint
    process.exit(1);
  }

```

For the kill the service controller we picked a different approach `@ApiExcludeEndpoint(!IS_DEVELOPMENT)`. I think this is more elegant. 


Also I think we should either move the Kill service controller to the test controller file or move this validate-indirect-relations to the script controller file/ folder.

services/121-service/test/scope/indirect-relation-config-exhaustiveness.test.ts
```diff
@@ -0,0 +1,23 @@
...
+import { resetDB } from '@121-service/test/helpers/utility.helper';
+
+describe('Development Controller - Indirect Relations Validation', () => {
+  beforeEach(async () => {
+    await resetDB(SeedScript.nlrcMultiple, __filename);
+  });
+
+  // This test checks if indirectRelationConfig (in scoped.repository.ts) is exhaustive, as it is hard to maintain this manually over time.
+  // It is done via API-test instead of unit-test, because it was not possible to it requires the database to be set up.
+  // If this test fails, then you can find more details on missing details in the 121-service logs, which can be used to update the indirectRelationConfig.
```
RubenGeo, line 14: Now that I read this it would be more convenient if the endpoint return the error message in the body. 

Than you can compare it here in this test file with an empty string so you can conveniently read it from CI/CD if it goes wrong.

services/121-service/src/app.module.ts
jannisvisser: Agree. I decided to move the kill-service endpoint to the test.controller, as the other way around made no sense at all. I also aligned naming more (the apitags/classname/etc is all 'test' now, instead of partly test, partly 'development'.

services/121-service/test/scope/indirect-relation-config-exhaustiveness.test.ts
jannisvisser: done

PR #7144, opened 2025-08-26
Dynamic callback url safaricom

RubenGeo:
AB#37746 

## Portal preview-deployment

### DIFF
--- a/services/mock-service/src/fsp-integration/safaricom/safaricom.mock.service.ts
+++ b/services/mock-service/src/fsp-integration/safaricom/safaricom.mock.service.ts
@@ -3,7 +3,11 @@ import { Injectable } from '@nestjs/common';
 import { setTimeout } from 'node:timers/promises';
 import { lastValueFrom } from 'rxjs';
 
-import { API_PATHS, EXTERNAL_API_ROOT } from '@mock-service/src/config';
+import {
+  API_PATHS,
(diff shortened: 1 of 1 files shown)


elwinschmitz (changes requested):
I think it _DOES_ make sense to do _some validation_...

Doesn't need to be very fancy; But something like:

```ts
function validateUrlIsSafe(rawUrl: string) {
  try {
    const url = new URL(rawUrl);
    return url.hostname === "121.global";
  } catch (e) {
    return false;
  }
}
```

And then:
```
let url = IS_DEVELOPMENT
      ? `${EXTERNAL_API_ROOT}/${API_PATHS.safaricomTransferCallback}`
      : validateUrlIsSafe(transferDto.ResultURL) ? transferDto.ResultURL : '';
```

Or something similar.

elwinschmitz:
> ```ts
> function validateUrlIsSafe(rawUrl: string) {
> }
> ```

Hmm.. Wait. We shouldn't be hard-coding the 121.global hostname there..

We should use something like:

```ts
function validateUrlIsSafe(rawUrl: string) {
  try {
    const url = new URL(rawUrl);
    const safeUrl = new URL(env.EXTERNAL_121_SERVICE_URL);
    return url.hostname === safeUrl.hostname;
  } catch (e) {
    return false;
  }
}

```

PR #7186, opened 2025-09-02
feat: logout on deprecated permissions

jannisvisser:
[AB#37836](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37836)

## Describe your changes

- This change came out of the program-to-project rename work, where permissions (e.g. 'program:metrics.read) were also renamed. For logged in users after deploy, this would lead to portal where certain things are not visible (Monitoring page menu-item) or not work.
- After discussion [here](https://teams.microsoft.com/l/message/19:ce634fd37f34477d9f98391f70e1f0ef@thread.skype/1756730859255?tenantId=d3ab9790-6ae2-4bd8-aa5e-02864483e7c7&groupId=55f2fa49-7d9b-4c63-b616-b986f3136c3a&parentMessageId=1756730859255&teamName=510%20-%20CVA&channelName=121%20Development&createdTime=1756730859255) I in the end arrived at the 'force manual re-login' option. 
- I looked into the option of directly calling GET /user/current as well, but I think it's more complicated, a.o. because there is a risk of an infinite loop involved. I am happy to create a follow-up somehow (item/spike/SA parking lot) to improve on this iteratively in the future.
- For now, Letting the user do one manual re-login is fine I think, given the scope of this refactor.

## Portal preview-deployment

https://happy-rock-0411d2003-7186.westeurope.3.azurestaticapps.net

### DIFF
--- a/interfaces/portal/src/app/services/auth.service.spec.ts
+++ b/interfaces/portal/src/app/services/auth.service.spec.ts
@@ -1,19 +1,24 @@
+import { Injector } from '@angular/core';
 import { TestBed } from '@angular/core/testing';
 import { Router } from '@angular/router';
-import { Injector } from '@angular/core';
 
 import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
 
-import { AuthService } from './auth.service';
+import { AuthService } from '~/services/auth.service';
 import { LogService } from '~/services/log.service';
 import { LocalStorageUser } from '~/utils/local-storage';
+
+interface MockAuthStrategy {
+  isUserExpired: jasmine.Spy<() => boolean>;
+  logout: jasmine.Spy<() => Promise<void>>;
+}
 
 describe('AuthService - hasDeprecatedPermissions', () => {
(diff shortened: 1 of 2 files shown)


nielsbom:
Code is good and works, but UI-wise it's probably not working as we want.

jannisvisser:
- @nielsbom this front-end change addresses the issue you raised. The problem was - as identified together with @aberonni - that even though the check on deprecated permissions in the user getter was evaluated correctly when it should, the subsequent isLoggedIn check was not (only on major route changes). 
- As for potentially directly refreshing the user object instead of forcing logout (and manual re-login), we agreed current behaviour is fine as it aligns for example with isUserExpired flow. This can always be changed/improved later.
- I ignored one lint error in my test because I could not get it the test to work otherwise and spent too much time on it already. Happy to look at it together with someone, but also fine like this for me.

nielsbom (approved):

PR #7223, opened 2025-09-08
chore: Remove unused Event-tracking via ApplicationInsights

elwinschmitz:
See [AB#37920](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/37920)

## Describe your changes

As we won't be using ApplicationInsights for "User-intent/behavior event-tracking", we can get rid of some of the code especially for it.

- [X] Adding tests is unnecessary/irrelevant

## Portal preview-deployment

https://happy-rock-0411d2003-7223.westeurope.3.azurestaticapps.net

### DIFF
--- a/interfaces/portal/src/app/services/log.service.ts
+++ b/interfaces/portal/src/app/services/log.service.ts
@@ -1,13 +1,7 @@
 import { Injectable } from '@angular/core';
 
-import { DebugPlugin } from '@microsoft/applicationinsights-debugplugin-js';
 import { ApplicationInsights } from '@microsoft/applicationinsights-web';
 import { environment } from 'src/environments/environment';
-
-export enum LogEvent {
-  userLogin = 'user/login',
-  userLogout = 'user/logout',
-}
 
 @Injectable({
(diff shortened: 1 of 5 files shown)


interfaces/portal/src/app/services/log.service.ts
```diff
@@ -76,15 +47,8 @@ export class LogService {
...
-  public logPageView(name?: string): void {
-    if (this.appInsightsInitialized) {
-      this.appInsights.trackPageView({ name });
-    }
-    console.info(`LOG: PageView: "${name ?? ''}"`);
-  }
-
   public logEvent(
-    name: LogEvent,
+    name: string,
```
jannisvisser, line 51: what is the point of keeping this method (and the other unused methods here) if they are not used?

jannisvisser (commented):
I'm sure this is fine, but can you maybe point me to (or elaborate on) the decision that is the premise of this PR? (that we don't do event-tracking via application insights).

PR #7254, opened 2025-09-16
Add batch of fast K6 tests

Piotrk39:
AB#38067 

## Describe your changes

Adds __ENV parameter to speed up the K6 execution speed so they can run every PR.

- [ ] Adding tests is unnecessary/irrelevant
- [ ] The changes do not touch the UI/UX

## Portal preview-deployment


k6/models/payments.js
```diff
@@ -4,7 +4,22 @@ import http from 'k6/http';
 import config from './config.js';
 const { baseUrl } = config;
 
-export default class paymentsModel {
+export default class PaymentsModel {
+  verifyPaymentDryRun(programId, amount) {
```
RubenGeo, line 8: I am not sure if I understand it correctly. But should you not also use verifyPaymentDryRunUntilSuccess here?

RubenGeo (changes requested):
Great changes!  Hope this also gets rid of the flakeyness. Just some small comments on the code.

For the rest I think it would also be nice if you add a small section to the readme of how to run these tests locally and than set your own duplicate number. :)

k6/tests/findDuplicates100kRegistrations.js
```diff
@@ -27,7 +42,19 @@ function checkAndFail(response, checks) {
...
+}
+
 export default function () {
+  // Wait for the service to restart and become available
+  let serviceUp = false;
+  while (!serviceUp) {
+    sleep(1); // wait for 1 second before retrying
+    serviceUp = isServiceUp();
+  }
+  // reset DB
```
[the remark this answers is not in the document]
Piotrk39: Yes without this after the finish of previous test and the reset of the other some APIs are still unavailable and at this step test needs to implicitly wait for the service to be up. Without it it fails every time. Tests in the fast run are executed a bit differently than in cronjob.

k6/README.md
```diff
@@ -1,67 +1,111 @@
...
 ```
 
-Make sure ENV-variable `EXTERNAL_121_SERVICE_URL` is set to `http://localhost:3000` (and not to some `ngrok` address)
+**To:**
 
-**To run the tests:**
+```javascript
+baseUrl: __ENV.EXTERNAL_121_SERVICE_URL || 'http://localhost:3000/',
+```
+
```
RubenGeo, line 62: Instead of this you could change config.js to: 

```
/* global __ENV */
export default {
  baseUrl: __ENV.EXTERNAL_121_SERVICE_URL
    ? __ENV.EXTERNAL_121_SERVICE_URL.replace(/\/+$/, '') + '/' // This ensures there's exactly one trailing slash
    : 'http://localhost:3000/',
  credentials: {
    username: __ENV.USERCONFIG_121_SERVICE_EMAIL_ADMIN,
    password: __ENV.USERCONFIG_121_SERVICE_PASSWORD_ADMIN,
  },
};
```

So you always trim the / from the url and you should never end up with 2 slashes

k6/README.md
```diff
@@ -1,67 +1,111 @@
...
 
-Then:
+#### Adjusting Test Load
 
-```shell
-npm run test -- tests/<name of the test>.js
+For faster CI execution, tests use:
+
+```javascript
+const duplicateNumber = parseInt(__ENV.DUPLICATE_NUMBER || '5');
```
RubenGeo, line 75:
```suggestion
For faster CI execution, tests use:

```shell
DUPLICATE_NUMBER=5 npm run test -- tests/<name-of-the-test>.js
```

One of the nice things of using ENV variables is that you do not need to make code changes to run code with a different configuration. 

So this way you use the env variable locally the same way as you would use it in the ci pipeline

RubenGeo (changes requested):
See 2 comments :)

k6/README.md
Piotrk39: I changed it as per suggestion

PR #7261, opened 2025-09-18
fix: Make checks around (error-)status/response-codes easier to read, less error-prone = less (unnecessary) noise in the logs

elwinschmitz:
[AB#38126](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38126)

## Describe your changes

Add a utility to use when dealing with response/status-codes coming from (insufficiently typed) third-parties.

So the logic is easier to read AND it prevents 'false negatives';
In which case a "don't throw an error for this code" was incorrectly handled and still causing a lot of (unecessary) log-lines.

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/utils/comparison.helper.ts
+++ b/services/121-service/src/utils/comparison.helper.ts
@@ -1,8 +1,8 @@
 /**
  * Compares if two (status-) codes are _equivalent_, not taking types into account.
- * @example isSameStatus('200', 200) returns true
+ * @example isSameAsString('200', 200) returns true
  */
-export function isSameStatus(
+export function isSameAsString(
   a: number | string | undefined,
   b: number | string | undefined,
 ): boolean {
--- a/services/121-service/src/utils/comparison.helper.ts
+++ b/services/121-service/src/utils/comparison.helper.ts
@@ -1,0 +1,10 @@
+/**
+ * Compares if two (status-) codes are _equivalent_, not taking types into account.
+ * @example isSameStatus('200', 200) returns true
+ */
+export function isSameStatus(
+  a: number | string | undefined,
+  b: number | string | undefined,
+): boolean {
+  return String(a) === String(b);
+}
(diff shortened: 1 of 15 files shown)


services/121-service/src/utils/comparison.helper.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Compares if two (status-) codes are _equivalent_, not taking types into account.
+ * @example isSameStatus('200', 200) returns true
+ */
+export function isSameStatus(
```
jannisvisser, line 5: I don't find this name very intuitive. The location and filename of this helper imply generic use. So does the actual functionality, which is not specifically about 'status' at all. So a better name would be something more generic than 'status'. More something like 'isStringifiedEqualTo'. (I'm not actually implying this as a better suggestion, but at least it conceptually describes better what it does.) 

I'm also wondering, is there not some typescript-code that already does this same thing somehow?

jannisvisser (changes requested):

services/121-service/src/utils/comparison.helper.ts
elwinschmitz: Hmm.. now that I think about it...
There is maybe another approach + fix.

These enum-error-codes should actually almost never be `number`s... (as you wouldn't do calculations with them)..
Then it would make sense to do the `String(error.code) === enum`-checks everywhere in code...

But the "you just have to know which API-response returns a number of a string"-issue would remain.
With a generic helper, you would never have to think about "which situation  is this?", because the helper takes care of all possible scenario's...

🤔

jannisvisser: I think the generic method/code is fine, but still think a generic name is better than. The more specific name is not going to help you remember to use it in some file where you haven't used this method yet. And there might be more use cases for using this, which would also be fine.

PR #7287, opened 2025-09-25
waitForMessagesToComplete refactor

Yuri-510:
[AB#30120](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30120) 

## Portal preview-deployment

### DIFF
--- a/services/121-service/test/helpers/program.helper.ts
+++ b/services/121-service/test/helpers/program.helper.ts
@@ -428,7 +428,7 @@ export async function waitForMessagesToComplete({
   minimumNumberOfMessagesPerReferenceId?: number;
   expectedMessageAttribute?: {
     key: keyof MessageActivity['attributes'];
-    values: string[];
+    values: MessageActivity['attributes'][keyof MessageActivity['attributes']][];
   };
 }): Promise<void> {
   const maxWaitTimeMs = 25_000;
@@ -490,10 +490,13 @@ function filterByExpectedAttribute({
   messageHistories,
   expectedMessageAttribute,
 }: {
-  messageHistories: { referenceId: string; messageHistory: any[] }[];
+  messageHistories: {
+    referenceId: string;
+    messageHistory: MessageActivity[];
+  }[];
   expectedMessageAttribute: {
     key: keyof MessageActivity['attributes'];
-    values: string[];
+    values: MessageActivity['attributes'][keyof MessageActivity['attributes']][];
   };
 }): string[] {
   return messageHistories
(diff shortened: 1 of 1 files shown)


services/121-service/test/helpers/program.helper.ts
```diff
@@ -414,16 +414,25 @@ export async function waitForStatusUpdateToComplete(
...
+  expectedValues,
+  expectedValueType,
 }: {
   programId: number;
   referenceIds: string[];
   accessToken: string;
   minimumNumberOfMessagesPerReferenceId?: number;
+  expectedValues?: string[];
+  expectedValueType?: messageValueTypeEnum;
 }): Promise<void> {
```
RubenGeo, line 436: I think an interface like this could be clearer for the reader:

```
export async function waitForMessagesToComplete({
  programId,
  referenceIds,
  accessToken,
  expectedMinimumNumberOfMessagesPerReferenceId,
  expectedBody,
  expectedContentType,
}: {
  programId: number;
  referenceIds: string[];
  accessToken: string;
  expectedMinimumNumberOfMessagesPerReferenceId?: number;
  expectedBody?: string[];
  expectedContentType?: string[];
}): Promise<void> {
``` 

and than in the function you could have a pattern like (disclaimer this code was generated by co pilot:

```

if (expectedBody) {
  referenceIdsWaitingForMessages = filterByExpectedBody({
    messageHistories,
		expectedBody,
  });
} else if (expectedContentType) {
  referenceIdsWaitingForMessages = filterMessagesByExpectedContentType({
 	 expectedContentType,
    expectedValueType: 'contentType',
  });
} else {
  referenceIdsWaitingForMessages = filterMessagesByMinimumNr({
    messageHistories,
    minimumNumberOfMessages: expectedMinimumNumberOfMessagesPerReferenceId ?? 1,
  });
}
```

This way callers of the function don’t need to understand the expectedValueType enum or how it maps to expectedValues. What do you think?

RubenGeo: You could also add something like this at the start. To ensure we never use it in a way for which is was not intended

```
const expectations = [
  expectedMinimumNumberOfMessagesPerReferenceId,
  expectedBody,
  expectedContentType,
].filter((v) => v !== undefined);

if (expectations.length !== 1) {
  throw new Error('Exactly one of expectedMinimumNumberOfMessagesPerReferenceId, expectedBody, or expectedContentType must be provided.');
}

```

RubenGeo (approved):
Great that we are implementing this improvement! I only have a comment on the readability

RubenGeo (changes requested):
Great that we are implementing this improvement! I only have a comment on the readability. 

I accidentally pressed approve before,, but I meant to click request changes

services/121-service/test/helpers/program.helper.ts
Yuri-510: What I've written now makes the use of the helper more dynamic than checking for body content and/or content type. With my changes you can say which type of message attributes you'd like to check. So if you want to check for body content, you pass 'body' as value type and the body's expected value as parameters. But if you want to check for content type, or say status, that is possible too.

RubenGeo (approved):
Nice :)

RubenGeo (approved):

PR #7302, opened 2025-09-29
send message when status completed

Yuri-510:
[AB#30120](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/30120) 

## Portal preview-deployment

https://happy-rock-0411d2003-7302.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/test/payment/payment-count-completed.test.ts
+++ b/services/121-service/test/payment/payment-count-completed.test.ts
@@ -61,12 +61,13 @@ describe('Do a payment to a PA with maxPayments=1', () => {
         status: RegistrationStatusEnum.included,
         accessToken,
       });
+      const paymentReferenceIds = [registrationAh.referenceId];
 
       // Act
       const doPaymentResponse = await doPayment({
         programId,
         amount,
-        referenceIds: [registrationAh.referenceId],
+        referenceIds: paymentReferenceIds,
         accessToken,
       });
       const paymentId = doPaymentResponse.body.id;
@@ -74,7 +75,7 @@ describe('Do a payment to a PA with maxPayments=1', () => {
       // Assert
       await waitForPaymentTransactionsToComplete({
         programId,
-        paymentReferenceIds: [registrationAh.referenceId],
+        paymentReferenceIds,
         accessToken,
         maxWaitTimeMs: 10_000,
       });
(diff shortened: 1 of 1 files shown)


services/121-service/test/payment/payment-count-completed.test.ts
```diff
@@ -124,5 +129,64 @@ describe('Do a payment to a PA with maxPayments=1', () => {
       );
       expect(statusChangeToCompleted.length).toBe(1);
     });
+
+    it('should send a template message when status is complete', async () => {
+      // Arrange
```
RubenGeo, line 134: I think we could save a bit of processing time if you modify the test at the top of this file to also assert whether the templated messages are send, because I think for the rest the setup of these 2 tests are the same. 

The integration tests take a lot longer to run than the unit tests so there we have to be bit more careful if an extra test is warranted

RubenGeo (changes requested):
Looking good, some minor comments

services/121-service/src/transaction-jobs/services/transaction-jobs-helper.service.ts
```diff
@@ -82,26 +83,32 @@ export class TransactionJobsHelperService {
...
-        await this.setStatusToCompleteIfApplicable({
-          registration,
-          programId,
-          paymentCount,
-        });
+      const maxPaymentsExceeded = await this.isMaxPaymentsExceeded({
+        registration,
+        programId,
+        paymentCount,
+      });
```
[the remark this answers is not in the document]
RubenGeo: I do think the function checks 2 things

1) Is max payment enabled for this program
2) does the registration current payments exceed the max amount of payments for this registrations 

And it only returns true if both conditions apply.

services/121-service/test/payment/payment-count-completed.test.ts
Yuri-510: I'm afraid we'll run into problems at a later stage when using `beforeAll()`. For now I merged the tests, as you proposed in your original comment.

RubenGeo (dismissed):

RubenGeo (approved):

PR #7365, opened 2025-10-08
Check deleted result in pending message for race conditions

RubenGeo:
AB#38402 
AB#38422
## Describe your changes

Prevents sending double messages by checking delete result

- [ ] I have updated the [121 Service Module Dependency Diagram in the wiki](https://github.com/global-121/121-platform/wiki/121-Service-Module-Diagram) when the 121 module dependencies have changed.

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/notifications/services/message.service.ts
+++ b/services/121-service/src/notifications/services/message.service.ts
@@ -167,7 +167,14 @@ export class MessageService {
     if (!existingPendingMessage) {
       return;
     }
-    await this.whatsappPendingMessageRepo.delete(pendingMessageId);
+
+    // Try to delete the pending message, if it was already deleted by another worker we do not need to process it
+    // This should solve any possible as record in a db cannot be delete twice
+    const deleteResult =
+      await this.whatsappPendingMessageRepo.delete(pendingMessageId);
+    if (deleteResult?.affected !== 1) {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


services/121-service/src/notifications/services/message.service.ts
```diff
@@ -167,7 +167,14 @@ export class MessageService {
     if (!existingPendingMessage) {
       return;
     }
-    await this.whatsappPendingMessageRepo.delete(pendingMessageId);
+
+    // Try to delete the pending message, if it was already deleted by another worker we do not need to process it
+    // This should solve any possible as record in a db cannot be delete twice
+    const deleteResult =
+      await this.whatsappPendingMessageRepo.delete(pendingMessageId);
+    if (deleteResult?.affected !== 1) {
```
aberonni, line 175:
```suggestion
    if (deleteResult?.affected === 0) {
```

Feels safer? I wonder if `affected` could be > 1. But maybe I'm wrong.

RubenGeo: Ids are unique in a table due to a database constaint. I think I tried your way but I think I got a type-error as `affected` is potentially undefined (if you would use typeorm with a different underlying db than postgres)

jannisvisser (approved):

PR #7392, opened 2025-10-15
mimetype check bugfix

Yuri-510:
[AB#38455](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38455) 

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/programs/program-attachments/program-attachments.controller.ts
+++ b/services/121-service/src/programs/program-attachments/program-attachments.controller.ts
@@ -76,6 +76,7 @@ export class ProgramAttachmentsController {
   ): Promise<CreateProgramAttachmentResponseDto> {
     const allowedMimeTypes = [
       'image/jpeg',
+      'image/jpg',
       'image/png',
       'image/gif',
       'image/webp',


services/121-service/src/programs/program-attachments/program-attachments.controller.ts
```diff
@@ -78,6 +74,19 @@ export class ProgramAttachmentsController {
...
+      'image/jpeg',
+      'image/png',
+      'image/gif',
+      'image/webp',
+      'application/msword',
+      'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+      'application/pdf',
+    ];
+    if (!allowedMimeTypes.includes(file.mimetype)) {
+      throw new BadRequestException('Something went wrong: Validation failed');
```
elwinschmitz: I thought we had a guideline to "only throw (generic) `Error`s"; (I think @RubenGeo had dealt with that at some point)

But it _does make sense_, to me at least, to throw something specific like this _in a controller_...

services/121-service/src/programs/program-attachments/program-attachments.controller.ts
```diff
@@ -78,6 +74,19 @@ export class ProgramAttachmentsController {
     @Param('programId', ParseIntPipe) programId: number,
     @Req() req: ScopedUserRequest,
   ): Promise<CreateProgramAttachmentResponseDto> {
+    const allowedMimeTypes = [
+      'image/jpeg',
+      'image/png',
+      'image/gif',
+      'image/webp',
+      'application/msword',
+      'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
```
elwinschmitz: This 'explicit list' is a lot more readable then the 'magical regex'! 👍👍


Would an Excel-sheet/workbook-file also fall under this type? (I guess not).

Or was that left out "on purpose"? (Quick-Check with Tys maybe?)

Also, I heard about failing to upload a JPG; Would adding `image/jpg` help in that regard? Or is is handled as the same type? (would also make sense to me)

Yuri-510: Ah good catch. JPG is indeed a different mime-type and I added it.

Excel was not supported by our magical regex.

PR #7383, opened 2025-10-14
feat: project settings -> fsp settings page

aberonni:
[AB#38563](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38563)

## Checklist before requesting a code review

- [x] I have performed a self-review of my code
- [ ] I have added tests for my changes, or: Adding tests is unnecessary/irrelevant
   - [x] Unit tests
   - [ ] E2E tests
- [x] I have asked the design team to review these changes, or: The changes do not touch the UI/UX
	- [x] Tal review
	- [ ] ~Gal review~ N/A
	- [x] Tijs review
- [x] I have made sure that all automated checks pass before requesting a review
- [x] I do not need any deviation from our PR guidelines
	- [x] Please wait for design review and decision on whether we should wait for tests to merge
- [x] I have updated the [121 Service Module Dependency Diagram in the wiki](https://github.com/global-121/121-platform/wiki/121-Service-Module-Diagram) when the 121 module dependencies have changed.

## Portal preview-deployment

https://happy-rock-0411d2003-7383.westeurope.3.azurestaticapps.net

PR #7393, opened 2025-10-15
chore: convert `FSP_SETTINGS` to a `Record`

aberonni:
[AB#38560](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38560)

## Describe your changes

This came up while working on #7383 because it caused a lot of useless checks in the code.

The main change is in the definition of `FSP_SETTINGS`:

```ts
// Before
export const FSP_SETTINGS: FspDto[] = [
// After
export const FSP_SETTINGS: Record<Fsps, FspDto> = {
```

Then changing all of the places where it was used accordingly.

The main advantage is that getting a single `FspDto` from this const is easier, quicker, and less error prone.

```ts
// Before
const fspName: Fsps = Fsps.excel;
const foundFsp = FSP_SETTINGS.find((fsp) => fsp.name === fspName);
if (!foundFsp) {
  // even though we can be certain this exists, we still need to check this to make TS happy
  throw new Error(`Fsp with name ${name} not found`);
} else {
  return foundFsp;
}

// After
const fspName: Fsps = Fsps.excel;
return FSP_SETTINGS[fspName];
```

Other advantages:
- There is a slight (probably negligible) performance advantage in accessing something from an object, as opposed to looping over an array
- Whenever a new Fsp is added, TS will complain if we don't add it to `FSP_SETTINGS`. That wouldn't have happened before

## Portal preview-deployment

https://happy-rock-0411d2003-7393.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/src/fsps/enums/fsp-name.enum.ts
+++ b/services/121-service/src/fsps/enums/fsp-name.enum.ts
@@ -6,7 +6,7 @@ export enum Fsps {
   airtel = 'Airtel',
   commercialBankEthiopia = 'Commercial-bank-ethiopia',
   excel = 'Excel',
-  // XXX: can we just remove this one? And then simplify FSP_SETTINGS accordingly?
+  // TODO AB#38595 remove this deprecated FSP
   deprecatedJumbo = 'Intersolve-jumbo-physical',
   nedbank = 'Nedbank',
   onafriq = 'Onafriq',


RubenGeo (dismissed):
Great! I approved it we can remove jumbo later if we want

PR #7426, opened 2025-10-27
Always add the scope queries to fail fast in CI

RubenGeo:
[AB#38684](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38684) 

- Always add the scoped related queries regardless if the user has scope or not 
  - This way the earlier bug would have failed early in CI (see test shard 4 of this [run](https://github.com/global-121/121-platform/actions/runs/18841151665/job/53753534057?pr=7426))
  - If as user has scope '' and a registration has scope '' the user would still see the correct registrations

After adding this I found 2 additional bugs

1. program attachments repository was scoped, but it should not be
2. findOneOrFail visa card was failing. The reason when reversing the order of the relations the original object was modified

It could be that this a negative impact on performance for users that are now browsing our platform without scope

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/utils/scope/createFindWhereOptions.helper.ts
+++ b/services/121-service/src/utils/scope/createFindWhereOptions.helper.ts
@@ -58,10 +58,6 @@ export function convertToScopedOptions<T, Options extends FindManyOptions<T>>(
   // Create default options if undefined
   const baseOptions: FindOptionsCombined<T> = options || {};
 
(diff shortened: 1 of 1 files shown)


jannisvisser (commented):
You say "If as user has scope '' and a registration has scope '' the user would still see the correct registrations". What about if a user has scope '' and registration does have some scope?

This possible negative impact of performance, what is your gut feel about magnitude?

jannisvisser (approved):

PR #7432, opened 2025-10-28
Prototype: Transfer Value Calculation (Kobo-style) — Spike

RubenGeo:
[AB#38785](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38785) 

## Describe your changes

This PR introduces a prototype for supporting dynamic transfer value calculation in the 121 platform, inspired by how KoboToolbox handles calculation logic in XLSForms.

**Key changes:**

Added logic to parse and evaluate calculation formulas (e.g., ${dragon} * 2, if(${age} < 65, 100, 200)) using the [expr-eval](vscode-file://vscode-app/Users/rubenvandervalk/Documents/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) npm package.
Preprocesses formulas to convert Kobo/XLSForm-style ${variable} references to plain variable names for compatibility with expr-eval.

**Why expr-eval?**

KoBo uses [Enketo](https://github.com/enketo/enketo) for calculations, but no official Node.js library exists for Enketo’s calculation engine.

I also considered running the Enketo Docker image locally (see [GitHub](https://github.com/enketo/enketo)), but Enketo Express is designed to render forms and process submissions for web clients — not to expose a general-purpose calculation API.

Another option was to use [`openrosa-xpath-evaluator`](https://github.com/enketo/openrosa-xpath-evaluator), the XPath engine used by Enketo/KoBo. However, that package assumes a browser DOM environment, which makes it difficult to use directly in a Node.js backend (though workarounds exist).

In the end, I chose to use [`expr-eval`](https://www.npmjs.com/package/expr-eval), one of the most widely used and lightweight expression evaluators. I tested all [scenarios](https://teams.microsoft.com/l/message/19:d5a2a22ac7a14a7395c24c3a2a458a9a@thread.skype/1761571913778?tenantId=d3ab9790-6ae2-4bd8-aa5e-02864483e7c7&groupId=55f2fa49-7d9b-4c63-b616-b986f3136c3a&parentMessageId=1761571913778&teamName=510%20-%20CVA&channelName=General&createdTime=1761571913778) described by Tijs, and it works for all of them except `round(number, places)`, as it only supports rounding to whole numbers (the `places` argument isn’t supported).

I recommend using `expr-eval` because it behaves closely enough to KoBo’s calculation logic without requiring complex workarounds. We can always switch to a different calculation engine later if user needs evolve.

**Why not dynamically calculate transfer value with for example a database view?**

Writing custom queries for all possible logic is too complex

- [ ] I have updated the [121 Service Module Dependency Diagram in the wiki](https://github.com/global-121/121-platform/wiki/121-Service-Module-Diagram) when the 121 module dependencies have changed.

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/registration/services/inclusion-score.service.ts
+++ b/services/121-service/src/registration/services/inclusion-score.service.ts
@@ -1,15 +1,15 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
+import { Parser } from 'expr-eval';
 import { Equal, Repository } from 'typeorm';
 
 import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
 import { ProgramRegistrationAttributeEntity } from '@121-service/src/programs/entities/program-registration-attribute.entity';
 import { RegistrationEntity } from '@121-service/src/registration/entities/registration.entity';
 import { RegistrationAttributeTypes } from '@121-service/src/registration/enum/registration-attribute.enum';
-import { RegistrationDataService } from '@121-service/src/registration/modules/registration-data/registration-data.service';
 import { RegistrationUtilsService } from '@121-service/src/registration/modules/registration-utils/registration-utils.service';
 import { RegistrationScopedRepository } from '@121-service/src/registration/repositories/registration-scoped.repository';
-
+import { RegistrationsPaginationService } from '@121-service/src/registration/services/registrations-pagination.service';
 @Injectable()
 export class InclusionScoreService {
   @InjectRepository(ProgramEntity)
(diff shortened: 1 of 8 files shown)


services/121-service/src/registration/services/inclusion-score.service.ts
```diff
@@ -18,39 +18,65 @@ export class InclusionScoreService {
...
+      '🚀 ~ InclusionScoreService ~ calculatePaymentAmountMultiplier ~ numericRegistrationProperties:',
+      valueObject,
+    );
+    const processedFormula = preprocessFormula(paymentAmountMultiplierFormula);
+    console.log(
+      '🚀 ~ InclusionScoreService ~ calculatePaymentAmountMultiplier ~ processedFormula:',
+      processedFormula,
+    );
+    const parser = new Parser();
+    const expr = parser.parse(processedFormula);
```
jannisvisser: so I guess this part would in reality happen on program create or update right?

PR #7434, opened 2025-10-29
Split language into language for UI and preferred language

nielsbom:
[AB#39147](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/39147) 

## Describe your changes

For the reviewer: I suggest starting at `shared/types` and the new files there, then spiraling out to other places.

- I split up languages into languages for web app users (`UILanguage`) and languages for PAs
- I added some explicit types for "full" and "partial" translations
- I made the split and connection between locale and languages a bit more explicit by renaming some things
- I renamed the language switcher component to locale switcher component (because that's what it is)
- `pt_BR` is now `pt` (we misused `pt_BR` for historical reasons, now no longer a UI language)
- corrected the language code for Indonesian: `in` is now `id`
- some minor refactorings (`addMessageTemplates` vs `addMessageTemplate`)

## Portal preview-deployment

https://happy-rock-0411d2003-7434.westeurope.3.azurestaticapps.net

### DIFF
--- a/interfaces/portal/src/app/services/translatable-string.service.ts
+++ b/interfaces/portal/src/app/services/translatable-string.service.ts
@@ -1,7 +1,7 @@
 import { inject, Injectable, LOCALE_ID } from '@angular/core';
 
 import { UILanguage } from '@121-service/src/shared/enum/ui-language.enum';
-import { UILanguageTranslationPartial } from '@121-service/src/shared/types/ui-language-translation-partial.type';
+import { UILanguageTranslation } from '@121-service/src/shared/types/ui-language-translation.type';
 
 import { getUILanguageFromLocale, Locale } from '~/utils/locale';
 
@@ -10,7 +10,7 @@ import { getUILanguageFromLocale, Locale } from '~/utils/locale';
  * strings.
  *
  * Certain data structures, mostly DTOs in the 121 system use
- * UILanguageTranslationPartial types to store strings in 0..n languages. This service
+ * UILanguageTranslation types to store strings in 0..n languages. This service
  * can translate those strings. This can be done one by one using the
  * `translate` method, or for lists of strings using the `commaSeparatedList`
  * method.
@@ -22,7 +22,7 @@ export class TranslatableStringService {
   private currentLocale = inject<Locale>(LOCALE_ID);
 
   translate(
-    value: null | number | string | UILanguageTranslationPartial | undefined,
+    value: null | number | string | UILanguageTranslation | undefined,
   ): string | undefined {
     if (value === null || value === undefined) {
       return undefined;
@@ -65,7 +65,7 @@ export class TranslatableStringService {
   }
 
   commaSeparatedList(
-    values: string[] | UILanguageTranslationPartial[],
+    values: string[] | UILanguageTranslation[],
     style: Intl.ListFormatStyle = 'narrow',
   ): string {
     const list = values
--- a/services/121-service/src/programs/entities/program.entity.ts
+++ b/services/121-service/src/programs/entities/program.entity.ts
@@ -11,7 +11,7 @@ import { ProgramAttachmentEntity } from '@121-service/src/programs/program-attac
 import { RegistrationEntity } from '@121-service/src/registration/entities/registration.entity';
 import { Attribute } from '@121-service/src/registration/enum/registration-attribute.enum';
 import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
-import { UILanguageTranslationPartial } from '@121-service/src/shared/types/ui-language-translation-partial.type';
+import { UILanguageTranslation } from '@121-service/src/shared/types/ui-language-translation.type';
 import { WrapperType } from '@121-service/src/wrapper.type';
 
 @Entity('program')
@@ -20,7 +20,7 @@ export class ProgramEntity extends Base121Entity {
   public location: string | null;
 
   @Column('json', { nullable: true })
-  public titlePortal: UILanguageTranslationPartial | null;
+  public titlePortal: UILanguageTranslation | null;
 
   @Column({ type: 'character varying', nullable: true })
   public ngo: string | null;
@@ -51,7 +51,7 @@ export class ProgramEntity extends Base121Entity {
   public targetNrRegistrations: number | null;
 
   @Column('json', { nullable: true })
-  public description: UILanguageTranslationPartial | null;
+  public description: UILanguageTranslation | null;
 
   @Column({ default: true })
   public validation: boolean;
--- a/interfaces/portal/src/app/components/data-list/data-list.component.ts
+++ b/interfaces/portal/src/app/components/data-list/data-list.component.ts
@@ -16,7 +16,7 @@ import { RouterLink } from '@angular/router';
 
 import { SkeletonModule } from 'primeng/skeleton';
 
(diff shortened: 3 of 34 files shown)


services/121-service/src/shared/enum/registration-preferred-language.enum.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Source: https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes
+ * This is "Set 1" which are all 2 letter codes. The list does change sometimes.
+ */
+export enum RegistrationPreferredLanguage {
+  aa = 'aa',
+  ab = 'ab',
+  ae = 'ae',
+  af = 'af',
+  ak = 'ak',
+  am = 'am',
```
jannisvisser: why do we include all languages here, and not in the ui-language? Is the difference that the user could theoretically customize to use any language for messages?

interfaces/portal/src/app/utils/locale.ts
```diff
@@ -1,36 +1,52 @@
...
+  es = UILanguage.es,
+  fr = UILanguage.fr,
+  nl = UILanguage.nl,
+  sk = UILanguage.sk,
+  /* eslint-enable @typescript-eslint/prefer-literal-enum-member -- emphasize the relationship between this enum and UILanguage */
 }
 
+const localeToUILanguageMap: Record<Locale, UILanguage> = {
+  [Locale.ar]: UILanguage.ar,
+  [Locale.en]: UILanguage.en,
```
jannisvisser: is it correct that here you do use the enum and not 'en-GB'? Otherwise I'm not following.

interfaces/portal/src/app/services/linguonyms/spanish.ts
```diff
@@ -0,0 +1,187 @@
...
+  cy: 'galés',
+  da: 'danés',
+  de: 'alemán',
+  dv: 'maldivo (o dhivehi)',
+  dz: 'dzongkha',
+  ee: 'ewé',
+  el: 'griego (moderno)',
+  en: 'inglés',
+  eo: 'esperanto',
+  es: 'español (o castellano)',
```
jannisvisser: is this intentional with lowercase letter and including the (o castellano)? This is how this shows in the language switcher now.

Image

interfaces/portal/src/app/services/get-registration-preferrred-language-name.service.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
+import { UILanguage } from '@121-service/src/shared/enum/ui-language.enum';
+
+import { languageCodesToLinguonymsForArab } from '~/services/linguonyms/arab';
+import { languageCodesToLinguonymsForDutch } from '~/services/linguonyms/dutch';
+import { languageCodesToLinguonymsForEnglish } from '~/services/linguonyms/english';
+import { languageCodesToLinguonymsForFrench } from '~/services/linguonyms/french';
+import { languageCodesToLinguonymsForSlovak } from '~/services/linguonyms/slovak';
+import { languageCodesToLinguonymsForSpanish } from '~/services/linguonyms/spanish';
```
jannisvisser: I guess this answers my previous question in part (about something breaking). I guess my question is my general: is it clear enough what to do upon adding a new ui language? Is there documentation also?

interfaces/portal/src/app/services/translatable-string.service.ts
```diff
@@ -26,28 +36,36 @@ export class TranslatableStringService {
...
+      return undefined;
+    }
+
+    if (Object.keys(translationMapping).length === 0) {
+      return undefined;
     }
 
-    return undefined;
+    // Just the first available language.
+    return translationMapping[Object.keys(translationMapping)[0] as UILanguage];
```
jannisvisser: is this whole if-structure above reason enough for a unit test on this?

services/121-service/src/programs/entities/program.entity.ts
```diff
@@ -84,7 +84,7 @@ export class ProgramEntity extends Base121Entity {
   public fullnameNamingConvention: string[] | null;
 
   @Column('json', { default: [] })
-  public languages: LanguageEnum[];
+  public languages: RegistrationPreferredLanguage[];
```
jannisvisser: is this indeed completely unrelated to UIlanguage?

services/121-service/src/registration/mappers/registration-views.mapper.ts
```diff
@@ -142,10 +143,9 @@ export class RegistrationViewsMapper {
...
     value: unknown,
   ): unknown {
+    const english = RegistrationPreferredLanguage.en;
     const selectedOption = attribute.options?.find((o) => o.option === value);
-    if (selectedOption && selectedOption.label && selectedOption.label['en']) {
-      return selectedOption.label['en'];
-    }
-    return value;
+
+    return selectedOption?.label?.[english] ?? value;
```
jannisvisser: is the significance of english here actually to be fallbackLanguage? If so, we better name it as such.

services/121-service/src/registration/validators/registrations-input-validator.ts
```diff
@@ -402,9 +402,12 @@ export class RegistrationsInputValidator {
...
   ): Record<string, string> {
-    const languageNamesApi = new Intl.DisplayNames(['en'], {
-      type: 'language',
-    });
+    const languageNamesApi = new Intl.DisplayNames(
+      [RegistrationPreferredLanguage.en],
+      {
+        type: 'language',
+      },
+    );
```
jannisvisser: it seems off here now, that we use this api here, whereas elsewhere we now use fully defined lists of languages and list of language linguonyms defined in our repo. Could those not also use this API? Or vice versa, could this use a list like that as well?

services/121-service/src/shared/types/ui-language-translation-partial.type.ts
```diff
@@ -0,0 +1,14 @@
+import { UILanguage } from '@121-service/src/shared/enum/ui-language.enum';
+import { TranslationPartial } from '@121-service/src/shared/types/translation-partial.type';
+
+/**
+ * An object that contains 0..n string translations for languages the UI
+ * supports. Use this if you want to allow incomplete translations.
```
jannisvisser: can this (and the preferred lanuage variant of this) use a bit more explanation? Maybe some example of full vs partial? In fact, the 'full' variant is only used for preferred languages and only for the linguonyms right? Doesn't this put a bit of unjustified focus on the partial nature here, with also the phrase 'allowing incomplete translations' sound a big negative, while it's in fact the default apparently.

jannisvisser (changes requested):
Nice splitting up :)

services/121-service/src/shared/enum/registration-preferred-language.enum.ts
nielsbom: We want the 121 Platform to communicate with PAs in more languages than the languages we offer the platform in. This is already the case: Ukranian is used as a RegistrationPreferredLanguage but we don't have it as a UI language.

services/121-service/src/programs/entities/program.entity.ts
jannisvisser: todo: am

PR #7435, opened 2025-10-31
payment lock instead of actions

RubenGeo:
[AB#38760](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38760) 

## Portal preview-deployment


services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -80,17 +70,15 @@ export class PaymentsExecutionService {
...
       );
 
-      void this.createTransactionJobs({
+      await this.createTransactionJobs({
         programId,
         transactionIds: transactionsOfIncludedRegistrations.map((t) => t.id),
         userId,
         isRetry: false,
       });
     } finally {
```
RubenGeo: I added the await here:
- If we do not await here there is a significant time window where the payment is not lock but there are no transaction jobs in the queue
- another option is to have a .catch, .than at createTransactionJobs

but since we want to increase performance anyway I am now puzzeling with that

services/121-service/src/payments/services/payments-creation.service.ts
```diff
@@ -85,45 +81,47 @@ export class PaymentsCreationService {
         );
       }
 
+      await this.paymentsHelperService.checkFspConfigurationsOrThrow(
+        programId,
+        programFspConfigurationNames,
+      );
```
jannisvisser: this should also go inside createPaymentDryRunPart according to the logic I was aiming for (put all shared code in there)

services/121-service/src/payments/services/payments-execution.service.ts
jannisvisser: - Sure good point, but how long does startPayment now take for e.g. 100k transactions?
- Also the performance you're working on was on createPayment right? so unrelated?

services/121-service/test/payment/payment-in-progress.test.ts
```diff
@@ -1,59 +1,117 @@
...
 } from '@121-service/test/helpers/utility.helper';
 import {
   programIdOCW,
   programIdPV,
-  registrationOCW4,
   registrationsOCW,
 } from '@121-service/test/registrations/pagination/pagination-data';
 
-// NOTE: these test focus on payment being in progress because of active queue. Being in-progress because of actions is too short to cover with API-tests (with low volumes) and is either way better already covered by unit test on payments-creation.service.spec.ts
+// NOTE: these test focus on payment being in progress because of active queue or payment lock. Being in-progress because of actions is too short to cover with API-tests (with low volumes) and is either way better already covered by unit test on payments-creation.service.spec.ts
```
jannisvisser: this comment still mentions actions so seems out of date. Because of your change of focus, the whole comment can be removed I think.

jannisvisser (changes requested):
yes, this is better :) why did you opt for the boolean column in the end? 

see some comments.

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -80,17 +68,15 @@ export class PaymentsExecutionService {
         },
       );
 
-      void this.createTransactionJobs({
+      await this.createTransactionJobs({
```
jannisvisser: add ##TODO

jannisvisser (approved):
see reamining comments, but feel free to merge

services/121-service/src/payments/services/payments-creation.service.ts
```diff
@@ -85,51 +78,45 @@ export class PaymentsCreationService {
...
+        await this.paymentsProgressHelperService.unlockPaymentsForProgram(
           programId,
-          ActionType.endBlockNewPayment,
         );
       }
     }
   }
 
-  private async createPaymentDryRunPart({
+  private async doPaymentDryRunFlow({
```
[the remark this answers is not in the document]
RubenGeo: It also throws now so I made the name more generic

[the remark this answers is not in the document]
RubenGeo: After looking at it again: getPaymentDryRunDetailsOrThrow @jannisvisser

PR #7507, opened 2025-11-10
docs: Amend hotfix-policy

elwinschmitz:
[AB#38990](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/38990)

## Portal preview-deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,11 +406,12 @@ Azure App Service uses deployment slots that can cause temporary database errors
 
 ### Hotfix-release Checklist
 
-Only in rare, very specific circumstances it might be required to do a "Hotfix-release". _**Only** when_:
+Only in rare, very specific circumstances it might be required to do a "hotfix-release". For example when:
 
-- a critical issue affects 1 (or a few) instances
-- some migration-steps are already in `main`
-- new (incomplete) features, that are not yet communicated to end-users, are already in `main`
+- A new (incomplete) feature, that is not yet communicated to end-users, is already in `main`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -404,7 +404,13 @@ Azure App Service uses deployment slots that can cause temporary database errors
...
 **Recognition**: Errors only appear during releases and disappear when you recheck the same endpoint.
 
-### Patch/Hotfix Checklist
+### Hotfix-release Checklist
+
+Only in rare, very specific circumstances it might be required to do a "Hotfix-release". _**Only** when_:
+
+- a critical issue affects 1 (or a few) instances
+- some migration-steps are already in `main`
+- new (incomplete) features, that are not yet communicated to end-users, are already in `main`
```
jannisvisser: - A specific scenario that we mentioned in the chit-chat was: if main contains features that warrant a complex migration (to be carefully done outside of office hours), while the hotfix is urgent to be done before that. This can be added on the same "level" as the last bullet I think.
- The first bullet on its own is not actually reason enough for a hotfix I think, because you could still deploy to all (if not one of the other bullets applies).
- I'm not actually getting what you mean with bullet 2.

jannisvisser (changes requested):

README.md
elwinschmitz: I _think_ what I meant in bullet 2 is what you described as the scenario we've talked about. :)

And the first bullet was indeed meant to be "this at least/first AND these other requirements"..

But lets give it another try altogether. :D

PR #7675, opened 2025-12-18
feat: fail early on missing FSP environment variables

nielsbom:
[AB#39299](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/39299) 

## Describe your changes

❗ This PR both adds and removes a bunch of FSP related environment variables. The relevant instances need to be updated.

- add `{FSP}_MODE` environment variables for each FSP
- remove `{FSP}_ENABLED` everywhere
- remove `MOCK_{FSP}` everywhere
- dynamically construct `FSP_ENV_VARIABLE_SETTINGS`
- add `FspEnvVariableValidationService` to Payments module
- run `FspEnvVariableValidationService` on Payments module init

- [x] I have performed a self-review of my code (and addressed any Copilot comments)
- [x] I have updated all [documentation](https://github.com/global-121/121-internal-documentation/blob/main/documentation.md) where necessary

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/fsp-integrations/shared/enum/fsp-mode.enum.ts
+++ b/services/121-service/src/fsp-integrations/shared/enum/fsp-mode.enum.ts
@@ -1,34 +1,39 @@
 export enum FspMode {
-  // This specific FSP integration is explicitly disabled. It should not be
-  // configurable using the backend API or even visible in the admin UI. See
-  // AB#39825 and AB#38588. In this mode we do *not* require other environment
-  // variables to be set for the FSP.
-  //
-  // This is the default value.
-  //
-  // This is used:
-  // - in production
-  // - locally (potentially)
+  /**
+   * This specific FSP integration is explicitly disabled. It should not be
+   * configurable using the backend API or visible in the admin UI. See AB#39825
+   * and AB#38588. In this mode we do *not* require other environment variables
+   * to be set for the FSP.
+   *
+   * This is the default value.
+   *
+   * This is used:
+   * - in production
+   * - locally (potentially)
+   */
   disabled = 'DISABLED',
-
-  // The 121-service talks to the mock service for this FSP. In this mode we
-  // don't require other environment variables to be set for the FSP.
-  //
-  // This is used:
-  // - locally (dev and testing)
-  // - in CI
-  // - in environments where we don't want to use the real external service
+  /**
+   * The 121-service talks to the mock service for this FSP. In this mode we
+   * don't require other environment variables to be set for the FSP.
+   *
+   * This is used:
+   * - locally (dev and testing)
+   * - in CI
+   * - in environments where we don't want to use the real external service
+   */
   mock = 'MOCK',
 
-  // The 121-service talks to a real external service for this FSP.
-  // In this mode we require all other environment variables for the FSP to be
-  // set.
-  //
-  // This is used:
-  // - locally against sandbox/staging environments (during initial development
-  //   of an FSP integration for example)
-  // - in demo/staging environments against sandbox/staging environments of the
-  //   FSP
-  // - in production environments against the production environment of the FSP
+  /**
+   * The 121-service talks to a real external service for this FSP.
+   * In this mode we require all other environment variables for the FSP to be
+   * set.
+   *
+   * This is used:
+   * - locally against sandbox/staging environments (during initial development
+   *   of an FSP integration for example)
+   * - in demo/staging environments against sandbox/staging environments of the
+   *   FSP
+   * - in production environments against the production environment of the FSP
+   */
   external = 'EXTERNAL',
 }


services/121-service/src/payments/services/fsp-env-variable-validation.service.spec.ts
jannisvisser: very clear!

services/121-service/src/payments/payments.module.ts
```diff
@@ -71,4 +74,25 @@ import { createScopedRepositoryProvider } from '@121-service/src/utils/scope/cre
     PaymentsProgressHelperService,
   ],
 })
-export class PaymentsModule {}
+export class PaymentsModule implements OnModuleInit {
+  private readonly logger = new Logger(PaymentsModule.name);
```
jannisvisser: this is not something we're using anywhere else already right? What is the added value of deviating from the standard?

services/121-service/src/payments/services/fsp-env-variable-validation.service.ts
```diff
@@ -0,0 +1,59 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { Fsps } from '@121-service/src/fsp-management/enums/fsp-name.enum';
+import { FspEnvVariablesDto } from '@121-service/src/fsp-management/fsp-env-variables.dto';
+
+@Injectable()
+export class FspEnvVariableValidationService {
+  private readonly logger = new Logger(FspEnvVariableValidationService.name);
+
+  // A pure function.
```
jannisvisser: what do you mean with this? My understanding is that a function that is part of a class, is a method. 

Also when you don't define private or public, I though the default was private. But if so, then how can you call it in the spec file? 

Please enlighten me :)

services/121-service/src/payments/payments.module.ts
```diff
@@ -71,4 +74,25 @@ import { createScopedRepositoryProvider } from '@121-service/src/utils/scope/cre
...
+
+  onModuleInit() {
```
jannisvisser: so when does this run? on application startup, or is it lazy-loaded whenever the payment module is used?

Either way, it seems that these env-variables are mostly needed for 'do payment', while the payment module has wider use than that (e.g. getting payments/transactions). Shouldn't we tie it more to that by doing this at payment creation or something? Actually, I'm now reading the original task-title from Ruben where he suggested to do this in payment-execution service, which means at actually starting/retrying the payment. In that case we could (potentially) also handle errors by letting transactions fail, which is a very standard and clear pattern for us and our users. In your current case, what would a failure look like?

services/121-service/src/fsp-management/fsp-specific/airtel/airtel-env-variable-settings.const.ts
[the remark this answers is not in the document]
nielsbom: Yes, I'll also move this.

services/121-service/src/payments/payments.module.ts
nielsbom: Bit of copypasta from another reply:

- I wanted the application to **not** start when we have missing FSP variables, basically fail early and loud and at the moment when a developer is already focused on the instance/application. Failing when a payment is created and then having an end user going through all the channels to end up at a developer to fix this is just a lot of unnecessary overhead imho.
- using `onModuleInit` on the PaymentsModule seems the right place to hook into application startup

Let's discuss

services/.env.example
```diff
@@ -292,6 +292,8 @@ KOBO_CONNECT_API_URL=https://kobo-connect.example.org
 # ---------------------------
 # Third-party FSP: Intersolve
 # ---------------------------
+INTERSOLVE_VOUCHER_PAPER_ENABLED=false
+INTERSOLVE_VOUCHER_WHATSAPP_ENABLED=false
```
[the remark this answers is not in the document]
nielsbom: Good point, I'll combine this into one and add a comment for [our future refactor of Intersolve Voucher](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/10288)

services/121-service/src/payments/payments.module.ts
nielsbom: Agreed with Jannis to keep this.

services/121-service/src/payments/services/fsp-env-variable-validation.service.ts
nielsbom: Jannis abides.

services/121-service/src/fsp-integrations/settings/fsp-env-variable-settings.const.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// Please keep sorted for readability. (doing this with ESLint requires a custom rule)
+export const FSP_ENV_VARIABLE_SETTINGS: FspEnvVariableSettingsRecord = {
+  [Fsps.airtel]: AIRTEL_ENV_VARIABLE_SETTINGS,
+  [Fsps.commercialBankEthiopia]: COMMERCIAL_BANK_ETHIOPIA_ENV_VARIABLE_SETTINGS,
+  [Fsps.cooperativeBankOfOromia]:
+    COOPERATIVE_BANK_OF_OROMIA_ENV_VARIABLE_SETTINGS,
+  [Fsps.excel]: EXCEL_ENV_VARIABLE_SETTINGS,
+  [Fsps.intersolveVisa]: INTERSOLVE_VISA_ENV_VARIABLE_SETTINGS,
+  ['Intersolve-voucher']: INTERSOLVE_VOUCHER_ENV_VARIABLE_SETTINGS,
```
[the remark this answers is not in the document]
nielsbom: This is an exception, as commented on above. No action.

services/121-service/src/payments/payments.module.ts
```diff
@@ -71,4 +74,29 @@ import { createScopedRepositoryProvider } from '@121-service/src/utils/scope/cre
...
+
+    const messagePrefix = 'FSP environment variable validation';
+    if (validationResult.ok) {
+      this.logger.log(
+        `${messagePrefix} succeeded, ${validationResult.messages[0]}`,
+      );
+    } else {
+      validationResult.messages.forEach((msg) => this.logger.error(msg));
+      // We don't throw an HTTPException here because we want the application
+      // to not start in case of a misconfiguration with environment variables.
```
RubenGeo:
```suggestion
```

services/.env.example
```diff
@@ -292,6 +292,7 @@ KOBO_CONNECT_API_URL=https://kobo-connect.example.org
 # ---------------------------
 # Third-party FSP: Intersolve
 # ---------------------------
+INTERSOLVE_VOUCHER_ENABLED=false
```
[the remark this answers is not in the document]
RubenGeo: there are items created for this

RubenGeo (approved):
I really like this `MODE` approach. I left 2 optional comments.

Be sure to update production env variables accordingly or make tasks in the release to do it later. 

Maybe it's also good to inform the rest of the dev team they need to change their local setup after we merge this

PR #7683, opened 2025-12-24
Add performance test for exporting high volume of registrations

Piotrk39:
[AB#39755](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/39755) 

## Describe your changes

Adds old ExportHighVolumeOfRegistrations in the form of performance test for 100k+ registrations in OCW program.

## Portal preview-deployment

### DIFF
--- a/services/121-service/test/performance/export-100k-registrations-ocw.test.ts
+++ b/services/121-service/test/performance/export-100k-registrations-ocw.test.ts
@@ -12,11 +12,13 @@ import {
   getAccessToken,
   resetDB,
 } from '@121-service/test/helpers/utility.helper';
+import { calculateMilliseconds } from '@121-service/test/performance/helpers/performance.helper';
 import { programIdOCW } from '@121-service/test/registrations/pagination/pagination-data';
 
 const duplicateLowNumber = 5;
 const duplicateHighNumber = 17; // cronjob duplicate number should be 2^17 = 131072
-const testTimeout = 150_000; // 2,5 minutes
+const testTimeout = calculateMilliseconds({ minutes: 3 }); // Overall test timeout to prevent hanging
+const maximumExportTime = calculateMilliseconds({ minutes: 2.5 }); // Performance assertion limit for export operation
 const duplicateNumber =
   // eslint-disable-next-line n/no-process-env -- Required to detect high data volume mode for performance testing
   process.env.HIGH_DATA_VOLUME === 'true'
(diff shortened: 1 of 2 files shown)


services/121-service/test/performance/export-100k-registrations-ocw.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    // Export registrations
+    const exportResponse = await exportAllRegistrations(
+      programIdOCW,
+      accessToken,
+    );
+
+    expect(exportResponse.statusCode).toBe(HttpStatus.OK);
+
+    // Verify we exported a significant number of registrations
+    const expectedMinRegistrations =
```
nielsbom: nitpick 
```suggestion
    const expectedMinimumNumberOfRegistrations =
```

services/121-service/test/performance/export-100k-registrations-ocw.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      duplicateNumber === duplicateHighNumber ? 100000 : 25;
+    expect(exportResponse.body.data.length).toBeGreaterThan(
+      expectedMinRegistrations,
+    );
+    const elapsedTime = Date.now() - startTime;
+    expect(elapsedTime).toBeLessThan(testTimeout);
```
nielsbom: `testTimeout` is playing 2 roles: it's
- used to time out the test so that it doesn't take too long
- we assert that the elapsedTime is less than `testTimeout`

But if the test would ever take longer than `testTimeout` we'd never reach this line. So either we should have 2 different variables (`testTimeout` and `maximumTimeAllowed` for example) because we want to use them for different things, or we _only_ use `testTimeout` for the actual timeout and let the test just fail if takes too long.

I vote for introducing another variable because we don't really care about how long the setup takes, just how long exporting the 100K registrations takes. We can then use `testTimeout` to make sure the test as a whole times out when it's taking way too long.

services/121-service/test/performance/export-100k-registrations-ocw.test.ts
```diff
@@ -0,0 +1,72 @@
...
+const duplicateHighNumber = 17; // cronjob duplicate number should be 2^17 = 131072
+const testTimeout = 150_000; // 2,5 minutes
+const duplicateNumber =
+  // eslint-disable-next-line n/no-process-env -- Required to detect high data volume mode for performance testing
+  process.env.HIGH_DATA_VOLUME === 'true'
+    ? duplicateHighNumber
+    : duplicateLowNumber;
+
+jest.setTimeout(testTimeout);
+describe('Export High volume of registrations', () => {
```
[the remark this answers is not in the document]
Piotrk39: Changed!

services/121-service/test/performance/export-100k-registrations-ocw.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  let accessToken: string;
+
+  it('Setup, add registrations and export', async () => {
+    // Arrange
+    const startTime = Date.now();
```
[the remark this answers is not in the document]
Piotrk39: You are kind of right it was not clear where exactly the timers should start and if we want to measure the flow's entire performance or just a focused bit. I believe it is how I translated old K6 tests that had the timeout connected with the timer so we ended up with something similar here too. I'll take a look!

services/121-service/test/performance/export-100k-registrations-ocw.test.ts
Piotrk39: Done!

services/121-service/test/performance/export-100k-registrations-ocw.test.ts
Piotrk39: Changed

nielsbom (approved):

PR #7703, opened 2026-01-02
Use koboform to update 121 program and program registration attributes

RubenGeo:
[AB#39717](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/39717) 

## Describe your changes

- Adds program registration attributes based on the deployed kobo form
- Updates program languages based on the deployed kobo form
 
- [x] I have updated all [documentation](https://github.com/global-121/121-internal-documentation/blob/main/documentation.md) where necessary - `seperate task for that`

## Portal preview-deployment

https://happy-rock-0411d2003-7703.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/src/kobo/services/kobo.validation.service.spec.ts
+++ b/services/121-service/src/kobo/services/kobo.validation.service.spec.ts
@@ -236,7 +236,7 @@ describe('KoboValidationService', () => {
       expect(error.getStatus()).toBe(HttpStatus.BAD_REQUEST);
       expect(error.message).toMatchInlineSnapshot(`
          "Kobo form definition validation failed:
-         - Kobo form attribute "nationalId" has incompatible type for 121 attribute: expected: "background-audio, xml-external, acknowledge, audio, barcode, calculate, date, dateTime, file, geopoint, geoshape, geotrace, hidden, image, rank, select_multiple_from_file, select_multiple, text, time, video", got "integer"  "
+         - Kobo form attribute "nationalId" has incompatible type for 121 attribute, expected one of the following types: "background-audio, xml-external, acknowledge, audio, barcode, calculate, date, dateTime, file, geopoint, geoshape, geotrace, hidden, image, rank, select_multiple_from_file, select_multiple, text, time, video", got "integer"  "
         `);
     });
   });
@@ -497,7 +497,7 @@ describe('KoboValidationService', () => {
       expect(error.getStatus()).toBe(HttpStatus.BAD_REQUEST);
       expect(error.message).toMatchInlineSnapshot(`
          "Kobo form definition validation failed:
-         - Kobo form attribute "phoneNumber" has incompatible type for 121 attribute: expected: "text", got "integer"  "
+         - Kobo form attribute "phoneNumber" has incompatible type for 121 attribute, expected one of the following types: "text", got "integer"  "
         `);
     });
   });
@@ -657,7 +657,7 @@ describe('KoboValidationService', () => {
       expect(error.getStatus()).toBe(HttpStatus.BAD_REQUEST);
       expect(error.message).toMatchInlineSnapshot(`
        "Kobo form definition validation failed:
-       - Kobo form attribute "scope" has incompatible type for 121 attribute: expected: "background-audio, xml-external, acknowledge, audio, barcode, calculate, date, dateTime, file, geopoint, geoshape, geotrace, hidden, image, rank, select_multiple_from_file, select_multiple, text, time, video", got "integer"  "
+       - Kobo form attribute "scope" has incompatible type for 121 attribute, expected one of the following types: "background-audio, xml-external, acknowledge, audio, barcode, calculate, date, dateTime, file, geopoint, geoshape, geotrace, hidden, image, rank, select_multiple_from_file, select_multiple, text, time, video", got "integer"  "
       `);
     });
   });
--- a/services/121-service/src/kobo/services/kobo.validation.service.ts
+++ b/services/121-service/src/kobo/services/kobo.validation.service.ts
@@ -362,7 +362,7 @@ export class KoboValidationService {
     }
 
     if (!expectedKoboTypes.includes(surveyItemType)) {
-      return `Kobo form attribute "${attributeName}" has incompatible type for 121 attribute: expected: "${expectedKoboTypes.join(', ')}", got "${surveyItemType}"  `;
+      return `Kobo form attribute "${attributeName}" has incompatible type for 121 attribute, expected one of the following types: "${expectedKoboTypes.join(', ')}", got "${surveyItemType}"  `;
     }
   }
 
--- a/services/121-service/src/kobo/mappers/kobo.mapper.spec.ts
+++ b/services/121-service/src/kobo/mappers/kobo.mapper.spec.ts
@@ -241,9 +241,6 @@ describe('KoboMapper', () => {
                 name: 'favoriteSize',
                 select_from_list_name: 'sizes',
               }),
-              createSurveyItem({
-                name: 'textField',
-              }), // Check if mapper does not break if select_from_list_name is not there for a choice
             ],
             choices: [
               createChoice({


services/mock-service/src/kobo/kobo.mock.service.ts
```diff
@@ -0,0 +1,387 @@
...
+              label: ['Car', 'Auto'],
+              list_name: 'xg8ep41',
+              $autovalue: 'car',
+            },
+          ],
+        },
+      },
+    };
+
+    if (uid_asset === KoboMockAssetUids.happyFlowWithChanges) {
```
nielsbom: Move this outside of the class. At first I thought the new form definition was created dynamically based on the input but it isn't. Moving it out of the controller makes that more clear because it's clear that it's done exactly once. I'd put it in an IIFE and assign it, see [example code](https://gist.github.com/nielsbom/462ac6358f22d6ba0c558f840e0c294e).

services/mock-service/src/kobo/kobo.mock.service.ts
```diff
@@ -0,0 +1,387 @@
...
+      const happyFlowFromDefinitionWithChanges = happyFlowFromDefinition;
```
nielsbom: I'd use structuredClone here, even if you keep the code inside the service method.

services/121-service/src/kobo/entities/kobo.entity.ts
```diff
@@ -0,0 +1,31 @@
+import { Column, Entity, Index, JoinColumn, OneToOne, Relation } from 'typeorm';
+
+import { Base121Entity } from '@121-service/src/base.entity';
+import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
+
+@Entity('kobo')
+export class KoboEntity extends Base121Entity {
+  @Column({ nullable: false, unique: true, type: 'varchar' })
+  public assetId: string;
```
nielsbom: For consistency change this to `assetUid`.

services/121-service/src/kobo/kobo.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+    status: HttpStatus.CREATED,
+    description:
+      'The Kobo form has been successfully integrated with the program',
+    type: KoboIntegrationResultDto,
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    description:
+      'Dry run completed successfully - the integration would succeed if executed',
+    type: KoboIntegrationResultDto,
```
nielsbom: The example response in Swagger has `"dryRun": false` which probably isn't an actual response.

services/121-service/src/kobo/kobo.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+  })
+  @ApiResponse({
+    status: HttpStatus.BAD_REQUEST,
+    description:
+      'Validation failed - the Kobo form does not meet program requirements. ',
+  })
+  @ApiResponse({
+    status: HttpStatus.UNAUTHORIZED,
+    description:
+      'User is not authenticated, Kobo API token is invalid, or user lacks admin privileges',
```
nielsbom: nitpick - I'd add examples here of all error states

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -0,0 +1,223 @@
...
+import { KoboApiService } from '@121-service/src/kobo/services/kobo-api.service';
+import { KoboSurveyProcessorService } from '@121-service/src/kobo/services/kobo-survey-processor.service';
+import { ProgramFspConfigurationRepository } from '@121-service/src/program-fsp-configurations/program-fsp-configurations.repository';
+import { ProgramService } from '@121-service/src/programs/programs.service';
+import { ProgramRepository } from '@121-service/src/programs/repositories/program.repository';
+import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
+
+@Injectable()
+export class KoboService {
+  // kobo repo typeorm import
```
nielsbom: Can be removed I guess?

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -0,0 +1,223 @@
...
+      programId,
+    });
+
+    return {
+      message: 'Kobo form integrated successfully',
+      dryRun: false,
+    };
+  }
+
+  public async upsertKoboEntity({
```
nielsbom: This can be private, only called from this class.

services/121-service/src/kobo/services/kobo.service.ts
nielsbom: This file also contains quite some logic, I would expect some tests here too.

services/121-service/src/kobo/services/kobo-api.service.ts
```diff
@@ -0,0 +1,73 @@
...
+  public async getDeployedAssetOrThrow({
+    assetId,
+    token,
+    baseUrl,
+  }: {
+    assetId: string;
+    token: string;
+    baseUrl: string;
+  }): Promise<KoboFormDefinition> {
+    const apiUrl = `${baseUrl}/api/v2/assets/${assetId}/deployment/?format=json`;
```
nielsbom: > When constructing URLs, URL parameters or Headers for use with [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) use the available interfaces/objects for them: [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) and [Headers ](https://developer.mozilla.org/en-US/docs/Web/API/Headers). 

[source](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#g-use-url-and-header-objects-for-constructing-and-manipulating-urls-and-headers)

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -0,0 +1,223 @@
...
+      languages: combinedLanguages,
+    });
+  }
+
+  private getLanguageIsoCodes({
+    koboLanguages: koboLanguages,
+  }: {
+    koboLanguages: string[];
+  }): RegistrationPreferredLanguage[] {
+    const isoCodes: RegistrationPreferredLanguage[] = [];
```
nielsbom: nitpick - The format of how Kobo sends languages is a bit hidden 2 helper functions down. I suspect this is also not a standard format, I looked into Kobo docs for this.

I'd make that a bit more explicit, with at least a comment, but better something like this and then some set logic to get language codes we have in our system.
```typescript
function extractLanguageCodes(input: string): string[] {
  // Pattern to match content within parentheses
  const pattern = /\(([a-z]{2})\)/gi;
  const matches = input.match(pattern);
  
  if (!matches) return [];
  
  // Extract the language codes without parentheses and convert to lowercase
  return matches.map(match => match.replace(/[()]/g, '').toLowerCase());
}

// Example usage
const input = "English (en)', 'Dutch (nl)";
const codes = extractLanguageCodes(input);
console.log(codes); // Output: ['en', 'nl']
```

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -0,0 +1,223 @@
...
+    });
+
+    if (dryRun) {
+      return {
+        message: 'Dry run successful - validation passed',
+        dryRun: true,
+      };
+    }
+
+    await this.upsertKoboEntity({
```
nielsbom: If possible good to take the following three persisting steps in a database transaction.

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -0,0 +1,223 @@
...
+    programId: number;
+    assetId: string;
+    token: string;
+    url: string;
+  }): Promise<void> {
+    const existingKoboEntity = await this.koboRepository.findOne({
+      where: { programId: Equal(programId) },
+    });
+
+    if (existingKoboEntity) {
```
nielsbom: I'd expect a mapper here, mapping from `koboFormDefinition` to `KoboEntity`.

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -0,0 +1,321 @@
...
+
+        if (errorMessage) {
+          errorMessages.push(errorMessage);
+        }
+      }
+    }
+    return errorMessages;
+  }
+
+  private validateFullNameNamingConventionInKoboSurveyItems({
```
nielsbom: I'd move this one method down, as `validateSurveyItemTypeMatchExpected121Type` is a helper for `validateFspAttributesTyping`.

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -0,0 +1,321 @@
...
+
+    // Only validate existence if empty phone numbers are not allowed
+    if (!phoneNumberItem && !allowEmptyPhoneNumber) {
+      errors.push(
+        'Kobo form must contain a question with name phoneNumber (should be a text type and country code should be included) or program.allowEmptyPhoneNumber must be set to true.',
+      );
+    }
+
+    // Also validate type if the phone number item exists
+    if (phoneNumberItem) {
```
nielsbom: I'd do an early return here.

services/121-service/src/kobo/services/kobo.validation.service.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+        expect(error.message).toMatchInlineSnapshot(`
+         "Kobo form definition validation failed:
+         - Kobo form attribute "nationalId" has incompatible type for 121 attribute: expected: "text, select_multiple, select_multiple_from_file, rank, geopoint, geotrace, geoshape, date, time, dateTime, image, audio, background-audio, video, file, barcode, calculate, acknowledge, hidden, xml-external", got "integer"  "
+        `);
+      });
+    });
+  });
+
+  describe('fullnameNamingConvention attributes validation', () => {
+    it('should pass validation when all fullnameNamingConvention attributes are present', async () => {
```
nielsbom:
```suggestion
    it('should pass validation when attributes are present', async () => {
```
Otherwise we're repeating the name in the describe.

services/121-service/src/kobo/services/kobo-api.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    // This should never happen but it makes TypeScript happy and throws a meaningful error
+    if (!responseBody.version_id || !responseBody.asset) {
+      throw new Error('Kobo information is missing version_id or asset');
+    }
+
+    const surveyItemsCleaned = KoboMapper.surveyItemsDtosToInterfaces({
+      koboSurveyItems: responseBody.asset.content.survey || [],
+    });
+    return {
+      name: responseBody.asset.name ?? '',
```
nielsbom: I would take this out of the api.service, I would only keep HTTP interaction and validation here.

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,585 @@
...
+
+    const programId = createProgramResponse.body.id;
+    const getProgramResponseBeforeKoboIntegration = await getProgram(
+      programId,
+      accessToken,
+    );
+    const programBeforeKoboIntegration =
+      getProgramResponseBeforeKoboIntegration.body;
+
+    // Act
```
nielsbom: A lot of "act" here is just prepping data to assert on it, I'd place that under "assert" as it's not executing the (relevant) code under test. That makes it more clear which operation we're testing.

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,585 @@
...
+    const expectedAttributeNamesAfterKobo = [
+      ...expectedAttributeNamesBeforeKobo,
+      'How_are_you_today_select_one',
+      'What_is_2_2_number',
+    ];
+
+    // Assert
+    expect(programAttributesBeforeKobo.sort()).toEqual(
+      expectedAttributeNamesBeforeKobo.sort(),
+    );
```
nielsbom: If we skip the "act" section this assert still passes, so what's the reason for it?

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,585 @@
...
+      accessToken,
+    });
+
+    const programId = createProgramResponse.body.id;
+    const getProgramResponseBeforeKoboIntegration = await getProgram(
+      programId,
+      accessToken,
+    );
+    const programBeforeKoboIntegration =
+      getProgramResponseBeforeKoboIntegration.body;
```
nielsbom:
```suggestion
    const programBeforeKoboIntegration = (
      await getProgram(programId, accessToken)
    ).body;
```

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,585 @@
...
+    ];
+
+    const expectedAttributeNamesAfterKobo = [
+      ...expectedAttributeNamesBeforeKobo,
+      'How_are_you_today_select_one',
+      'What_is_2_2_number',
+    ];
+
+    // Assert
+    expect(programAttributesBeforeKobo.sort()).toEqual(
```
nielsbom: Shall we use [this jest extension library](https://jest-extended.jestcommunity.dev/docs/matchers/array/#toincludesamemembersmembers) to compare arrays? I see quite a bit of test code where we do this and I think it's easier to read. Just in this file it should help already. And we don't ship test code.

services/121-service/src/kobo/services/kobo.validation.service.spec.ts
```diff
@@ -0,0 +1,738 @@
...
+        } catch (e) {
+          error = e;
+        }
+
+        // Assert
+        expect(error).toBeInstanceOf(HttpException);
+        expect(error.getStatus()).toBe(HttpStatus.BAD_REQUEST);
+        expect(error.message).toMatchInlineSnapshot(`
+         "Kobo form definition validation failed:
+         - Kobo form attribute "nationalId" has incompatible type for 121 attribute: expected: "text, select_multiple, select_multiple_from_file, rank, geopoint, geotrace, geoshape, date, time, dateTime, image, audio, background-audio, video, file, barcode, calculate, acknowledge, hidden, xml-external", got "integer"  "
```
nielsbom: This error makes me think `nationalId` can have any of the listed types, but it can only be "text" right? Shouldn't we respond with that?

services/121-service/src/kobo/services/kobo-survey-processor.service.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+    const module: TestingModule = await Test.createTestingModule({
+      providers: [KoboSurveyProcessorService],
+    }).compile();
+
+    service = module.get<KoboSurveyProcessorService>(
+      KoboSurveyProcessorService,
+    );
+  });
+
+  describe('surveyToProgramRegistrationAttributes', () => {
```
nielsbom: I'd add a group for testing multiple survey items and then also add a test that processes multiple items. And then just that it results in multiple items, not their contents.

Another group of tests can then focus on `surveyItemToProgramRegistrationAttribute` and then you don't have to repeat 
`expect(result).toHaveLength(1);` or `result[0]` 5 times.

services/121-service/src/kobo/interfaces/kobo-survey-item-cleaned.interface.ts
```diff
@@ -0,0 +1,7 @@
+export interface KoboSurveyItemCleaned {
+  name: string;
+  type: string;
+  label?: string[];
+  required?: boolean;
+  select_from_list_name?: string;
```
nielsbom: Maybe rename to `options_for_select`? I know this is the Kobo name, but I think `select_from_list_name` is a confusing name.

services/121-service/src/kobo/services/kobo-survey-processor.service.ts
```diff
@@ -0,0 +1,193 @@
...
+import { KoboSurveyItemCleaned } from '@121-service/src/kobo/interfaces/kobo-survey-item-cleaned.interface';
+import { ProgramRegistrationAttributeDto } from '@121-service/src/programs/dto/program-registration-attribute.dto';
+import { RegistrationAttributeTypes } from '@121-service/src/registration/enum/registration-attribute.enum';
+import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
+import { RegistrationPreferredLanguageTranslation } from '@121-service/src/shared/types/registration-preferred-language-translation.type';
+import { UILanguageTranslation } from '@121-service/src/shared/types/ui-language-translation.type';
+
+@Injectable()
+export class KoboSurveyProcessorService {
+  private static readonly DEFAULT_FALLBACK_LANGUAGE =
```
nielsbom: Why are we using `static` here? We're always instantiating this class (unlike mapper classes). Making these fields static makes us have to call these fields via the class name versus `this.`.

services/121-service/src/kobo/dtos/kobo-api/kobo-survey-item.dto.ts
```diff
@@ -0,0 +1,15 @@
+export interface KoboSurveyItemDto {
+  name?: string;
+  type: string;
```
nielsbom: Maybe helpful to add some tsdocs here too? Especially around the more complex properties like `type`, `label` or `select_from_list_name`.

services/121-service/src/kobo/consts/kobo-survey-to-121-attribute-type.const.ts
```diff
@@ -0,0 +1,30 @@
...
+  audio: RegistrationAttributeTypes.text,
+  'background-audio': RegistrationAttributeTypes.text,
+  video: RegistrationAttributeTypes.text,
+  file: RegistrationAttributeTypes.text,
+  barcode: RegistrationAttributeTypes.text,
+  calculate: RegistrationAttributeTypes.text,
+  acknowledge: RegistrationAttributeTypes.text,
+  hidden: RegistrationAttributeTypes.text,
+  'xml-external': RegistrationAttributeTypes.text,
+};
```
nielsbom: New ordering does require a few snapshot updates
```suggestion
const { numeric, text, dropdown } = RegistrationAttributeTypes;

type TypeMapping = Record<string, RegistrationAttributeTypes>;
export const KOBO_TO_121_TYPE_MAPPING: TypeMapping = {
  // numeric
  decimal: numeric,
  integer: numeric,
  range: numeric,
  // dropdown
  select_one: dropdown,
  select_one_from_file: dropdown,
  // text
  'background-audio': text,
  'xml-external': text,
  acknowledge: text,
  audio: text,
  barcode: text,
  calculate: text,
  date: text,
  dateTime: text,
  file: text,
  geopoint: text,
  geoshape: text,
  geotrace: text,
  hidden: text,
  image: text,
  rank: text,
  select_multiple_from_file: text,
  select_multiple: text,
  text,
  time: text,
  video: text,
};
```

services/121-service/src/fsp-management/fsp-attribute-type-mapping.ts
```diff
@@ -0,0 +1,21 @@
...
+  [FspAttributes.addressHouseNumber]: RegistrationAttributeTypes.numeric,
+  [FspAttributes.addressHouseNumberAddition]: RegistrationAttributeTypes.text,
+  [FspAttributes.addressPostalCode]: RegistrationAttributeTypes.text,
+  [FspAttributes.addressCity]: RegistrationAttributeTypes.text,
+  [FspAttributes.bankAccountNumber]: RegistrationAttributeTypes.text,
+  [FspAttributes.whatsappPhoneNumber]: RegistrationAttributeTypes.tel,
+  [FspAttributes.firstName]: RegistrationAttributeTypes.text,
+  [FspAttributes.lastName]: RegistrationAttributeTypes.text,
+  [FspAttributes.phoneNumberPayment]: RegistrationAttributeTypes.tel,
+};
```
nielsbom:
```suggestion
const { numeric, text, tel } = RegistrationAttributeTypes;

export const FINANCIAL_SERVICE_PROVIDER_ATTRIBUTE_TYPE_MAPPING: Record<
  FspAttributes,
  RegistrationAttributeTypes
> = {
  [FspAttributes.addressCity]: text,
  [FspAttributes.addressHouseNumber]: numeric,
  [FspAttributes.addressHouseNumberAddition]: text,
  [FspAttributes.addressPostalCode]: text,
  [FspAttributes.addressStreet]: text,
  [FspAttributes.bankAccountNumber]: text,
  [FspAttributes.firstName]: text,
  [FspAttributes.fullName]: text,
  [FspAttributes.lastName]: text,
  [FspAttributes.nationalId]: text,
  [FspAttributes.phoneNumber]: tel,
  [FspAttributes.phoneNumberPayment]: tel,
  [FspAttributes.whatsappPhoneNumber]: tel,
};

```

services/121-service/src/kobo/kobo.controller.ts
```diff
@@ -0,0 +1,145 @@
...
+  @ApiResponse({
+    status: HttpStatus.OK,
+    description:
+      'Dry run completed successfully - the integration would succeed if executed',
+    type: KoboIntegrationResultDto,
+  })
+  @ApiResponse({
+    status: HttpStatus.NOT_FOUND,
+    description:
+      'Program does not exist, or Kobo form could not be found at the specified URL/asset ID',
```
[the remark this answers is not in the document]
RubenGeo: I think it's too much effort for what value it brings. We would need a central location to store all error states to do this properly:

Something like create a seperate file with all possible error states, maybe an enum that lists them and than use those across the controller and the places where I throw errors.

services/121-service/src/kobo/kobo.controller.ts
RubenGeo: https://github.com/global-121/121-platform/pull/7703#discussion_r2687232158 same comment as here

services/121-service/src/kobo/services/kobo-api.service.ts
```diff
@@ -0,0 +1,73 @@
...
+    assetId: string;
+    token: string;
+    baseUrl: string;
+  }): Promise<KoboFormDefinition> {
+    const apiUrl = `${baseUrl}/api/v2/assets/${assetId}/deployment/?format=json`;
+    const headers = new Headers();
+    headers.append('Authorization', `Token ${token}`);
+
+    const response = await this.httpService.get<
+      AxiosResponse<KoboAssetResponseDto>
```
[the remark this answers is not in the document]
RubenGeo: I think checking if every property exists is a bit overengineering for this stage. A pattern I can see is class validator or zod and than throw an internal server error. So we can take actions if the kobo api is lying to us.

services/121-service/src/kobo/services/kobo.service.ts
RubenGeo: Hmm I think instead I will call the extract language code earlier before the validation. Than I will throw a specific error in the validation service if one of the languages in the asset.summary.languages property from kobo does not comply with an iso code. And than we can see if Sanne or Garbriele comes back to us with complaints. Which is a bit stricter of a stricter approach so we can learn more about what kobo actually sends

services/121-service/src/programs/programs.service.ts
```diff
@@ -340,6 +340,37 @@ export class ProgramService {
...
   }
 
+  public async upsertProgramRegistrationAttributes({
+    programId,
+    programRegistrationAttributes,
+  }: {
+    programId: number;
+    programRegistrationAttributes: ProgramRegistrationAttributeDto[];
+  }): Promise<void> {
+    for (const attribute of programRegistrationAttributes) {
```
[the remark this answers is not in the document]
RubenGeo: I save all program registrations attributes now in one save called directly on the repo and am not calling the private methods updateProgramRegistrationAttribute and createProgramRegistrationAttributeEntity anymore. 

These methods mostly contained unneeded validation which we now skip :)

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,585 @@
...
+    expect(getKoboResponse.status).toBe(HttpStatus.OK);
+    expect(getKoboResponse.body).toHaveProperty('assetId', koboLinkDto.assetId);
+    expect(getKoboResponse.body).toHaveProperty('versionId');
+
+    // Verify program registration attributes were updated based on Kobo form definition
+    expect(programAttributeNamesAfterKobo.sort()).toEqual(
+      expectedAttributeNamesAfterKobo.sort(),
+    );
+
+    // Test common properties that should be the same for all attributes
```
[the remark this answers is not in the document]
RubenGeo: I reduced it to just validating all names and just validate the first program attribute completely. To see if the whole flow works between all our parts. Let's discuss IRL if you feel strong about a different approach

services/121-service/src/kobo/services/kobo-survey-processor.service.ts
```diff
@@ -0,0 +1,193 @@
...
+      string,
+      { option: string; label: RegistrationPreferredLanguageTranslation }[]
+    >;
+    languageIsoCodes: RegistrationPreferredLanguage[];
+  }): ProgramRegistrationAttributeDto | undefined {
+    const primaryKoboType =
+      koboSurveyItem.type.split(' ')[
+        KoboSurveyProcessorService.KOBO_TYPE_SEPARATOR_INDEX
+      ];
+    const attributeType = KOBO_TO_121_TYPE_MAPPING[primaryKoboType];
```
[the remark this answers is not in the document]
RubenGeo: Added tests for this

services/121-service/src/kobo/services/kobo.service.ts
nielsbom: Agreed

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -0,0 +1,391 @@
...
+    }
+  }
+
+  private getKoboTypesFrom121Type(type: string): string[] {
+    return Object.entries(KOBO_TO_121_TYPE_MAPPING)
+      .filter(([_, value]) => value === type)
+      .map(([key]) => key);
+  }
+
+  private collectErrors({
```
nielsbom: > This is a very good idea thanks for the suggestion! I thought about it for a bit and would like to go for this approach:
..
> because I think we mostly program our backend in a functional approach instead of an OOP approach and I think this fits that approach a bit better, while retaining most of the advantages of your suggestion

Excellent solution. Personally I'd also move the function out of the class but keep it in the module (so not as an export).

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -0,0 +1,418 @@
...
+    const koboVersionIdAfterSecondImport = updatedKobo.body.versionId;
+    expect(koboVersionIdAfterSecondImport).toEqual(
+      'asset-id-happy-flow-with-changes', // version id from the second kobo definition
+    );
+
+    const getProgramResponseAfterUpdate = await getProgram(
+      programId,
+      accessToken,
+    );
+    const programAfterUpdate = getProgramResponseAfterUpdate.body;
```
nielsbom:
```suggestion
    const programAfterUpdate = (await getProgram(programId, accessToken)).body;
```

services/121-service/test/kobo/kobo-load-form-definition.test.ts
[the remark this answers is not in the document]
nielsbom: This was done.

services/121-service/src/kobo/services/kobo.validation.service.spec.ts
nielsbom: Created a new comment on this.

PR #7756, opened 2026-01-20
Chore: get distribution-by-mail-config for non admins

Yuri-510:
[AB#40014](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/40014) 

## Describe your changes

Added endpoint for - and created public FSP config props, so that non-admin users can access these.

- [x] I have updated all [documentation](https://github.com/global-121/121-internal-documentation/blob/main/documentation.md) where necessary

## Portal preview-deployment

https://happy-rock-0411d2003-7756.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/test/program/manage-program-fsp-configuration.test.ts
+++ b/services/121-service/test/program/manage-program-fsp-configuration.test.ts
@@ -450,10 +450,10 @@ describe('Manage Fsp configurations', () => {
     });
   });
 
-  it('Should return allowlisted public properties of a program Fsp configuration for non-admin users with registration.read permission', async () => {
+  it('Should return allowlisted public properties of a program Fsp configuration for users with program.read permission', async () => {
     // Arrange
-    const nonAdminAccessToken = await createAccessTokenWithPermissions({
-      permissions: [PermissionEnum.RegistrationREAD],
+    const programReadAccessToken = await createAccessTokenWithPermissions({
+      permissions: [PermissionEnum.ProgramREAD],
       adminAccessToken: accessToken,
       programId: programIdVisa,
     });
@@ -462,7 +462,7 @@ describe('Manage Fsp configurations', () => {
     const result = await getPublicProgramFspConfigurationProperties({
       programId: programIdVisa,
       configName: Fsps.intersolveVisa,
-      accessToken: nonAdminAccessToken,
+      accessToken: programReadAccessToken,
     });
 
     // Assert


services/121-service/src/program-fsp-configurations/program-fsp-configurations.controller.ts
```diff
@@ -197,6 +198,41 @@ export class ProgramFspConfigurationsController {
     );
   }
 
+  @AuthenticatedUser({ permissions: [PermissionEnum.RegistrationREAD] })
+  @ApiOperation({
+    summary:
+      'Retrieve allowlisted public properties for Fsp Configuration. Only returns properties that are safe to expose to non-admin users based on the Fsp type.',
```
RubenGeo:
```suggestion
    summary:
      'Retrieve allowlisted public properties for Fsp Configuration. Only returns properties that are safe to expose to non-admin users based on the Program Fsp Configuration Property type.',
```

services/121-service/test/program/manage-program-fsp-configuration.test.ts
```diff
@@ -443,4 +446,40 @@ describe('Manage Fsp configurations', () => {
...
+    const result = await getPublicProgramFspConfigurationProperties({
+      programId: programIdVisa,
+      configName: Fsps.intersolveVisa,
+      accessToken: nonAdminAccessToken,
+    });
+
+    // Assert
+    expect(result.statusCode).toBe(HttpStatus.OK);
+    expect(result.body).toHaveLength(1);
+    expect(result.body[0]).toEqual(
```
RubenGeo: I think it's a bit better to validate here doing something like 
```
const returnedPropertyNames = result.body.map((p) => p.name);
const allowlistedPropertyNames =
      PublicFspConfigurationProperties[config.fspName];

expect((returnedPropertyNames.sort()).toEqual(allowlistedPropertyNames.sort())

```


To make the test a little less fragile for when we are adding an extra property (like you are doing now) and than maybe dynamically get all `nonAllowelistedProperties` and validate those

services/121-service/src/program-fsp-configurations/program-fsp-configurations.service.spec.ts
```diff
@@ -120,6 +120,69 @@ describe('ProgramFspConfigurationsService', () => {
...
+      );
+
+      const result = await service.getPublicFspConfigurationProperties(
+        programId,
+        configName,
+      );
+
+      expect(
+        mockProgramFspConfigurationPropertyRepository.find,
+      ).not.toHaveBeenCalled();
```
RubenGeo:
```suggestion
```

I think we do not have to check the internal working of the service in this way. Checking only the end result makes the test less fragile

PR #7836, opened 2026-02-12
Add E2E tests for kobo integration

Piotrk39:
[AB#39715](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/39715) 

## Describe your changes

Adds E2E tests for Kobo Integration:

1. Add "Kobo toolbox" registrations integration successfully and validate it is present in the interface
2. Failed "Kobo toolbox" registration integration (error message)

## Portal preview-deployment


services/121-service/src/env.ts
```diff
@@ -325,6 +325,10 @@ export const env = createEnv({
     KOBO_CONNECT_API_URL: z
       .url()
       .pipe(z.transform((url) => withoutTrailingSlash(url))),
+    KOBO_SERVER_URL: z
```
nielsbom: Integrating with a Kobo form is done by the user inputting a server URL, asset ID and API key into a form, this will be saved in the database and used when (re-)requesting the Koboform to update the project. So this explicitly should not be in the environment variables of the 121 service as it can also be different per project.

For running E2E tests against the mock service I suggest just setting those values in `AddKoboIntegration.spec`.

services/.env.example
```diff
@@ -290,6 +290,8 @@ TWILIO_MESSAGING_SID=MG
 # -------------------------
 # (Required) Converts Kobo/ODK-forms to a format that can be used by the platform.
 KOBO_CONNECT_API_URL=https://kobo-connect.example.org
+KOBO_SERVER_URL=http://mock-service:3001/api/kobo
```
nielsbom: Also not needed, see other comment.

PR #7837, opened 2026-02-12
poc: shared files

Yuri-510:
[AB#40276](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/40276) 
This is a small proof of concept of having a shared package that would house shared enums, interfaces, dto's etc.
It creates this package and moves one shared enum "UILanguages" to that package and installs that package using "npm link".

More candidates (shared files) for moving to the shared packages are:

**enums**
-activity-type
-export-type
-fsp-integration-type
-fsp-name
-registration-preferred-language
-registration-status
-ui-language
-user-type

**interfaces**
-activity
-paginated-response
-user

**types**
-http-method
-language
-ui-language-translation

- [ ] I have updated all [documentation](https://github.com/global-121/121-internal-documentation/blob/main/documentation.md) where necessary
- [ ] I have checked [the list of integrations](https://github.com/global-121/121-platform/wiki/Integrations-with-external-systems#api-endpoints-used-by-external-systems) with the 121 API for [changed endpoints](https://github.com/global-121/121-platform/blob/main/services/121-service/README.md#api)

## Portal preview-deployment

https://happy-rock-0411d2003-7837.westeurope.3.azurestaticapps.net

### DIFF
--- a/interfaces/portal/package-lock.json
+++ b/interfaces/portal/package-lock.json
@@ -8,6 +8,7 @@
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
+        "@121-platform/shared": "file:../../packages/shared",
         "@angular/animations": "^20.3.15",
         "@angular/common": "^20.3.15",
         "@angular/compiler": "^20.3.15",
@@ -88,6 +89,20 @@
         "node": ">=24",
         "npm": ">=11"
       }
+    },
+    "../../packages/shared": {
+      "name": "@121-platform/shared",
+      "version": "0.0.0",
+      "dependencies": {
+        "prettier": "^3.8.1"
+      },
+      "devDependencies": {
+        "typescript": "^5.9.3"
+      }
+    },
+    "node_modules/@121-platform/shared": {
+      "resolved": "../../packages/shared",
+      "link": true
     },
     "node_modules/@algolia/abtesting": {
       "version": "1.1.0",
--- a/services/121-service/package-lock.json
+++ b/services/121-service/package-lock.json
@@ -7,6 +7,7 @@
       "name": "121-service",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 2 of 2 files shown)


packages/shared/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "description": "Minimal shared types for 121-platform (PoC)",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.js",
+      "import": "./dist/index.js"
+    }
+  },
```
elwinschmitz: Would there be a way to only distribute the source-code?
So that there's no real notion of these generated files?

Each consuming app would generate these `*.js` files anyway when it needs it, I'd guess..

package-lock.json
```diff
@@ -1769,7 +1769,6 @@
       "integrity": "sha512-UOnG6LftzbdaHZcKoPFtOcCKztrQ57WkHDeRD9t/PTQtmT0NHSeWWepj6pS0z/N7+08BHFDQVUrfmfMRcZwbMg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
```
elwinschmitz: ¯\_(ツ)_/¯ 

I don't get all these differences everywhere...
Probably something to do with slightly different `npm` versions or settings..

elwinschmitz (commented):
This looks promising! 🥳 

I haven't 'tried it out' yet, to see what the "autocomplete"/"auto import this dependency"-flow would be in VSCode.. (when used from the different folders)

If we can get rid of some of the 'noisy' TS/npm-config files/changes, it would be nice.
But I think that is mainly only for this one-time-setup.. Not for every added use-case...

PR #8215, opened 2026-05-06
feat: Manage approvers eligible approvers API

nameerhqureshi:
[AB#42096](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42096) 

## Describe your changes

- Added a separate APIs to get the eligible approvals

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/user/dto/get-user-response.dto.ts
+++ b/services/121-service/src/user/dto/get-user-response.dto.ts
@@ -19,6 +19,9 @@ export class GetUserReponseDto {
   @ApiProperty()
   scope: string;
 
+  @ApiProperty({ example: true })
+  isEligiblePaymentApprover: boolean;
+
   @ApiProperty()
   roles: {
     id: number;
--- a/services/121-service/src/user/user.service.spec.ts
+++ b/services/121-service/src/user/user.service.spec.ts
@@ -19,6 +19,7 @@ describe('UserService', () => {
   let userRepository: Repository<UserEntity>;
   let permissionRepository: Repository<PermissionEntity>;
   let userRoleRepository: Repository<UserRoleEntity>;
+  let assignmentRepository: ProgramAidworkerAssignmentRepository;
   let userEmailsService: UserEmailsService;
 
   const mockRequest = {
@@ -50,6 +51,7 @@ describe('UserService', () => {
           useValue: {
             isApprover: jest.fn(),
             findByUserId: jest.fn(),
+            createQueryBuilder: jest.fn(),
           },
         },
         {
@@ -75,6 +77,9 @@ describe('UserService', () => {
     );
     userRoleRepository = module.get<Repository<UserRoleEntity>>(
       getRepositoryToken(UserRoleEntity),
+    );
+    assignmentRepository = module.get<ProgramAidworkerAssignmentRepository>(
+      ProgramAidworkerAssignmentRepository,
     );
     userEmailsService = module.get<UserEmailsService>(UserEmailsService);
   });
@@ -936,6 +941,77 @@ describe('UserService', () => {
 
       expect(userRepository.save).toHaveBeenCalled();
       expect(userEmailsService.sendPasswordReset).toHaveBeenCalled();
+    });
+  });
+
+  describe('getUsersInProgram', () => {
+    function createRawUser({
+      scope,
+    }: {
+      scope: string;
+    }): {
+      id: number;
+      username: string;
+      admin: boolean;
+      active: boolean;
+      lastLogin: Date | null;
+      rolesid: number[];
+      role: string[];
+      label: string[];
+      scope: string;
+    } {
+      return {
+        id: 1,
+        username: 'user@example.org',
+        admin: false,
+        active: true,
+        lastLogin: null,
+        rolesid: [11],
+        role: ['Program Manager'],
+        label: ['Program Manager'],
+        scope,
+      };
+    }
+
+    function mockQueryBuilder(
+      rawUsers: ReturnType<typeof createRawUser>[],
+    ): void {
+      const queryBuilderMock = {
+        leftJoinAndSelect: jest.fn().mockReturnThis(),
+        where: jest.fn().mockReturnThis(),
+        andWhere: jest.fn().mockReturnThis(),
+        select: jest.fn().mockReturnThis(),
+        groupBy: jest.fn().mockReturnThis(),
+        getRawMany: jest.fn().mockResolvedValue(rawUsers),
+      };
+
+      jest
+        .spyOn(assignmentRepository, 'createQueryBuilder')
+        .mockReturnValue(queryBuilderMock as never);
+
+      return;
+    }
+
+    it('should set isEligiblePaymentApprover=true for unscoped assignments', async () => {
+      // Arrange
+      mockQueryBuilder([createRawUser({ scope: '' })]);
+
+      // Act
+      const result = await service.getUsersInProgram(1);
+
+      // Assert
+      expect(result[0].isEligiblePaymentApprover).toBe(true);
+    });
+
+    it('should set isEligiblePaymentApprover=false for scoped assignments', async () => {
+      // Arrange
+      mockQueryBuilder([createRawUser({ scope: 'region-1' })]);
+
+      // Act
+      const result = await service.getUsersInProgram(1);
+
+      // Assert
+      expect(result[0].isEligiblePaymentApprover).toBe(false);
     });
   });
 });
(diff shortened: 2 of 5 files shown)


services/121-service/src/user/controllers/user-assignments.controller.ts
```diff
@@ -177,4 +178,28 @@ export class UserAssignmentsController {
...
+    description:
+      'Returns a list of users who can approve transactions for the program, excluding the currently logged-in user to prevent self-approval.',
+  })
+  @ApiParam({ name: 'programId', required: true, type: 'integer' })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    description: 'Retrieved eligible approvers successfully',
+    type: [GetUserReponseDto],
+  })
+  @Get('programs/:programId/eligible-approver-users')
```
RubenGeo: Hey! 

Functionally this PR makes a lot of sense and the testing coverage is great! However I think it breaks a bit with the design pattern we want to follow for the 121-platform with regards to APIs.

I was thinking a better design pattern could be to that instead of introducing a dedicated GET endpoint what about extending the existing GET /programs/:programId/users response with an isEligibleApprover: boolean flag. The business logic rules of scope === '' can than live in there. In the Portal, filter out the current user in the approver-picker component.

This is because we try to organize APIs around recourses/entities and not use-cases, see our software architecture [guidelines H1](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#h-api-design).

These principles are inspired by this [YouTube](https://www.youtube.com/watch?app=desktop&v=P0a7PwRNLVU)) video, which is a nice watch I think. We don't follow all principles of this video. And also our older code does not follow these principles as it was written before we had the architecture guidelines.

 - In this case I think we do not want to bias toward one use case in the frontend. So this endpoint `programs/:programId/eligible-approver-users` is really tailored to the component you are building now.  A more neutral list of program users with an eligibility flag i think will be more useful for future components, script, or integrations by cash-im./
- I would leave the current user filtering to the FE because I think the responses of a GET endpoint should be caller-agnostic. E.g. if Niels is assigned to a program he is conceptually an eligible approver, even when Niels himself is the one calling the endpoint. Excluding him from the response would make the same URL return different results depending on who's asking, which couples the response to one specific UI need. The frontend already knows who the current user is and can filter them out in the approver-picker component, which is more of a UX choice to help the user

Curious what you think!

RubenGeo (changes requested):

services/121-service/src/user/controllers/user-assignments.controller.ts
nameerhqureshi: Hey,

I think that makes sense and I was thinking of this very initially. I agree the current dedicated endpoint is more use-case specific than resource oriented. I somehow agree the current-user exclusion can handled in the Portal as a UX concern as BE is handling this check on approvers update call, so it would be still safe.

I’ll refactor GET `/programs/:programId/users` to include an eligibility flag (`canApprovePayments`) and update the Portal to filter out the logged-in user locally.

services/121-service/src/user/dto/get-user-response.dto.ts
```diff
@@ -19,6 +19,9 @@ export class GetUserReponseDto {
   @ApiProperty()
   scope: string;
 
+  @ApiProperty({ example: true })
+  canApprovePayments: boolean;
```
RubenGeo: I think I would prefer a name along the lines of `isEligibleApprover` or `canBeAssignedAsApprover`. Because I think `canApprovePayments` implies that the user already has the approve ability

services/121-service/src/user/user.service.spec.ts
```diff
@@ -938,4 +943,121 @@ describe('UserService', () => {
...
+            {
+              id: 12,
+              role: 'Field Officer',
+              label: 'Field Officer',
+            },
+          ],
+          scope: 'region-1',
+          canApprovePayments: false,
+        },
+      ]);
```
RubenGeo: I think you could go a to a more focused test approach: 

```
 describe('getUsersInProgram', () => {
    function createRawUser({ scope }: { scope: string }) {
      return {
        id: 1,
        username: 'user@example.org',
        admin: false,
        active: true,
        lastLogin: null,
        rolesid: [11],
        role: ['Program Manager'],
        label: ['Program Manager'],
        scope,
      };
    }

    function mockQueryBuilder(rawUsers: ReturnType<typeof createRawUser>[]) {
      const mock = {
        leftJoinAndSelect: jest.fn().mockReturnThis(),
        where: jest.fn().mockReturnThis(),
        andWhere: jest.fn().mockReturnThis(),
        select: jest.fn().mockReturnThis(),
        groupBy: jest.fn().mockReturnThis(),
        getRawMany: jest.fn().mockResolvedValue(rawUsers),
      };
      jest
        .spyOn(assignmentRepository, 'createQueryBuilder')
        .mockReturnValue(mock as never);
    }

    it('should set canApprovePayments=true for unscoped assignments', async () => {
      mockQueryBuilder([createRawUser({ scope: '' })]);

      const result = await service.getUsersInProgram(1);

      expect(result[0].canApprovePayments).toBe(true);
    });

    it('should set canApprovePayments=false for scoped assignments', async () => {
      mockQueryBuilder([createRawUser({ scope: 'region-1' })]);

      const result = await service.getUsersInProgram(1);

      expect(result[0].canApprovePayments).toBe(false);
    });

```

RubenGeo: Another idea to make the mocking in the unit test less complex could be to move the query builder code to the ProgramAidworkerAssignmentRepository and create function called `getUsersInprogram` in there. 

```
// Query builder code
await this.assignmentRepository
      .createQueryBuilder('assignment')
      .leftJoinAndSelect('assignment.roles', 'roles')
      .leftJoinAndSelect('assignment.user', 'user')
      .where('assignment.programId = :programId', { programId })
      .andWhere('user.userType = :userType', { userType: UserType.aidWorker })
      .select([
        'user.id AS id',
        'user.username AS username',
        'user.admin AS admin',
        'user.active AS active',
        'user.lastLogin AS "lastLogin"',
        'ARRAY_AGG(roles.id) AS rolesId',
        'ARRAY_AGG(roles.role) AS role',
        'ARRAY_AGG(roles.label) AS label',
        'MAX(assignment.scope) AS scope',
      ])
      .groupBy('user.id')
      .getRawMany();

```

The query builder logic in getUsersInProgram lives in the service, but per our architecture [guidelines](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture) ( J.4-5), all database interactions should be in repositories.

Moving the query to ProgramAidworkerAssignmentRepository would:
- Align with our guidelines
- Simplify these tests significantly you could only mock one repository method return value instead of the entire query builder chain

I know this code is old and predates the guidelines, so I leave it up to you to decide whether to include it in this PR or leave it out of scope for this PR

RubenGeo (changes requested):

services/121-service/src/user/dto/get-user-response.dto.ts
nameerhqureshi: Yes we can change. I was thinking to include 'payment' in the name since this is specifically for Payment Approvers (unless we plan to use this flag for other approval processes too). Would `isEligiblePaymentApprover` or `canBePaymentApprover` be a better fit? Let me know whats your suggestion, please.

services/121-service/src/user/user.service.spec.ts
nameerhqureshi: - I will incorporate your suggestion for the unit test above. (After the variable name recommendation confirmation in previous comment)

- Yes this is a good suggestion and definitely the right way as we discussed before. To keep the scope of this PR manageable and avoid blocking the current changes, I will suggest to leave it out of scope for now and create a separate ticket to move this query builder to the repository.
I can make a reminder ticket to create this task or even I can work on it separately after this task.

services/121-service/src/user/dto/get-user-response.dto.ts
RubenGeo: image
Good suggestion both are clear to me :). So pick which you like more

services/121-service/src/user/user.service.spec.ts
RubenGeo: > I will suggest to leave it out of scope for now and create a separate ticket to move this query builder to the repository.

That's fine

> I can make a reminder ticket to create this task or even I can work on it separately after this task.

We can just leave it to our future selves whenever we touch this code again. I don't think it's important enough to track it in a seperate ticket

RubenGeo (approved):
Love it!

PR #8216, opened 2026-05-06
add koboLabel + show labels in settings page

arsforza:
[AB#42113](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42113)

## Portal preview-deployment

https://happy-rock-0411d2003-8216.westeurope.3.azurestaticapps.net


services/121-service/src/programs/entities/program-registration-attribute.entity.ts
```diff
@@ -24,6 +24,9 @@ export class ProgramRegistrationAttributeEntity extends Base121Entity {
   @Column('json')
   public label: RegistrationPreferredLanguageTranslation;
 
+  @Column('json', { nullable: true })
+  public customLabel: RegistrationPreferredLanguageTranslation | null;
+
```
RubenGeo, line 29: Hey I would have the following 2 labels: 

```
@Column('json', { nullable: true })
public label: RegistrationPreferredLanguageTranslation | null;  // user-defined, priority

@Column('json', { nullable: true })
public koboLabel: RegistrationPreferredLanguageTranslation | null; 
```

I get where your frustration with comes from with unclear refinement text, when I reread the refinement text of this item.

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -170,6 +170,7 @@ describe('Import a Kobo form definition', () => {
 
     expect(firstAttributeSnapshot).toMatchInlineSnapshot(`
      {
```
RubenGeo, line 172: I would also include the following changes in this PR:

1.  Fill the koboLabel when integrating with Kobo. 
  a. See surveyItemToProgramRegistrationAttribute in services/121-service/src/kobo/services/kobo-survey-processor.service.ts and the private functions it calls
  b. Ensure it later get's stored in upsertProgramRegistrationAttributes 
2. In the FE in places like the registration profile page and the registration table ensure that we use koboLabel if 'label' does not exist (for that specific languages I guess)
3. Change the unit tests and this test (services/121-service/test/kobo/kobo-load-form-definition.test.ts) should now fill koboLabel instead of label
4. Some tests updates

Unfortunately a couple of FE and BE changes need to be in 1 PR. If we only do the BE changes in 1 PR we risk that we release this and that Gabrielle than links a program to kobo and that all the labels are stored in koboLabel and never show in the FE

RubenGeo: All the other stuff 

Image does not have to be part of this PR

RubenGeo (changes requested):

services/121-service/test/kobo/kobo-load-form-definition.test.ts
arsforza: resolving this because there's new code

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,152 @@
...
+    computed(() => {
+      const { label, koboLabel, name } = attribute;
+
+      console.log('\n\n🚀 ~');
+      console.log('🚀 ~ RegistrationQuestionsCardComponent ~ label:', label);
+      console.log(
+        '🚀 ~ RegistrationQuestionsCardComponent ~ koboLabel:',
+        koboLabel,
+      );
+      console.log('🚀 ~ RegistrationQuestionsCardComponent ~ name:', name);
```
RubenGeo: There are some console.logs still here :)

services/121-service/src/migration/1778621834090-set-attribute-labels-nullable.ts
```diff
@@ -0,0 +1,23 @@
...
+export class SetAttributeLabelsNullable1778621834090 implements MigrationInterface {
+  name = 'SetAttributeLabelsNullable1778621834090';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."program_registration_attribute" ALTER COLUMN "koboLabel" DROP NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "121-service"."program_registration_attribute" ALTER COLUMN "label" DROP NOT NULL`,
+    );
```
RubenGeo: It's better to have one migration file, because TypeORM wraps each migration in its own transaction. If the first migration runs successfully (column added) but the second one fails (e.g. due to a deploy issue), the database is left in a partially migrated state.

FYI: A database transaction is a bit like a git commit, so all queries are sort of put ready and than when the transaction is closed they are committed to the DB.

You can merge both files by just copy pasting the queries from the file to services/121-service/src/migration/1778059063560-registration-question-kobo-label.ts (and than locally you probably have to rerun `npm run start:service:detach` and or recreate your 121-service db schema)

services/121-service/src/programs/dto/program-registration-attribute.dto.ts
```diff
@@ -80,8 +80,15 @@ export class ProgramRegistrationAttributeDto extends BaseProgramRegistrationAttr
       fr: "Remplissez votre nom, s'il vous plaît:",
     },
   })
-  @IsNotEmpty()
-  public readonly label: RegistrationPreferredLanguageTranslation;
+  public readonly label?: RegistrationPreferredLanguageTranslation | null;
```
RubenGeo, line 83: So I think it makes sense if the `koboLabel` can only be set in the flow that sets up the kobo integration and it cannot/ should not be change by people using our API.

However adding it here to this dto also adds it to the dto of creating a program `services/121-service/src/programs/dto/create-program.dto.ts`. Which means someone could also set the koboLabel manually when creating a program via our API.

Reading this code made me realize that we use ProgramRegistrationAttributeDto for a couple of purposes now. Which makes it really complex

1. API request body POST (when creating a program, which cause the error we debugged together)
2. API response Body program-return.dto.ts
3. Internally in the 121-service we use it as an interface to pass it around in several places (for example in applyProgramRegistrationAttributesFallbackIfNecessary)

According to our [guidelines](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#e-121-service-api-request-and-response-bodies) i think we should probably refactor this DTO in 3 separate dtos/interfaces

1. CreateProgramRegistrationAttributeDto -> what the API accepts on POST. Excludes koboLabel (and anything else only the system should set).
2. ProgramRegistrationAttributeResponseDto -> what the API returns on GET. Includes koboLabel.
3. ProgramRegistrationAttributeInterface -> used internally in the 121 service when passing around the registration attribute data between functions. 

This way I think we can make the code a bit more readable (hopefully), though it will be a bit more verbose.

I do think however it may be better to do the changes I propose in a seperate PR. And we can leave this as it is for now. What do you think @arsforza, does it make sense to you?

services/121-service/src/programs/entities/program-registration-attribute.entity.ts
```diff
@@ -21,8 +21,11 @@ export class ProgramRegistrationAttributeEntity extends Base121Entity {
   @Column()
   public name: string;
 
-  @Column('json')
-  public label: RegistrationPreferredLanguageTranslation;
+  @Column('json', { nullable: true })
+  public label?: RegistrationPreferredLanguageTranslation | null; // user-defined, priority
+
+  @Column('json', { nullable: true })
+  public koboLabel?: RegistrationPreferredLanguageTranslation | null;
```
RubenGeo:
```suggestion
  @Column('json', { nullable: true })
  public label: RegistrationPreferredLanguageTranslation | null; // user-defined, priority

  @Column('json', { nullable: true })
  public koboLabel: RegistrationPreferredLanguageTranslation | null;
```


If you change this you'll get a type-error in the service you can fix by doing this:

Image

services/121-service/test/metrics/get-native-dashboard-data.test.ts
```diff
@@ -83,10 +83,6 @@ describe('Get all payments aggregates', () => {
       const today = new Date().toISOString().split('T')[0];
       expect(aggregate.id).toEqual(2);
       expect(aggregate.date).toContain(today);
-      console.log(
-        'aggregate.aggregatedStatuses: ',
-        aggregate.aggregatedStatuses,
-      );
```
RubenGeo, line 89: nice 🥇

services/121-service/test/program/__snapshots__/create-program.test.ts.snap
```diff
@@ -1,988 +0,0 @@
-// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
```
RubenGeo, line 1: nice 🥇

interfaces/portal/src/locale/messages.xlf
```diff
@@ -2331,6 +2331,21 @@
       <trans-unit id="submission-result-skipped" datatype="html">
         <source>Submissions skipped</source>
       </trans-unit>
+      <trans-unit id="3860042976960140382" datatype="html">
+        <source>Below are the <x equiv-text="programAttributesLength.toString()" id="PH"/> questions from the linked Kobo form. Please edit each label to a shorter version which will be displayed in the table of the registration page and on the profile page. The default lanaguage is set by the language of the instance.</source>
```
RubenGeo: `questions from the linked Kobo form.`

These question could also have created not via kobo but via the API. 

So for example locally I did not integrate my kobo form yet and I still see this text.


Image

Maybe we should discus this with tal

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,152 @@
...
+    });
+
+  readonly sortedAttributes = computed(() => {
+    if (!this.programAttributes.isSuccess()) {
+      return [];
+    }
+
+    return this.programAttributes
+      .data()
+      .sort((a, b) => a.name.localeCompare(b.name));
```
RubenGeo: I was thinking maybe we should do this sorting on the BE. 

So than anywhere in the portal we use it it's sorted. 

I think for the registration profile page we sort attributes by created date (paTableAttributes). So maybe it makes sense to also sort here by created date instead of name, so the attributes on this page are the same as on the registration profile page

interfaces/portal/src/app/domains/program/program.api.service.ts
```diff
@@ -261,24 +274,45 @@ export class ProgramApiService extends DomainApiService {
...
+              ...attribute,
+              label: this.getAttributeTranslatedLabelOrName(attribute),
+            };
+          },
+        );
+      },
     });
   }
 
+  private getAttributeTranslatedLabelOrName(attribute: Attribute): string {
```
RubenGeo: Nice 👍

interfaces/portal/src/app/domains/program/program.api.service.ts
```diff
@@ -240,6 +240,19 @@ export class ProgramApiService extends DomainApiService {
     });
   }
 
+  getProgramRegistrationAttributesWithUntranslatedLabels({
```
RubenGeo, line 245: I think this is a good approach having this seperate function here

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,152 @@
...
+    enabled: !!this.programId(),
+  }));
+
+  readonly cardSubtitle = computed(() => {
+    if (!this.programAttributes.isSuccess()) {
+      return ``;
+    }
+
+    const programAttributesLength = this.programAttributes.data().length;
+    return $localize`Below are the ${programAttributesLength.toString()} questions from the linked Kobo form. Please edit each label to a shorter version which will be displayed in the table of the registration page and on the profile page.\n\nThe default lanaguage is set by the language of the instance.`;
```
RubenGeo: lanaguage > language

RubenGeo (changes requested):
nice work!

I am not good at giving critical feedback at FE code. I looks good to me :)

services/121-service/src/programs/dto/program-registration-attribute.dto.ts
arsforza: Copying this to a new task since the changes are going to be made in another PR

interfaces/portal/src/app/components/program-language-tabs/program-language-tabs.component.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  readonly getTabLabel = (language: Language) =>
+    computed(() => {
+      let languageToShowNameIn = UILanguage.en;
+      if (this.defaultLocale) {
+        languageToShowNameIn = this.defaultLocale as UILanguage;
+      }
+
+      return getLinguonym({
+        languageToDisplayNameOf: language,
```
RubenGeo:
```
readonly tabLabels = computed(() => {
  const labels = new Map<RegistrationPreferredLanguage, string>();
  let languageToShowNameIn = UILanguage.en;
  if (this.defaultLocale) {
    languageToShowNameIn = this.defaultLocale as UILanguage;
  }
  for (const language of this.programLanguages()) {
    labels.set(
      language,
      getLinguonym({ languageToDisplayNameOf: language, languageToShowNameIn }),
    );
  }
  return labels;
});
```
We could use something like this and than in the html use `{{ tabLabels().get(language) | titlecase }}`

Else I think a computed value is created for every time you change tabs. While if you create a map of the tabLabels is only recomputes the map if the prefered languages change. Not also when changing tabs.

So I think it's making better use of the computed() functionality. 

I think performance wise it's a neligeble way but I do feel it uses computed more correctly, which makes the code more readable.

interfaces/portal/src/app/components/program-language-tabs/program-language-tabs.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  readonly programLanguages = computed(() => {
+    if (!this.program.isSuccess()) {
+      return [];
+    }
+
+    const languages = this.program.data().languages;
+
+    // put the preferred language first
+    const reorderedLanguages: RegistrationPreferredLanguage[] = [];
+    const preferredLanguage = languages.find((l) => l === this.defaultLocale);
```
RubenGeo: Let's change this so we by default select the lanaguage of the Portal user instead of changing the order

interfaces/portal/src/app/domains/program/program.api.service.ts
```diff
@@ -263,24 +276,45 @@ export class ProgramApiService extends DomainApiService {
...
+              ...attribute,
+              label: this.getAttributeTranslatedLabelOrName(attribute),
+            };
+          },
+        );
+      },
     });
   }
 
+  private getAttributeTranslatedLabelOrName(attribute: Attribute): string {
```
RubenGeo: I am sorry I did not mention it in my previous review. But I think this fuction would be a perfect function to unit test.

You probably need to put it in a helper file to make it public first

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,126 @@
...
+
+      if (koboLabel) {
+        const labelToShow = koboLabel[this.currentLanguage()];
+        if (labelToShow) {
+          return labelToShow;
+        }
+      }
+
+      return name;
+    });
```
RubenGeo: I think it would also be cleaner here to have map of attribute labels at once like I proposed in this [comment](https://github.com/global-121/121-platform/pull/8216/changes#r3318005683) for tabs

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,126 @@
...
+  readonly programAttributes = injectQuery(() => ({
+    ...this.programApiService.getProgramRegistrationAttributesWithUntranslatedLabels(
+      {
+        programId: this.programId,
+      },
+    )(),
+    enabled: !!this.programId(),
+  }));
+
+  readonly cardSubtitle = computed(() => {
```
RubenGeo: I don't think this cardSubtitle is used

interfaces/portal/src/app/domains/program/program.api.service.ts
RubenGeo: Maybe it's also possible to re-use the functionality to get the correct label here than, or is the logic in that file different? interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts

RubenGeo (dismissed):
Also the bug I had locally needs to be fixed. (That the languages stay on Dutch). Let me know if we need to reproduce it together

interfaces/portal/src/app/domains/program/program.api.service.ts
arsforza: > I am sorry I did not mention it in my previous review. But I think this fuction would be a perfect function to unit test.
> 
> You probably need to put it in a helper file to make it public first

It could be becaus of lack of knowledge on my part, but you can't inject a service (TranslatableStringService in this case) in a helper file.

> Maybe it's also possible to re-use the functionality to get the correct label here than, or is the logic in that file different? interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts

These are two different logics.
The one using TranslatableStringService, is used by pages in the portal to show the labels translated in the user's selected language for the portal.
The one in the component shows the labels of the language tab currently selected in the settings page.

interfaces/portal/src/app/components/program-language-tabs/program-language-tabs.component.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
+import { UILanguage } from '@121-service/src/shared/enum/ui-language.enum';
+import { Language } from '@121-service/src/shared/types/language.type';
+
+import { ProgramApiService } from '~/domains/program/program.api.service';
+import { getLinguonym } from '~/utils/get-linguonym';
+import { environment } from '~environment';
+
+const LOCAL_STORAGE_LOCALE_KEY = 'preferredLanguage';
```
elwinschmitz: This should be imported from the already existing: `interfaces/portal/src/app/utils/locale.ts`

That's the whole point of it being a `constant` :D

interfaces/portal/src/app/components/program-language-tabs/program-language-tabs.component.html
```diff
@@ -0,0 +1,28 @@
...
+  </p-tablist>
+  @if (!isCurrentLanguageSupported()) {
+    <p class="my-2 flex items-center">
+      <i class="pi pi-exclamation-triangle me-2 text-xl text-red-500"></i>
+      <span i18n>
+        Please note, the portal does not support
+        {{ currentLanguageLabel() }} language currently.
+      </span>
+    </p>
+  }
```
elwinschmitz: Moving this 'inside' each language's tab-panel would maybe remove the need for keeping track of "what is current".

Maybe, some day. only :D

interfaces/portal/src/app/components/program-language-tabs/program-language-tabs.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  readonly languageTemplate =
+    contentChild<TemplateRef<unknown>>('languageTemplate');
+
+  readonly programApiService = inject(ProgramApiService);
+
+  readonly defaultLocale = localStorage.getItem(LOCAL_STORAGE_LOCALE_KEY);
+
+  readonly portalLanguages = environment.locales
+    .split(',')
+    .map((locale) => (locale === 'en-GB' ? 'en' : locale));
```
elwinschmitz: Maybe put them as separate utils into `utils/locale.ts`?

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,161 @@
...
+import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
+
+import { injectQuery } from '@tanstack/angular-query-experimental';
+import { CardModule } from 'primeng/card';
+import { InputTextModule } from 'primeng/inputtext';
+import { TableModule } from 'primeng/table';
+
+import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
+
+import { UILanguage } from '../../../../../../../../services/121-service/src/shared/enum/ui-language.enum';
```
elwinschmitz: This should use the ` from '@121-service/...`-shape, right?

Isn't some lint-rule catching this? :'(

interfaces/portal/src/app/pages/program-settings-registration-data/program-settings-registration-data.page.html
```diff
@@ -16,5 +16,8 @@
       [programId]="programId()"
       class="w-1/2"
     ></app-kobo-integration-card>
+    <app-registration-questions-card
+      [programId]="programId()"
+    ></app-registration-questions-card>
```
elwinschmitz: In the design the "Registration integration"-section and the "Registration questions"-section are each similar looking "cards with content in them".

I'd think this page-level would be the good place to define that layout, so 'inside" the "`app-registration-questions`"-component, you shouldn't have to worry about the layout / text-content like headings etc..

(The component doesn't have to be a `*-card` then as well... (in both name AND HTML).

interfaces/portal/src/app/domains/program/program.api.service.ts
```diff
@@ -263,24 +276,46 @@ export class ProgramApiService extends DomainApiService {
...
+        return translated;
+      }
+    }
+
+    if (isGenericAttribute(attribute.name)) {
+      return ATTRIBUTE_LABELS[attribute.name];
+    }
+
+    return;
+  }
```
elwinschmitz, line 303: This feels like a duplication of here:

https://github.com/global-121/121-platform/pull/8216/changes#diff-eadb8d75aba3c9f48b818baba843b50832cdb6683b4f9d64863fd4dcd20c9348R136-R160

They're slightly different, but I think both places should use 1/2 helper-functions to keep the logic of "what string gets to be the label" all in 1 place (and easy to reuse)...

interfaces/portal/src/app/pages/program-settings-registration-data/components/registration-questions-card/registration-questions-card.component.ts
```diff
@@ -0,0 +1,161 @@
...
+
+    if (koboLabel) {
+      const labelToShow = koboLabel[language as unknown as UILanguage];
+      if (labelToShow) {
+        return labelToShow;
+      }
+    }
+
+    return;
+  }
```
elwinschmitz, line 159: See also: https://github.com/global-121/121-platform/pull/8216/changes#diff-9eabedbfbb3672554a37139311d19d74d29d20816198691d5a0d3570636f9120R293-R317

(and the comment there.)

elwinschmitz (approved):

PR #8246, opened 2026-05-15
Refactor interface of adding messages to queue

RubenGeo:
Signed-off-by: Ruben <vandervalk@geoit.nl>[AB#42246](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42246) 

## Describe your changes

Refactors the interface of adding messages to queue & centralizes message-job creation in RegistrationsService

**Why I created this refactor?**

- I tried to remove the phonenumber from the registration entity (the goal of the main item of subtask [AB#42246](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42246))
- However I ran into the fact that the widely used `queueMessageService.addMessageJob` required a complete registration entity of which it used the phonenumber (which I tried to remove from the registrationEntity). It required a lot of hassle and hacks to work around it without a more wider refactor
- Passing a complete entity is not in line with our [guidelines](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#naming-guidelines) B.4
- To keep stuff simple for the reviewer and myself I decided to do this refactor as a seperate PR 

**Why did I decide to add a method to the registration service?**

I refactored the MessageQueuesService.addMessageJob to accept flat fields (registrationId, phoneNumber, whatsappPhoneNumber, preferredLanguage) instead of complete registration entities. So each call side now has to get this information from the database. This means that each call side should already have knowledge of the 'concept' of a registration and should have used the registration repository or registration pagination service to get this information. This means that dependency wise this already makes sense.

**Alternatives I considered where:**

-  fetching the registration data in the MessageQueuesService (this was already done before in [hacky way](https://github.com/global-121/121-platform/pull/8246/changes#diff-394e74645996831e0fc1a084411df57691a86c2984f88757a3757b0304829a0dR83-R84) by using the RegistrationDataService), which is now removed in this PR. I would argue it would make a lot more sense conceptually for the registration module to know about the messaque module than the other way around
- Adding a method to the RegistrationModule called something like '`fetchRegistrationAttributesNeededForMessages`'  and than calling that at each call side just before calling addMessageJob. However the RegistrationModule already uses the MessageQueuesService to send message itself and a method like `fetchRegistrationAttributesNeededForMessages` also implies that the registration module knows about messaging. So I think such as solution would only result in overhead compared to the current implementation. 

**No we tests were added?**

I considered adding unit tests. However the only method I really added was createMessageJobForRegistration and I does not contains extra logic that is not already tested by integration tests

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
+++ b/services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
@@ -169,6 +169,11 @@ export class IntersolveVoucherService {
       return paResult;
     }
 
+    if (!whatsappPhoneNumber) {
+      // This should never happen because the API layer should prevent this, it's to make TS happy
+      throw Error('WhatsApp phone number is required when useWhatsapp is true');
+    }
+
     // Continue with WhatsApp:
     return await this.sendWhatsapp({
       programFspConfigurationId,
@@ -176,6 +181,7 @@ export class IntersolveVoucherService {
       paResult,
       transactionId,
       bulkSize,
+      whatsappPhoneNumber,
       userId,
     });
   }
@@ -273,6 +279,7 @@ export class IntersolveVoucherService {
     transactionId,
     bulkSize,
     userId,
+    whatsappPhoneNumber,
   }: {
     programFspConfigurationId: number;
     referenceId: string;
@@ -280,6 +287,7 @@ export class IntersolveVoucherService {
     transactionId: number;
     bulkSize: number;
     userId: number;
+    whatsappPhoneNumber: string;
   }): Promise<PaTransactionResultDto> {
     const transferResult = await this.sendVoucherWhatsapp({
       referenceId,
@@ -287,6 +295,7 @@ export class IntersolveVoucherService {
       bulkSize,
       userId,
       programFspConfigurationId,
+      whatsappPhoneNumber,
     });
 
     paResult.status = transferResult.status;
@@ -301,12 +310,14 @@ export class IntersolveVoucherService {
     bulkSize,
     userId,
     programFspConfigurationId,
+    whatsappPhoneNumber,
   }: {
     referenceId: string;
     transactionId: number;
     bulkSize: number;
     userId: number;
     programFspConfigurationId: number;
+    whatsappPhoneNumber: string;
   }): Promise<PaTransactionResultDto> {
     const result = new PaTransactionResultDto();
     result.referenceId = referenceId;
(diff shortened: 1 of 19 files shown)


services/121-service/src/registration/services/registrations.service.ts
```diff
@@ -100,21 +109,88 @@ export class RegistrationsService {
     return registration;
   }
 
+  public async createMessageJobForRegistration({
```
RubenGeo, line 112: I hate adding another method to the already large registration.service. I think however that creating a separate service just for this function in the registration module is a bit much. 

I would be better to (some day ;) ) create a seperate service for all the unique /deduplication functionality in the registration service. Or another more thorough refactor of the registration module

services/121-service/src/registration/services/registrations.service.ts
```diff
@@ -100,21 +109,88 @@ export class RegistrationsService {
...
+
+    await this.messageQueuesService.addMessageJob({
+      registrationId: registration.id,
+      referenceId,
+      programId,
+      preferredLanguage: registration.preferredLanguage,
+      phoneNumber:
+        registration[DefaultRegistrationDataAttributeNames.phoneNumber],
+      whatsappPhoneNumber:
+        registration[DefaultRegistrationDataAttributeNames.whatsappPhoneNumber],
```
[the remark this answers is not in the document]
RubenGeo: We expect that this array is almost always 1. (Because it's an array or registrations that share the same phonenumber)

services/121-service/src/registration/services/registrations.service.ts
```diff
@@ -100,21 +109,88 @@ export class RegistrationsService {
...
+        query: { path: '', limit: 1, select },
         programId,
         hasPersonalReadPermission: true,
         queryBuilder,
       });
+    if (paginateResult.data.length !== 1) {
+      throw new Error(
+        `Unexpected error: ReferenceId '${referenceId}' not found in pagination query results.`,
+      );
+    }
```
Yuri-510: Out of scope, but since we're doing a refactor it might be nice to take along:

This function is a bit unintuitive. It makes a reader, that is not familiar with the shape this func returns, question why a paginator is being used to get one row.

Looking at the error statement here, I question: "why must the length be 1"? Than I see "ah, it is limited to 1". But why then use a paginator? And then see a not so super descriptive comment.

Also, starting from the top, so by the function name, while quickly scanning it's easy to miss that it just returns one registration, because my brain reads `paginate` first.

Consider implementing something like:

```  
public async getRegistrationViewById({
    programId,
    id,
  }: {
    programId: number;
    id: number;
  }): Promise<MappedPaginatedRegistrationDto | null> {
    const queryBuilder = this.registrationViewScopedRepository
      .createQueryBuilder('registration')
      .andWhere({ id });

    const [result] = await this.getRegistrationViewsNoLimit({
      paginateQuery: { path: '' },
      programId,
      queryBuilder,
    });
    return result ?? null;
  }
```

services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
```diff
@@ -328,15 +339,20 @@ export class IntersolveVoucherService {
       language,
     });
 
+    // TODO: Refactor this to be called from the TransactionJobsIntersolveVoucherService
```
Yuri-510, line 342: Is this too large of a refactor to do in this PR?

Yuri-510 (commented):
Clean refactor! Only one question and one suggestion.

Yuri-510 (dismissed):

Yuri-510 (approved):

PR #8253, opened 2026-05-19
refactor: send emails by calling graph api directly

Yuri-510:
[AB#41674](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/41674) 

Description yet to be added

## Portal preview-deployment


services/121-service/src/emails/graph/graph.service.ts
```diff
@@ -0,0 +1,87 @@
...
+        typeof response.status === 'number' && response.status >= 100;
+
+      const detail = isHttpResponse
+        ? `HTTP ${response.status} ${response.statusText ?? ''}`.trim()
+        : `network error (${response.statusText ?? 'unknown'})`;
+
+      throw new EmailDeliveryError(`Failed to send email: ${detail}`);
+    }
+
+    console.log(`Email sent successfully to ${emailData.email}`);
```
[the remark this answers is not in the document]
Yuri-510: Keeping this till after all testing on production is done.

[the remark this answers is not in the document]
elwinschmitz: We could also put a 'masked' value in the logs?
So to remove some characters before AND after the `@`?

services/121-service/src/emails/graph/const/graph-attachment-data-type.cont.ts
```diff
@@ -0,0 +1 @@
+export const GRAPH_ATTACHMENT_DATA_TYPE = '#microsoft.graph.fileAttachment';
```
elwinschmitz:
```suggestion
export const GRAPH_ATTACHMENT_DATA_TYPE = '#microsoft.graph.fileAttachment';

```

This missing final newline should(?) probably fail the Prettier/ESLint-check?

Probably because of the merge-conflict that it isn't run fully? 🤷 😕

services/121-service/src/emails/graph/graph.service.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+        'https://graph-api.example.test/users/sender%40example.org/sendMail',
+      );
+    });
+
+    it('should map EmailData to the correct Graph payload', async () => {
+      httpService.post.mockResolvedValue({ status: HttpStatus.ACCEPTED });
+
+      await service.sendMail(emailData);
+
+      const payload = httpService.post.mock.calls[0][1];
```
elwinschmitz: Are there no nice, readable helper-methods to get to these mocked-calls?
Like `httpService.post.mock.first()` or something?

Or at leas something that helps understanding the magic of `calls[0][1]`?

services/121-service/src/emails/graph/graph.service.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    const response = await this.httpService.post<{
+      status: number;
+      statusText?: string;
+    }>(url, payload, headers);
+
+    if (response.status !== 202) {
+      const isHttpResponse =
+        typeof response.status === 'number' && response.status >= 100;
+
```
elwinschmitz:
```suggestion
    if (response.status !== HttpStatus.ACCEPTED) {
      const isHttpResponse =
        typeof response.status === 'number' && response.status >= HttpStatus.CONTINUE;

```

We can use these named values from the `@nestjs/common`-package.

But there is already some "is this an error-response of this type"-kind of helper method... (But maybe its only available in test-code?)

Update: Ah, its only in Portal-code: https://github.com/global-121/121-platform/blob/main/interfaces/portal/eslint.config.mjs#L1-L273

services/121-service/src/emails/emails.service.spec.ts
```diff
@@ -1,114 +1,100 @@
...
+             <h1>121 Portal</h1>
+           </div>
+
+           <div class="content">
+             Body
+           </div>
+
+           <div class="footer">
+             121 Support: <a href="mailto:support@121.global" style="color:#fff">support@121.global</a>
+           </div>
```
elwinschmitz, line 95: 😢 
This _is_ a quick way to test the output... but it feels a bit weird "just copy-and-paste everything". :/

services/mock-service/src/email/email.mock.service.ts
```diff
@@ -1,11 +1,17 @@
...
-    return { message: 'Email accepted' };
+  public sendMail({
+     sender: _sender,
+     request: _request,
+  }: {
+    sender: string;
+    request: SendMailRequestDto;
+  }): void {
+    // Mock implementation: accept and discard. Matches the real Microsoft
+    // Graph `sendMail` endpoint, which returns 202 Accepted with no body.
```
elwinschmitz, line 15: Maybe nice for local development; we could `console.log` something like: "An email would've been sent now..." ;)

services/121-service/src/emails/graph/const/graph-attachment-data-type.cont.ts
Yuri-510: Ah it was also named "cont". Updated!

services/121-service/src/emails/graph/graph.service.spec.ts
Yuri-510: I tried destructuring to make it a little less magic. It's not really pretty though.

services/121-service/src/emails/graph/graph.service.ts
Yuri-510: We have it. I updated 202 to `ACCEPTED`. However, I think it is more readable to have `status >= 100`, since 100 is where http statuses start. It makes less sense to assert that the status is above `CONTINUE`.

services/121-service/src/emails/emails.service.spec.ts
Yuri-510: This is from an inline snapshot ;)

services/121-service/src/emails/graph/graph.service.ts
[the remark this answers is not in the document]
Yuri-510: The logs have been removed

services/121-service/src/emails/graph/graph.service.spec.ts
elwinschmitz: 🙈

services/121-service/src/emails/graph/graph.service.ts
elwinschmitz: Yeah; Some sort of helper like "this looks like any valid HTTP Status-code"; (But that is a bit much to include in this PR. ;)

services/121-service/src/emails/emails.service.spec.ts
elwinschmitz: Yeah; the 'upside' of it being a snapshot is that you can use any built-in tools to "update the snapshot with the current new value", right?

So you don't _have to_ manually also update a pixel-value here when that CSS changes in its 'proper place'.

Its a sort of E2E-test now. ;)

services/121-service/src/emails/graph/const/graph-attachment-data-type.cont.ts
elwinschmitz: 🍑

elwinschmitz (approved):

PR #8289, opened 2026-05-27
feature: refresh existing kobo integrations

Yuri-510:
[AB#42383](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42383) 

Added an endpoint to refresh a program after updating a linked Kobo form

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/kobo/kobo.controller.ts
+++ b/services/121-service/src/kobo/kobo.controller.ts
@@ -39,6 +39,7 @@ import { KoboService } from '@121-service/src/kobo/services/kobo.service';
 import { KoboSubmissionService } from '@121-service/src/kobo/services/kobo-submission.service';
 import { MAX_IMPORT_RECORDS } from '@121-service/src/registration/services/registrations-creation.service';
 import { ScopedUserRequest } from '@121-service/src/shared/scoped-user-request';
+import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
 import { RequestHelper } from '@121-service/src/utils/request-helper/request-helper.helper';
 
 @UseGuards(AuthenticatedUserGuard)
@@ -166,6 +167,40 @@ export class KoboController {
     programId: number,
   ): Promise<KoboResponseDto> {
     return this.koboService.getKoboData({ programId });
+  }
+
+  @AuthenticatedUser({ permissions: [PermissionEnum.ProgramUPDATE] })
+  @ApiOperation({
+    summary: 'Refresh Kobo form integration for a Program',
+    description:
+      'Fetches the latest deployed Kobo form definition and updates the program registration attributes accordingly. Use this when you have made changes to your Kobo form and want to apply those changes to the program.',
+  })
+  @ApiParam({
+    name: 'programId',
+    required: true,
+    type: 'integer',
+    description: 'The unique identifier of the program',
+    example: 1,
+  })
+  @ApiResponse({
+    status: HttpStatus.OK,
+    description: 'Kobo form refreshed successfully',
+    type: KoboIntegrationResultDto,
+  })
+  @ApiResponse({
+    status: HttpStatus.NOT_FOUND,
+    description: 'No Kobo integration found for this program',
+  })
+  @Patch('programs/:programId/kobo')
+  public async refreshKoboForm(
+    @Param('programId', ParseIntPipe)
+    programId: number,
+  ): Promise<KoboIntegrationResultDto> {
+    const result = await this.koboService.refreshKoboForm({ programId });
+    return {
+      message: 'Kobo form refreshed successfully',
+      name: result.name,
+    };
   }
 
   @AuthenticatedUser({ isAdmin: true })
--- a/services/121-service/src/kobo/kobo.controller.ts
+++ b/services/121-service/src/kobo/kobo.controller.ts
@@ -173,7 +173,7 @@ export class KoboController {
   @ApiOperation({
     summary: 'Refresh Kobo form integration for a Program',
     description:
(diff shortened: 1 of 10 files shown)


services/121-service/src/kobo/kobo.controller.ts
```diff
@@ -168,6 +169,53 @@ export class KoboController {
...
+      'Validation failed - the latest Kobo form definition does not meet program requirements.',
+  })
+  @ApiResponse({
+    status: HttpStatus.UNAUTHORIZED,
+    description: 'Authentication required',
+  })
+  @ApiResponse({
+    status: HttpStatus.FORBIDDEN,
+    description: 'Insufficient permissions to refresh the Kobo form',
+  })
```
RubenGeo: No need to add these, they will come automatically from this file: 

services/121-service/src/guards/authenticated-user.decorator.ts

Maybe it would be nice to write an eslint rule for this so we do not forget in the future ;)
```suggestion
```

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -161,6 +192,17 @@ export class KoboService {
       formDefinition,
       programId,
     });
```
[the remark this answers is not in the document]
RubenGeo: We don't care about this enough @Yuri-510

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -161,6 +192,17 @@ export class KoboService {
...
+    const filter =
+      filterByCurrentVersionId !== undefined
+        ? { programId, versionId: filterByCurrentVersionId }
+        : { programId };
+
+    await this.koboRepository.update(filter, {
+      versionId: formDefinition.versionId,
+      dateDeployed: formDefinition.dateDeployed,
+      name: formDefinition.name,
+    });
```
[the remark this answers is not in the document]
RubenGeo: This will never happen. If it happens it does not matter, because it's 2 registrations that come in very fast/concurrently and than the first one already updated the kobo entity, which is fine

services/121-service/src/kobo/kobo.controller.ts
```diff
@@ -168,6 +169,53 @@ export class KoboController {
     return this.koboService.getKoboData({ programId });
   }
 
+  @AuthenticatedUser({ permissions: [PermissionEnum.ProgramUPDATE] })
+  @ApiOperation({
+    summary: 'Refresh Kobo form integration for a Program',
+    description:
+      'Fetches the latest deployed Kobo form definition and updates the program registration attributes accordingly. Use this when you have made changes to your Kobo form and want to apply those changes to the program.',
```
RubenGeo: I think you could add here that the refresh: also updates the languages of a program

services/121-service/test/kobo/kobo-refresh-form.test.ts
```diff
@@ -0,0 +1,66 @@
...
+const koboLinkDto: CreateKoboDto = {
+  token: 'mock-token',
+  assetUid: KoboMockAssetUids.happyFlow,
+  url: `${env.MOCK_SERVICE_URL}/api/kobo`,
+};
+
+describe('Refresh Kobo form', () => {
+  let accessToken: string;
+
+  beforeEach(async () => {
```
RubenGeo: I think it would be better to not reset the database on each test (to save a bit of processing time, also if someone adds tests in the future to this file it will add up). 

I think we have examples on how to use beforeAll for this on the other kobo tests.

services/121-service/test/kobo/kobo-refresh-form.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    const response = await refreshKoboForm({
+      programId: programIdSafaricom,
+      accessToken,
+    });
+
+    // Assert
+    expect(response.status).toBe(HttpStatus.OK);
+    expect(response.body).toMatchObject({
+      message: 'Kobo form refreshed successfully',
+      name: '25042025 Prototype Sprint',
```
RubenGeo, line 115: I think it is more robust to also assert also assert if the program was updated 👍

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -146,12 +146,43 @@ export class KoboService {
...
+    const formDefinition = await this.getFormDefinitionOrThrow({
+      assetUid: koboIntegration.assetUid,
+      token: koboIntegration.token,
+      url: koboIntegration.url,
+    });
+
+    await this.applyFormDefinitionToProgram({
+      formDefinition,
+      programId,
+    });
```
RubenGeo, line 174: Maybe we can also pass currentVersionId here. Since we already load the koboIntegration entity (which has the current versionId). Than we can remove the optionality from that param which would make the applyFormDefinitionToProgram less complex

RubenGeo (changes requested):
Nice work 👍 .

services/121-service/src/kobo/kobo.controller.ts
Yuri-510: Ah good idea to write a custom rule. Will do that in a different PR.

services/121-service/test/kobo/kobo-refresh-form.test.ts
Yuri-510: I updated and followed the pattern from the other Kobo tests

services/121-service/test/kobo/kobo-refresh-form.test.ts
Yuri-510: You're right. I updated the test.

RubenGeo (approved):
Bon!

PR #8307, opened 2026-06-01
app-card-with-link aligned with current and future design

nielsVoogt:
[AB#42528](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42528)

Aligns the FSP card with current UI design and prepares for future development on 'pending' FSP's.
Also related to [AB#42050](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42050)

## Portal preview-deployment

https://happy-rock-0411d2003-8307.westeurope.3.azurestaticapps.net


interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-integration-card/kobo-integration-card.component.html
```diff
@@ -1,49 +1,41 @@
...
-  [attr.data-testid]="'kobo-integration-card'"
 >
-  <ng-container card-content>
+  <ng-container card-subtitle>
     @if (isKoboIntegrated()) {
-      <div class="absolute inset-e-4 top-2">
-        <app-ellipsis-menu [menuItems]="menuItems()" />
+      <div class="flex items-end justify-between">
+        <app-external-link
+          [href]="externalFormUrl()"
```
elwinschmitz: The change that this method returns `''` instead of `null` has probably an effect on this `href`-property... (`null` makes it _not_ render, I think..)

But I don't know if this is an actual-possible scenario.. ;)

interfaces/portal/src/app/components/external-link/external-link.component.html
```diff
@@ -0,0 +1,13 @@
+<a
+  [href]="href()"
+  [rel]="'noopener noreferrer external'"
+  target="_blank"
+  title="Opens in a new window"
+  i18n-title="@@generic-opens-in-new-window"
+  class="underline"
```
elwinschmitz: We only have a component like this for the Manual-link, right?
And that had probably some other specific styling.

(I don't think it is necessary to include this component in that Manual-link-component though!.)

Probably because we hadn't really had multiple use-cases to warrant a _real_ component...

interfaces/portal/src/app/components/external-link/external-link.component.html
```diff
@@ -0,0 +1,13 @@
...
+>
+  {{ label() }}
+  <span
+    class="p-button-icon pi pi-external-link relative top-0.5 ms-2 text-sm"
```
elwinschmitz: If edditing (some) files anyway, this could also receive an extra fancy attribute...


```suggestion
    class="p-button-icon pi pi-external-link relative top-0.5 ms-2 text-sm"
    aria-hidden="true"
```

Just to be sure; I'm surprised I hadn't added it already. ;)

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-integration-card/kobo-integration-card.component.ts
```diff
@@ -61,18 +61,12 @@ export class KoboIntegrationCardComponent {
...
-    this.isKoboIntegrated() ? '' : $localize`Click to integrate`,
-  );
-
-  readonly externalFormUrl = computed<null | string>(() => {
-    if (!this.isKoboIntegrated()) {
-      return null;
-    }
+  readonly externalFormUrl = computed<string>(() => {
     const koboIntegrationData = this.koboIntegration.data();
     if (!koboIntegrationData) {
```
elwinschmitz, line 79: I can vaguely remember that this extra check was added-in-front to prevent some additional requests (and 404s) on loading that page...

Is its currently not doing extra/unnecessary requests?

interfaces/portal/src/app/components/external-link/external-link.component.html
nielsVoogt: @elwinschmitz Yes, we have one I and I wasn't aware of that. But the manual-link has a different style, and a hardcoded URL to the manual. I could align these as well, or remove the external-link component.

elwinschmitz: No; I think they both serve a slightly different purpose, so its fine to exist side-by-side.

As long as the icon-style/size/position is the same that's the only 'duplication' that matters..

As long as it won't be used all-over-the-place, its fine like this.

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-integration-card/kobo-integration-card.component.ts
nielsVoogt: @elwinschmitz There is a lot more going on buildKoboFormUrl so I'm not touching that.

interfaces/portal/src/app/components/external-link/external-link.component.html
nielsVoogt: Good one 👍

interfaces/portal/src/app/components/external-link/external-link.component.html
nielsVoogt: @elwinschmitz I removed the component and put it back in the kobo card itself. If we get more external URL's at other places (*not to manual) I will make a component out of it.

interfaces/portal/src/app/components/card-with-link/card-with-link.component.ts
```diff
@@ -7,12 +7,15 @@ import {
 } from '@angular/core';
 import { RouterLink } from '@angular/router';
 
+import { MenuItem } from 'primeng/api';
 import { CardModule } from 'primeng/card';
+import { ClassNamesModule } from 'primeng/classnames';
```
elwinschmitz: What does this thing do?

I think its not used, right? Or is it for the `pClass`-attribute/directive?

There is already built-in functionality from Angular for this, right?

nielsVoogt: Replaced it with Angulars `pClass`

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-integration-card/kobo-integration-card.component.html
```diff
@@ -1,49 +1,53 @@
...
-  <ng-container card-content>
+  <ng-container card-subtitle>
     @if (isKoboIntegrated()) {
-      <div class="absolute inset-e-4 top-2">
-        <app-ellipsis-menu [menuItems]="menuItems()" />
+      <div class="flex items-end justify-between">
+        @if (externalFormUrl(); as formUrl) {
+          <a
+            href="{{ formUrl }}"
+            rel="'noopener noreferrer external'"
```
elwinschmitz:
```suggestion
            [href]="formUrl"
            rel="external"
```

The `noopener` and `noreferrer` attributes are already covered in [some header-config](https://github.com/global-121/121-platform/blob/main/interfaces/portal/staticwebapp.config.base.json#L71-L77)

And the `i18n`-linting shouldn't complain about the `rel`-attribute, I'd think.

elwinschmitz (dismissed):
Oh, ok.... This was the last thing. :)

(You can also put these small things in a follow-up/spin-off PR, to safe some test-runner-time)

elwinschmitz (approved):

PR #8308, opened 2026-06-01
feat: implement batch visa debit card ordering with admin controls an…

nameerhqureshi:
…d scoped permissions

[AB#42323](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42323) 

## Portal preview-deployment


services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.controller.ts
```diff
@@ -299,4 +311,69 @@ export class IntersolveVisaAccountManagementController {
...
+
+  @AuthenticatedUser({ permissions: [PermissionEnum.FspDebitCardOrderCREATE] })
+  @ApiOperation({
+    summary: 'Order visa debit cards in batch for a branch delivery address',
+  })
+  @ApiParam({ name: 'programId', required: true, type: 'integer' })
+  @ApiBody({ type: CreateVisaCardOrderDto })
+  @ApiResponse({
+    status: HttpStatus.CREATED,
+    description: 'Cards ordered and successful order persisted.',
```
Yuri-510: We can leave out "successful order persisted". I don't think it will be useful to the end user.

services/121-service/src/fsp-integrations/integrations/intersolve-visa/services/intersolve-visa.service.ts
```diff
@@ -966,4 +966,25 @@ export class IntersolveVisaService {
...
+    contactInformation,
+  }: {
+    brandCode: string;
+    coverLetterCode: string;
+    contactInformation: ContactInformation;
+  }): Promise<void> {
+    const issuedToken = await this.intersolveVisaApiService.issueToken({
+      brandCode,
+      activate: false,
+    });
```
Yuri-510, line 983: Consider adding a check here:

```
if (!issuedToken?.code) {
  throw new IntersolveVisaApiError(
    `${IntersolveVisa121ErrorText.issueTokenError}: response did not contain a token`,
  );
}
```

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+    postalCode: string;
+    address: string;
+    addressee: string;
+    phoneNumber?: string;
+    userId: number;
+  }): Promise<{
+    noOfCardsSent: number;
+    noOfCardsOrdered: number;
+  }> {
+    const parsedAddress = this.parseAddressOrThrow(address);
```
Yuri-510: Could this be a public helper, so that we can properly unit test it?

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+        programId,
+        fspName: Fsps.intersolveVisa,
+      });
+
+    if (visaProgramFspConfigurations.length !== 1) {
+      throw new HttpException(
+        `Expected exactly 1 Intersolve Visa configuration for program ${programId}, found ${visaProgramFspConfigurations.length}.`,
+        HttpStatus.BAD_REQUEST,
+      );
+    }
```
Yuri-510, line 520: This check would be more natural as a database constraint: a @Unique(['programId', 'fspName']) on ProgramFspConfigurationEntity would enforce this invariant at the right layer and make the runtime guard redundant.

Yuri-510: A bit out of scope maybe. What do you think?

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+          FspConfigurationProperties.brandCode,
+          FspConfigurationProperties.coverLetterCode,
+          FspConfigurationProperties.cardDistributionByMail,
+        ],
+      });
+
+    const cardDistributionByMail = intersolveVisaFspConfig.find(
+      (property) =>
+        property.name === FspConfigurationProperties.cardDistributionByMail,
+    )?.value as boolean;
```
Yuri-510: We have a nice method for this:

```
const cardDistributionByMail =
      await this.programFspConfigurationRepository.getPropertyValueByNameOrThrow(
        {
          programFspConfigurationId,
          name: FspConfigurationProperties.cardDistributionByMail,
        },
      );
```

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+
+    const brandCode = intersolveVisaFspConfig.find(
+      (property) => property.name === FspConfigurationProperties.brandCode,
+    )?.value as string;
+
+    const coverLetterCode = intersolveVisaFspConfig.find(
+      (property) =>
+        property.name === FspConfigurationProperties.coverLetterCode,
+    )?.value as string;
+
```
Yuri-510: You can use the `getPropertyValueByNameOrThrow` method here as well.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+      (property) =>
+        property.name === FspConfigurationProperties.cardDistributionByMail,
+    )?.value as boolean;
+
+    if (!cardDistributionByMail) {
+      throw new HttpException(
+        'Card ordering by address is not allowed when card distribution by mail is disabled.',
+        HttpStatus.BAD_REQUEST,
+      );
+    }
```
Yuri-510: I think this should actually be the other way around. Until now we had two flows: either send a card by mail to the designated registration, or link a card to a registration on site (a registration comes by to pick up a card).

We're now creating the possibility to batch order cards so that we can link those card to registrations when they come by.

So ordering batches should only be possible when `cardDistributionByMail` is set to false.

Can you also change the message to 'Batch ordering Visa cards is not allowed when card distribution by mail is enabled.'

Yuri-510: Come to think of it, this can be confusing for a user. We might need to change `cardDistributionByMail` to something like `sendCardToRegistrationByMail`. However, this calls for another migration and is quite an overhaul. Something to consider

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+      (property) => property.name === FspConfigurationProperties.brandCode,
+    )?.value as string;
+
+    const coverLetterCode = intersolveVisaFspConfig.find(
+      (property) =>
+        property.name === FspConfigurationProperties.coverLetterCode,
+    )?.value as string;
+
+    const cardOrderPhoneNumber =
+      phoneNumber ?? this.defaultCardOrderPhoneNumber;
```
Yuri-510: I thought we would create an env var for the phone number? Did you consciously reconsidered this?

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+    order.city = city;
+    order.postalCode = postalCode;
+
+    try {
+      await this.cardOrderRepository.save(order);
+    } catch (error) {
+      this.logger.error(
+        `Card order persistence failed after Intersolve success for program ${programId}. noOfCards=${noOfCards} noOfCardsOrdered=${cardsOrdered} userId=${userId}`,
+        error,
+      );
```
Yuri-510: This is not necessary. We can see the exception in our logs.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,264 @@ export class IntersolveVisaAccountManagementService {
...
+      throw new HttpException(
+        'Cards were ordered, but saving the batch record failed. Please contact support for reconciliation.',
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+
+    return {
+      noOfCardsSent: noOfCards,
+      noOfCardsOrdered: cardsOrdered,
+    };
```
Yuri-510, line 618: Ok here it gets a bit confusing. I think it should be the the other way around:

```
return {
      noOfCardsSent: cardsOrdered,
      noOfCardsOrdered: noOfCards,
    };
```

For me `noOfCardsSent` means how many cards Intersolve is going to send to the person ordering the cards. `noOfCardsOrdered` means how many were ordered in the first place.

So maybe, to make the separation a bit more clear, change `cardsOrdered` to `cardsSentByIntersolve`.

The the return would look like this:

```
return {
      noOfCardsSent: cardsSentByIntersolve,
      noOfCardsOrdered: noOfCards,
    };
```

Yuri-510 (commented):
It's looking very good so far!

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,265 @@ export class IntersolveVisaAccountManagementService {
...
+
+  public async getVisaCardOrders({
+    programId,
+    paginateQuery,
+  }: {
+    programId: number;
+    paginateQuery: PaginateQuery;
+  }): Promise<FindAllVisaCardOrdersResultDto> {
+    const queryWithLimit: PaginateQueryLimitRequired = {
+      ...this.normalizeVisaCardOrderPaginateQuery({ paginateQuery }),
```
Yuri-510: The conversion of `orderedByUsername` → `user.username` now makes this code a bit overly complicated,

Lets just accept `user.username` as-is in the query params and drop the translation entirely. The dot notation is slightly awkward in a query string but not unusual (`?sortBy=user.username:DESC`). That would let you delete both private methods and replace normalizeVisaCardOrderPaginateQuery with a direct pass-through.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,265 @@ export class IntersolveVisaAccountManagementService {
...
+  }
+
+  public async getVisaCardOrders({
+    programId,
+    paginateQuery,
+  }: {
+    programId: number;
+    paginateQuery: PaginateQuery;
+  }): Promise<FindAllVisaCardOrdersResultDto> {
+    const queryWithLimit: PaginateQueryLimitRequired = {
```
Yuri-510: I think the data volume on batch orders will remain quite low. I think it'll be perfectly fine to do filtering/sorting in the frontend. That will make this PR a _whole_ lot simpler. What do you think?

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,265 @@ export class IntersolveVisaAccountManagementService {
...
+        address: this.formatCardOrderAddressForDisplay({
+          addressee: order.addressee,
+          address: order.address,
+          postalCode: order.postalCode,
+          city: order.city,
+        }),
+        orderedByUsername: order.user?.username ?? `${order.userId}`,
+        created: order.created,
+      })),
+    };
```
Yuri-510: We usually create a mapper class for this. You can search for "mapper" files to get an idea.

It will look something like this:

```
export class VisaCardOrderMapper {
  public static mapEntitiesToDtos(
    entities: VisaCardOrderEntity[],
  ): VisaCardOrderResponseDto[] {
    return entities.map((entity) => VisaCardOrderMapper.mapEntityToDto(entity));
  }

  public static mapEntityToDto(
    entity: VisaCardOrderEntity,
  ): VisaCardOrderResponseDto {
    return {
      noOfCardsOrdered: entity.noOfCardsOrdered,
      address: `${entity.addressee}, ${entity.address}, ${entity.postalCode}, ${entity.city}`,
      orderedByUsername: entity.user?.username ?? `${entity.userId}`,
      created: entity.created,
    };
  }
}
```

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +488,265 @@ export class IntersolveVisaAccountManagementService {
...
+      );
+    }
+
+    return {
+      addressStreet: parsedAddress.groups.addressStreet.trim(),
+      addressHouseNumber: parsedAddress.groups.addressHouseNumber.trim(),
+      addressHouseNumberAddition:
+        parsedAddress.groups.addressHouseNumberAddition?.trim() || undefined,
+    };
+  }
```
Yuri-510: I'm asking if we can implement something like this instead of a one line address:

Image

This would call for simplicity: less stuff that can be used wrong by user and less stuff that could break in the backend.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
Yuri-510: This would call for nice testing in isolation

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
Yuri-510: See my other comment: I asked design/frontend if we can implement a more robust address form, so we don't need to parse in the backend.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
nameerhqureshi: Yes, I agree with you, will discuss with Niels.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
nameerhqureshi: We can do that but on the frontend the house number is not a separate field.
image

Yuri-510: I'm sorry, I needed to be more clear: I'm asking the frontend if we can change it to separate fields. However, I see that in the backend we first parse to separate fields and then back to one line again in `intersolve-visa.api.service` line 745. Lets discuss this in a call.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
nameerhqureshi: Yes, its out of scope and also I see there are 2 fields `name` & `fspName`. One is enum and one is the string.

```
@Unique('programFspConfigurationUnique', ['programId', 'name'])
```

services/121-service/src/config.ts
```diff
@@ -75,6 +75,8 @@ export const EXTERNAL_API = {
 // Configure Public Twilio Settings:
 // ---------------------------------------------------------------------------
 export const TWILIO_SANDBOX_WHATSAPP_NUMBER = '+14155238886';
+export const INTERSOLVE_VISA_CARD_ORDER_PHONE_NUMBER =
+  env.INTERSOLVE_VISA_CARD_ORDER_PHONE_NUMBER;
```
Yuri-510: This is redundant. Lets just use `env` in the service.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +483,157 @@ export class IntersolveVisaAccountManagementService {
...
+          name: FspConfigurationProperties.cardDistributionByMail,
+        },
+      );
+
+    if (!cardDistributionByMail) {
+      throw new HttpException(
+        'Batch ordering Visa cards is only allowed when card distribution by mail is enabled.',
+        HttpStatus.BAD_REQUEST,
+      );
+    }
```
Yuri-510: This should still be updated: batch ordering should only be possible when `cardDistributionByMail` is set to false.

So the code would be:

```
if (cardDistributionByMail) {
      throw new HttpException(
        'Batch ordering Visa cards is only allowed when card distribution by mail is enabled.',
        HttpStatus.BAD_REQUEST,
      );
    }
```

Yuri-510: Be mindful that the tests should also be updated accordingly.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.spec.ts
```diff
@@ -373,4 +395,449 @@ describe('IntersolveVisaAccountManagementService', () => {
       ]);
     });
   });
+
```
Yuri-510, line 396: Before looking further at this test, lets first make it a bit more DRY:

```javascipt
describe('visa card orders', () => {
  // Extracted constants
  const visaFspConfigs = [{ id: 123 }] as any;

  const fspConfigProperties = [
    { name: FspConfigurationProperties.brandCode, value: 'BRAND' },
    { name: FspConfigurationProperties.coverLetterCode, value: 'COVER_LETTER' },
    { name: FspConfigurationProperties.cardDistributionByMail, value: true },
  ] as any;

  // Factory for the order input — only vary what actually differs
  function buildOrderInput(
    overrides: Partial<Parameters<typeof service.createVisaCardOrder>[0]> = {},
  ) {
    return {
      programId: 1,
      noOfCards: 1,
      addressStreet: 'Damrak',
      addressHouseNumber: '1',
      addressHouseNumberAddition: 'A',
      addressPostalCode: '1011AB',
      addressCity: 'Amsterdam',
      addressee: 'John Doe',
      userId: 7,
      ...overrides,
    };
  }

  beforeEach(() => {
    // Happy-path defaults — individual tests override only what they need
    programFspConfigurationRepository.getByProgramIdAndFspName
      .mockResolvedValue(visaFspConfigs);
    programFspConfigurationRepository.getPropertiesByNamesOrThrow
      .mockResolvedValue(fspConfigProperties);
    programFspConfigurationRepository.getPropertyValueByNameOrThrow
      .mockImplementation(async ({ name }) => {
        if (name === FspConfigurationProperties.cardDistributionByMail) return true;
        if (name === FspConfigurationProperties.brandCode) return 'BRAND';
        if (name === FspConfigurationProperties.coverLetterCode) return 'COVER_LETTER';
        if (name === FspConfigurationProperties.fundingTokenCode) return 'FUNDING_TOKEN';
        return '';
      });
  });

  it('should order cards and persist only successful count', async () => {
    intersolveVisaService.issueTokenAndCreatePhysicalCard
      .mockResolvedValueOnce()
      .mockRejectedValueOnce(new IntersolveVisaApiError('temporary error'))
      .mockResolvedValueOnce();

    const result = await service.createVisaCardOrder(
      buildOrderInput({ noOfCards: 3 }),
    );
    // ... assertions unchanged ...
  });

  it('should throw when card distribution by mail is disabled', async () => {
    // Only override the one property that differs from the beforeEach default
    programFspConfigurationRepository.getPropertyValueByNameOrThrow
      .mockImplementation(async ({ name }) => {
        if (name === FspConfigurationProperties.cardDistributionByMail) return false;
        // ...
      });

    await expect(service.createVisaCardOrder(buildOrderInput())).rejects.toThrow(...);
    // ...
  });

  it('should throw when multiple visa configurations are found', async () => {
    // Only override what differs
    programFspConfigurationRepository.getByProgramIdAndFspName
      .mockResolvedValue([{ id: 123 }, { id: 124 }] as any);

    await expect(service.createVisaCardOrder(buildOrderInput())).rejects.toThrow(...);
  });

  // etc. — every other test just calls buildOrderInput() ± one override
});
```

services/121-service/src/fsp-integrations/account-management/intersolve-visa/dto/create-visa-card-order.dto.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  @ApiProperty({
+    example: '+31612345678',
+    required: false,
+    description:
+      'Optional branch contact number for card order. Defaults to a system fallback when not provided.',
+  })
+  @IsOptional()
+  @IsString()
+  public readonly phoneNumber?: string;
```
Yuri-510: Since phone number is not in the form, lets omit it from the DTO.

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.ts
```diff
@@ -473,4 +483,157 @@ export class IntersolveVisaAccountManagementService {
...
+
+    const contactInformation: ContactInformation = {
+      name: addressee,
+      addressStreet,
+      addressHouseNumber,
+      addressHouseNumberAddition,
+      addressPostalCode,
+      addressCity,
+      phoneNumber: cardOrderPhoneNumber,
+    };
```
Yuri-510: If we omit phone number from the DTO, we can update this to:

```
const contactInformation: ContactInformation = {
      name: addressee,
      addressStreet,
      addressHouseNumber,
      addressHouseNumberAddition,
      addressPostalCode,
      addressCity,
      phoneNumber: env.INTERSOLVE_VISA_CARD_ORDER_PHONE_NUMBER,
    };
```

services/121-service/src/fsp-integrations/account-management/intersolve-visa/intersolve-visa-account-management.service.spec.ts
```diff
@@ -373,4 +395,299 @@ describe('IntersolveVisaAccountManagementService', () => {
...
+      // Verify batch was persisted with partial count
+      expect(cardOrderRepository.save).toHaveBeenCalledWith(
+        expect.objectContaining({
+          noOfCards: 700,
+          noOfCardsOrdered: 693,
+        }),
+      );
+    });
+  });
```
Yuri-510, line 580: What we see here is a lot of asserting on HOW things are done > there's a lot of "we expect this function to have been called".

Instead in unit testing, we like to test more on the different logic branches; test more of WHAT the expected output will be instead of HOW we got there.

We might for example change the code to do batch calls instead of one call per loop. The outcome will be the same, but the tests will break.

Consider the following approach:

```
beforeEach(() => {
      programFspConfigurationRepository.getByProgramIdAndFspName.mockResolvedValue(
        [{ id: 123 }] as any,
      );
      programFspConfigurationRepository.getPropertyValueByNameOrThrow.mockImplementation(
        async ({ name }) => {
          if (name === FspConfigurationProperties.cardDistributionByMail)
            return false;
          if (name === FspConfigurationProperties.brandCode) return 'BRAND';
          if (name === FspConfigurationProperties.coverLetterCode)
            return 'COVER_LETTER';
          return '';
        },
      );
      intersolveVisaService.issueTokenAndCreatePhysicalCard.mockResolvedValue();
    });

    it('throws when the program has no visa configuration', async () => {
      programFspConfigurationRepository.getByProgramIdAndFspName.mockResolvedValue(
        [] as any,
      );

      await expect(
        service.createVisaCardOrder(buildOrderInput()),
      ).rejects.toThrow(
        'Expected exactly 1 Intersolve Visa configuration for program 1, found 0.',
      );
    });

    it('throws when the program has multiple visa configurations', async () => {
      programFspConfigurationRepository.getByProgramIdAndFspName.mockResolvedValue(
        [{ id: 123 }, { id: 124 }] as any,
      );

      await expect(
        service.createVisaCardOrder(buildOrderInput()),
      ).rejects.toThrow(
        'Expected exactly 1 Intersolve Visa configuration for program 1, found 2.',
      );
    });

    it('throws when card distribution by mail is enabled', async () => {
      programFspConfigurationRepository.getPropertyValueByNameOrThrow.mockImplementation(
        async ({ name }) =>
          name === FspConfigurationProperties.cardDistributionByMail
            ? true
            : '',
      );

      await expect(
        service.createVisaCardOrder(buildOrderInput()),
      ).rejects.toThrow(
        'Batch ordering Visa cards is only allowed when card distribution by mail is disabled.',
      );
    });

    it('throws when all card orders fail', async () => {
      intersolveVisaService.issueTokenAndCreatePhysicalCard.mockRejectedValue(
        new IntersolveVisaApiError('api error'),
      );

      await expect(
        service.createVisaCardOrder(buildOrderInput({ noOfCards: 2 })),
      ).rejects.toThrow(HttpException);
    });

    it('re-throws unexpected errors from issueTokenAndCreatePhysicalCard', async () => {
      intersolveVisaService.issueTokenAndCreatePhysicalCard.mockRejectedValue(
        new Error('unexpected'),
      );

      await expect(
        service.createVisaCardOrder(buildOrderInput()),
      ).rejects.toThrow('unexpected');
    });

    it('throws when persisting the order record fails', async () => {
      cardOrderRepository.save.mockRejectedValue(new Error('db down'));

      await expect(
        service.createVisaCardOrder(buildOrderInput()),
      ).rejects.toMatchObject({
        message:
          'Cards were ordered, but saving the batch record failed. Please contact support for reconciliation.',
        status: HttpStatus.INTERNAL_SERVER_ERROR,
      });
    });

    it('returns the count of sent and ordered cards, counting only successes', async () => {
      intersolveVisaService.issueTokenAndCreatePhysicalCard
        .mockResolvedValueOnce()
        .mockRejectedValueOnce(new IntersolveVisaApiError('temporary error'))
        .mockResolvedValueOnce();

      const result = await service.createVisaCardOrder(
        buildOrderInput({ noOfCards: 3 }),
      );

      expect(result).toEqual({ noOfCardsSent: 2, noOfCardsOrdered: 3 });
    });
  });

  describe('getVisaCardOrders', () => {
    it('returns mapped card orders', async () => {
      cardOrderRepository.getForProgram.mockResolvedValue([
        {
          id: 42,
          noOfCardsOrdered: 5,
          addressee: 'John Doe',
          addressStreet: 'Damrak',
          addressHouseNumber: '1',
          addressHouseNumberAddition: 'A',
          addressPostalCode: '1011AB',
          addressCity: 'Amsterdam',
          userId: 7,
          user: { username: 'manager@example.org' },
          created: new Date('2026-05-26T08:30:00.000Z'),
        },
      ] as any);

      const result = await service.getVisaCardOrders({ programId: 1 });

      expect(result).toEqual([
        {
          id: 42,
          noOfCardsOrdered: 5,
          address: 'John Doe, Damrak 1 A, 1011AB, Amsterdam',
          orderedByUsername: 'manager@example.org',
          created: new Date('2026-05-26T08:30:00.000Z'),
        },
      ]);
    });
  });
```

Here we test for the different branches and what the expected output will be. We say: we want this method to behave this way, but we don't care how.

Yuri-510 (approved):

PR #8329, opened 2026-06-11
Show Kobo import errors in modal

nielsVoogt:
[AB#42586](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42586)

## Describe your changes

Better errors for KOBO integration modal

## Portal preview-deployment

https://happy-rock-0411d2003-8329.westeurope.3.azurestaticapps.net


services/121-service/src/kobo/interfaces/kobo-validation-error.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+  TypeMismatch = 'type_mismatch',
+  InvalidChoice = 'invalid_choice',
+  FormConfiguration = 'form_configuration',
+  ForbiddenAttribute = 'forbidden_attribute',
+  MissingEnglishLanguage = 'missing-english-language',
+  MatrixTypeFound = 'matrix-type-found',
+  InvalidLanguageCode = 'invalid-language-code',
+  SelectOneNoChoices = 'select-one-no-choices',
+  MissingFullnameAttributes = 'missing-fullname-attributes',
+}
```
RubenGeo: You have probably not looked at this documentation as you mostly worked on FE and not BE. But enums (that are exported) should be in a separate file according to our guidelines point b5 https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#b-nestjs-module-implementation

RubenGeo:
```suggestion
export enum KoboValidationErrorType {
  missingField = 'missingField',
  typeMismatch = 'typeMismatch',
  invalidChoice = 'invalidChoice',
  formConfiguration = 'formConfiguration',
  forbiddenAttribute = 'forbiddenAttribute',
  missingEnglishLanguage = 'missingEnglishLanguage',
  matrixTypeFound = 'matrixTypeFound',
  invalidLanguageCode = 'invalidLanguageCode',
  selectOneNoChoices = 'selectOneNoChoices',
  missingFullnameAttributes = 'missingFullnameAttributes',
}
```

This also fits our other enum conventions better

services/121-service/src/kobo/interfaces/kobo-validation-error.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface KoboValidationErrorBase {
+  type: KoboValidationErrorType;
+  field: string;
+  error: string;
+  solution: string;
+  info?: string;
+  message: string; // Legacy message
```
RubenGeo: What is the reason to keep the message legacy around? Can we not remove it?

services/121-service/src/kobo/interfaces/kobo-validation-error.interface.ts
```diff
@@ -0,0 +1,21 @@
...
+  InvalidChoice = 'invalid_choice',
+  FormConfiguration = 'form_configuration',
+  ForbiddenAttribute = 'forbidden_attribute',
+  MissingEnglishLanguage = 'missing-english-language',
+  MatrixTypeFound = 'matrix-type-found',
+  InvalidLanguageCode = 'invalid-language-code',
+  SelectOneNoChoices = 'select-one-no-choices',
+  MissingFullnameAttributes = 'missing-fullname-attributes',
+}
+export interface KoboValidationErrorBase {
```
RubenGeo: I like this solution with this error object 👍

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -206,22 +202,17 @@ export class KoboValidationService {
...
-
-    return errorMessages;
+    return fspAttributeNames
+      .map((fspAttributeName) =>
+        this.validateSingleFspAttributeType({
+          koboSurveyItems,
+          fspAttributeName,
+        }),
+      )
+      .filter((error): error is KoboValidationErrorBase => error !== undefined);
```
RubenGeo: I personally found the basic `for..of` loop here more readable the the `.map` with a `.filter`. 

What do you think? If you disagree with me feel free to leave to new approach, it's not a hill i want to die on

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -241,7 +232,7 @@ export class KoboValidationService {
     }
 
     return this.validateSurveyItemTypeMatchExpected121Type({
-      attributeName: fspAttributeName,
+      field: fspAttributeName,
```
RubenGeo: I find attributeName here more useful than field as it better matches what we use in the rest of the code base. 

What do you think?

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -254,29 +245,33 @@ export class KoboValidationService {
...
+    if (missingAttributes.length === 0) {
+      return [];
     }
 
-    return errors;
+    return missingAttributes.map((missingAttribute) => ({
+      type: KoboValidationErrorType.MissingFullnameAttributes,
+      field: missingAttribute,
+      error: `Attribute '${missingAttribute}' is missing`,
+      solution: `Add the missing attribute to the Kobo form`,
```
RubenGeo, line 268: Should we still mention in the `info` property that this error is because of the way the program naming convention was setup? Or is that still too confusing to our users?

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -333,69 +338,80 @@ export class KoboValidationService {
...
       (item) => item.name === GenericRegistrationAttributes.scope,
     );
     if (!scopeItem) {
-      return 'Kobo form must contain a scope item if program.enableScope is set to true.';
+      return {
+        type: KoboValidationErrorType.MissingField,
+        field: GenericRegistrationAttributes.scope,
+        error: `Attribute '${GenericRegistrationAttributes.scope}' is missing`,
+        solution:
+          'Add a scope field to the Kobo form (required when program.enableScope is true)',
```
RubenGeo, line 354: Should we refer to this part of the interface here "Use Scope":

Image

 instead of talking about the more technical term program.enableScope

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -425,56 +448,41 @@ export class KoboValidationService {
...
+      .filter((item) => item.name in KOBO_ALLOWED_REGISTRATION_VIEW_ATTRIBUTES)
+      .map(({ name, type }) =>
+        this.validateSurveyItemTypeMatchExpected121Type({
+          field: name,
+          surveyItemType: type,
+          expected121Type: KOBO_ALLOWED_REGISTRATION_VIEW_ATTRIBUTES[name],
+        }),
+      )
+      .filter((error): error is KoboValidationErrorBase => error !== undefined);
   }
```
RubenGeo: I think the old pattern of this function was more readable. Why did you refactor it?

services/121-service/src/kobo/services/kobo.validation.service.ts
```diff
@@ -425,56 +448,41 @@ export class KoboValidationService {
...
+          this.isRegistrationViewAttributeName(surveyItem.name) &&
+          !KOBO_ALLOWED_REGISTRATION_VIEW_ATTRIBUTES[surveyItem.name],
+      )
+      .map((surveyItem) => ({
+        type: KoboValidationErrorType.ForbiddenAttribute as const,
+        field: surveyItem.name,
+        error: `'${surveyItem.name}' is a reserved attribute name and cannot be filled from Kobo`,
+        solution: `Rename the field '${surveyItem.name}' to a non-reserved name`,
+        message: `Attribute '${surveyItem.name}' is a reserved attribute name and cannot be filled from Kobo.`,
+      }));
```
RubenGeo: I also liked to old logic pattern better here. If we want to use filters instead of a basic loop here I think someting like this is clearer

```suggestion
  const isNotScope = (surveyItem: KoboSurveyItemCleaned) =>
    surveyItem.name !== GenericRegistrationAttributes.scope;

  const isForbiddenAttribute = (surveyItem: KoboSurveyItemCleaned) =>
    this.isRegistrationViewAttributeName(surveyItem.name) &&
    !KOBO_ALLOWED_REGISTRATION_VIEW_ATTRIBUTES[surveyItem.name];

  return koboSurveyItems
    .filter(isNotScope)
    .filter(isForbiddenAttribute)
    .map((surveyItem) => ({
      type: KoboValidationErrorType.ForbiddenAttribute as const,
      field: surveyItem.name,
      error: `'${surveyItem.name}' is a reserved attribute name and cannot be filled from Kobo`,
      solution: `Rename the field '${surveyItem.name}' to a non-reserved name`,
      message: `Attribute '${surveyItem.name}' is a reserved attribute name and cannot be filled from Kobo.`,
    }));
```

services/121-service/test/kobo/kobo-load-form-definition.test.ts
```diff
@@ -225,14 +225,9 @@ describe('Import a Kobo form definition', () => {
...
-     "Kobo form definition validation failed:
-     - Missing required FSP attribute 'phoneNumber' for FSP 'SafaricomFsp' in Kobo asset survey.
-     - Missing required FSP attribute 'nationalId' for FSP 'SafaricomFsp' in Kobo asset survey.
-     - Kobo form must contain the following name attributes defined in program.fullnameNamingConvention. However the following attributes are missing: fullName
-     - Kobo form must contain a question with name phoneNumber (should be a text type and country code should be included) or program.allowEmptyPhoneNumber must be set to true.
-     - Invalid Kobo language code: null. Please use https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes"
-     `);
+    expect(linkKoboResponse.body.message).toMatchInlineSnapshot(
+      `"Kobo form definition validation failed"`,
+    );
```
RubenGeo, line 230: I think it make sense to also validate the the new errors part of the reponse and not just the message.

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-error-dialog/kobo-error-dialog.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  templateUrl: './kobo-error-dialog.component.html',
+  styles: ``,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class KoboErrorDialogComponent {
+  readonly errors = input<KoboValidationErrorBase[]>([]);
+  readonly dialogVisible = model(false);
+  readonly tryAgain = output();
+
+  readonly errorTable = computed(() =>
```
RubenGeo: Maybe a more readable pattern could be to make something like this. So we also have type safety when we add new error types in the future.  (untested code)

```
  enum KoboErrorDisplayType {                                                                                                                                                                                                                                                                                                                                             
    table = 'table',                                                                                                                                                                                                                                                                                                                                                    
    formSetting = 'formSetting',
    missingField = 'missingField',
  }

  const koboErrorDisplayTypeMap: Record<KoboValidationErrorType, KoboErrorDisplayType> = {
    [KoboValidationErrorType.MissingField]: KoboErrorDisplayType.MissingField,
    [KoboValidationErrorType.MissingEnglishLanguage]: KoboErrorDisplayType.FormSetting,                                                                                                                                                                                                                                                                                   
    [KoboValidationErrorType.InvalidLanguageCode]: KoboErrorDisplayType.FormSetting,
    [KoboValidationErrorType.TypeMismatch]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                                   
    [KoboValidationErrorType.InvalidChoice]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                                
    [KoboValidationErrorType.FormConfiguration]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                              
    [KoboValidationErrorType.ForbiddenAttribute]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                             
    [KoboValidationErrorType.MatrixTypeFound]: KoboErrorDisplayType.Table,
    [KoboValidationErrorType.SelectOneNoChoices]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                             
    [KoboValidationErrorType.MissingFullnameAttributes]: KoboErrorDisplayType.Table,                                                                                                                                                                                                                                                                                      
  };
                                                                                                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                                          
  
  readonly errorTable = computed(() =>                                                                                                                                                                                                                                                                                                                                    
    this.errors().filter(                                                                                                                                                                                                                                                                                                                                               
      (error) => koboErrorDisplayTypeMap[error.type] === KoboErrorDisplayType.table,
    ),
  );

  readonly formSettingErrors = computed(() =>                                                                                                                                                                                                                                                                                                                             
    this.errors().filter(
      (error) => koboErrorDisplayTypeMap[error.type] === KoboErrorDisplayType.formSetting,                                                                                                                                                                                                                                                                                
    ),                                                                                                                                                                                                                                                                                                                                                                  
  );

  readonly missingFieldErrors = computed(() => {
    const fields = this.errors()
      .filter((error) => koboErrorDisplayTypeMap[error.type] === KoboErrorDisplayType.missingField)                                                                                                                                                                                                                                                                       
      .map((error) => error.field);
    return [...new Set(fields)];                                                                                                                                                                                                                                                                                                                                          
  });                                                                                                                                                                                                                                                                                                                                                                   

```

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-error-dialog/kobo-error-dialog.component.html
```diff
@@ -0,0 +1,162 @@
...
+      </ul>
+    </section>
+  }
+
+  @if (errorTable().length > 0) {
+    <section class="mt-4 mb-2">
+      <div class="mb-4">
+        <h1
+          i18n
+          class="text-base! font-semibold"
```
RubenGeo, line 85: Reviewing html is not my strong suit. But is the use of an `!` an indication we should use a different html tag. Could we not use:

<p class="font-semibold"> here instead. 

What do you think?

( this can be applied to also some other cases in this file)

RubenGeo: It also feels a bit than an h1 is weird here since there is an h3 at the top of this file.

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-error-dialog/kobo-error-dialog.component.html
```diff
@@ -0,0 +1,162 @@
...
+        >
+          Other errors
+        </h1>
+      </div>
+      <p-table
+        [value]="errorTable()"
+        size="small"
+        i18n-size
```
RubenGeo: What does this do? Is it responsible for that this is in the language file? 


Image

RubenGeo (changes requested):
Great stuff! Just a few comments

services/121-service/test/kobo/kobo-load-form-definition.test.ts
RubenGeo: Should we validate the complete body or just a more 'light weight' test. Something like:

```
const { errors } = linkKoboResponse.body;
expect(errors).toEqual(expect.any(Array));
expect(errors.length).toBeGreaterThan(0);

// Spot-check one error has the expected shape
expect(errors).toEqual(
  expect.arrayContaining([
    expect.objectContaining({
      type: expect.any(String),
      field: expect.any(String),
      error: expect.any(String),
      solution: expect.any(String),
    }),
  ]),
```

I we validate the whole body we test 'more' of the response , however the drawback is that if we ever change an error message we have to update both the unit tests and the integration test. What are your thoughts on this?

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-error-dialog/kobo-error-dialog.component.html
nielsVoogt: @RubenGeo Ha, good catch. It's a prop from primeNg table. Removed it.

interfaces/portal/src/app/pages/program-settings-registration-data/components/kobo-error-dialog/kobo-error-dialog.component.html
nielsVoogt: A `section` is a new context, so the h1 is semanticly valid, although I agree that it feels a bit of..
The h3 is just copy-pasted from existing modals, but something we can maybe look into in the future.

services/121-service/src/kobo/services/kobo.validation.service.ts
nielsVoogt: @RubenGeo Agree

nielsVoogt:
@RubenGeo I didn't get to all comments today, reverting PR to draft en will finish monday.

services/121-service/test/kobo/kobo-load-form-definition.test.ts
nielsVoogt: I would think this solution is good, because it enforces a single generic error shape.

services/121-service/src/kobo/services/kobo.validation.service.ts
nielsVoogt: @RubenGeo No idea, this was Claude's doing (and not a improvement at that, weird it touched it at all). Reminder for myself to be more vigilant on certain Copilot changes. Reverted to old.

services/121-service/src/kobo/services/kobo.validation.service.ts
nielsVoogt: @RubenGeo I asked Tal about this, but let's not make this a blocker for now since it's a minor update we can do later.

services/121-service/src/kobo/services/kobo.validation.service.ts
nielsVoogt: @RubenGeo I asked Tal about this, but let's not make this a blocker for now since it's a minor update we can do later.

RubenGeo (dismissed):
Let's go!

RubenGeo (approved):

PR #8355, opened 2026-06-17
Translation sort change

nielsVoogt:
[AB#42857](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/42857)

## Describe your changes

Claude was like... **"Fix the root cause: make extraction deterministic"**

A lot of .xlf conflicts come from unstable ordering/IDs. `ng-extract-i18n-merge` (already in `package.json`) has options like sort (e.g. "stableAppendNew" vs "idAsc") and configurable IDs. Check your `angular.json` extract-i18n config.

If sorting is stableAppendNew, new entries append at the end → two branches both append → conflict. Switching to a fully deterministic sort (e.g. idAsc) means entries land in a stable position regardless of branch, drastically reducing conflicts.

**Pros:**
Reduces conflicts everywhere (local, CI, server) without new infrastructure. Keeps the file tracked and PR-visible.

**Cons:**
Doesn't eliminate conflicts when two branches genuinely change the same string; one big initial re-sort diff.

**Verdict:**
Strong low-effort win, complementary to any other option. **I'd do this regardless.**

## Portal preview-deployment

https://happy-rock-0411d2003-8355.westeurope.3.azurestaticapps.net

### DIFF
--- a/interfaces/portal/angular.json
+++ b/interfaces/portal/angular.json
@@ -183,7 +183,7 @@
             "outputPath": "src/locale",
             "prettyNestedTags": false,
             "resetTranslationState": false,
-            "sort": "stableAppendNew",
+            "sort": "idAsc",
             "sortNestedTagAttributes": true,
             "sourceFile": "messages.xlf",
             "targetFiles": [],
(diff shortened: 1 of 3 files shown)


elwinschmitz (dismissed):
Good idea; And indeed the benefits Claude pointed out seem reasonable and nice and "what we want"...

Maybe it would be necessary to do a "whole clean build" of the file (so it is fully sorted)?

elwinschmitz (dismissed):

RubenGeo (dismissed):

elwinschmitz (approved):

PR #8362, opened 2026-06-19
Eslint rule never throw 500 http exception

RubenGeo:
[AB#42833](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42833) 

## Describe your changes

Custom ESLint rule to prevent throwing HttpException with status 500.

Why?
-  Prevent keeping unexpected failures “unhandled” so NestJS logs full stacktraces.
- HttpException do not trigger the automated internal server error alerts we setup

## Portal preview-deployment

### DIFF
--- a/services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
+++ b/services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
@@ -25,6 +25,20 @@ export default {
     },
   },
   create(context) {
+    function getClassName(node) {
+      if (node.type === 'Identifier') {
+        return node.name;
+      }
+      if (
+        node.type === 'MemberExpression' &&
+        !node.computed &&
+        node.property.type === 'Identifier'
+      ) {
+        return node.property.name;
+      }
+      return null;
+    }
+
     function isInternalServerErrorStatus(node) {
       if (!node) {
         return false;
@@ -35,10 +49,9 @@ export default {
       if (
         node.type === 'MemberExpression' &&
         !node.computed &&
-        node.object.type === 'Identifier' &&
-        node.object.name === 'HttpStatus' &&
         node.property.type === 'Identifier' &&
-        node.property.name === 'INTERNAL_SERVER_ERROR'
+        node.property.name === 'INTERNAL_SERVER_ERROR' &&
+        getClassName(node.object) === 'HttpStatus'
       ) {
         return true;
       }
@@ -51,12 +64,12 @@ export default {
         if (!argument || argument.type !== 'NewExpression') {
           return;
         }
-        const callee = argument.callee;
-        if (callee.type !== 'Identifier') {
+        const className = getClassName(argument.callee);
+        if (!className) {
           return;
         }
 
-        if (callee.name === 'InternalServerErrorException') {
+        if (className === 'InternalServerErrorException') {
           context.report({
             node: argument,
             messageId: 'noInternalServerErrorException',
@@ -65,7 +78,7 @@ export default {
         }
 
         if (
-          callee.name === 'HttpException' &&
+          className === 'HttpException' &&
           isInternalServerErrorStatus(argument.arguments[1])
         ) {
           context.report({
--- a/services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
+++ b/services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
@@ -1,0 +1,79 @@
+/**
+ * Disallow throwing `HttpException` with status 500 / `HttpStatus.INTERNAL_SERVER_ERROR`,
+ * and `InternalServerErrorException`.
+ *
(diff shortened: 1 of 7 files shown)


services/121-service/src/fsp-integrations/reconciliation/safaricom/safaricom-reconciliation.service.ts
```diff
@@ -120,25 +115,16 @@ export class SafaricomReconciliationService {
...
-        newTransactionStatus: TransactionStatusEnum.error,
-        errorMessage: 'Transfer timed out',
-      });
-    } catch (error) {
-      // This should never happen. This way, if it happens, we receive an alert
-      if (error instanceof NotFoundException) {
-        throw new InternalServerErrorException(error.message);
-      }
+    const safaricomTransfer =
+      await this.safaricomTransferScopedRepository.getByOriginatorConversationIdOrThrow(
```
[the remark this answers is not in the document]
RubenGeo: I refactored too much it's now reverted again

services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
```diff
@@ -0,0 +1,79 @@
...
+    schema: [],
+    messages: {
+      noHttpException500:
+        'Do not throw HttpException with status 500. Throw a regular Error instead, so the stacktrace is logged.',
+      noInternalServerErrorException:
+        'Do not throw InternalServerErrorException. Throw a regular Error instead, so the stacktrace is logged.',
+    },
+  },
+  create(context) {
+    function isInternalServerErrorStatus(node) {
```
nameerhqureshi, line 42: Right now, we imports NestJS common utilities as a namespace (e.g., `import * as common from '@nestjs/common';`) and throws `new common.HttpException()`, this rule will bypass it because `callee.type` becomes a `MemberExpression` instead of an `Identifier`.

We can extract a helper function to safely get the class name whether it's called directly or via a namespace.

```
  create(context) {
    // Helper to extract class name from direct calls or namespaced calls
    function getClassName(node) {
      if (node.type === 'Identifier') return node.name;
      if (node.type === 'MemberExpression' && !node.computed && node.property.type === 'Identifier') {
        return node.property.name;
      }
      return null;
    }

    function isInternalServerErrorStatus(node) {
      if (!node) {
        return false;
      }
      if (node.type === 'Literal' && node.value === 500) {
        return true;
      }
      if (
        node.type === 'MemberExpression' &&
        !node.computed &&
        node.property.type === 'Identifier' &&
        node.property.name === 'INTERNAL_SERVER_ERROR'
      ) {
        return true;
      }
      return false;
    }

    return {
      ThrowStatement(node) {
        const argument = node.argument;
        if (!argument || argument.type !== 'NewExpression') {
          return;
        }
        
        const className = getClassName(argument.callee);
        if (!className) {
          return;
        }

        if (className === 'InternalServerErrorException') {
          context.report({
            node: argument,
            messageId: 'noInternalServerErrorException',
          });
          return;
        }

        if (
          className === 'HttpException' &&
          isInternalServerErrorStatus(argument.arguments[1])
        ) {
          context.report({
            node: argument,
            messageId: 'noHttpException500',
          });
        }
      },
    };
  },
```

services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
```diff
@@ -0,0 +1,79 @@
...
+      if (
+        node.type === 'MemberExpression' &&
+        !node.computed &&
+        node.object.type === 'Identifier' &&
+        node.object.name === 'HttpStatus' &&
+        node.property.type === 'Identifier' &&
+        node.property.name === 'INTERNAL_SERVER_ERROR'
+      ) {
+        return true;
+      }
```
nameerhqureshi, line 57: The rule strictly checks if the object name is exactly `HttpStatus`. If we imports it with an alias or uses a namespaced wrapper (e.g. `NestHttpStatus.INTERNAL_SERVER_ERROR`), the rule will be bypassed.

I think `INTERNAL_SERVER_ERROR` is highly unique, we can safely drop the strict object name check and solely rely on the property name matching `INTERNAL_SERVER_ERROR`.

```
      if (
        node.type === 'MemberExpression' &&
        !node.computed &&
        node.property.type === 'Identifier' &&
        node.property.name === 'INTERNAL_SERVER_ERROR'
      ) {
        return true;
      }
```

nameerhqureshi (changes requested):

services/121-service/eslint-custom-rules/no-internal-server-error-http-exception.mjs
RubenGeo: Good catch! Namespace imports like common.HttpStatus.INTERNAL_SERVER_ERROR are now caught too. Kept the HttpStatus check so we don't flag unrelated constants.

nameerhqureshi (approved):

PR #8363, opened 2026-06-19
chore: fail transactions on incomplete FSP config

Yuri-510:
[AB#42823](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42823) 

## Describe your changes

During creation and updating, the `program-fsp-configurations` service now checks if all required properties are set and updates the configuration state to pending or configured accordingly.

The `payments-execution` service now fails transactions where the FSP configuration state is not fully configured.

## Portal preview-deployment

https://happy-rock-0411d2003-8363.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/src/payments/services/payments-execution.service.ts
+++ b/services/121-service/src/payments/services/payments-execution.service.ts
@@ -3,10 +3,10 @@ import { PaginateQuery } from 'nestjs-paginate';
 import { Equal } from 'typeorm';
 
 import { Fsps } from '@121-service/src/fsp-integrations/shared/enum/fsp-name.enum';
-import { TransactionFailureParams } from '@121-service/src/payments/interfaces/transaction-failure-params.interface';
 import { PaymentEvent } from '@121-service/src/payments/payment-events/enums/payment-event.enum';
 import { PaymentEventsService } from '@121-service/src/payments/payment-events/payment-events.service';
 import { PaymentApprovalRepository } from '@121-service/src/payments/repositories/payment-approval.repository';
+import { PaymentsHelperService } from '@121-service/src/payments/services/payments-helper.service';
 import { PaymentsProgressService } from '@121-service/src/payments/services/payments-progress.service';
 import { PaymentsReportingService } from '@121-service/src/payments/services/payments-reporting.service';
 import { TransactionJobsCreationService } from '@121-service/src/payments/services/transaction-jobs-creation.service';
@@ -26,6 +26,7 @@ export class PaymentsExecutionService {
     private readonly transactionsService: TransactionsService,
     private readonly paymentsReportingService: PaymentsReportingService,
     private readonly paymentApprovalRepository: PaymentApprovalRepository,
+    private readonly paymentsHelperService: PaymentsHelperService,
   ) {}
 
   public async startPayment({
@@ -41,49 +42,52 @@ export class PaymentsExecutionService {
       programId,
     });
 
-    try {
-      const notCompletedApprovals = await this.paymentApprovalRepository.count({
-        where: {
-          paymentId: Equal(paymentId),
-          approved: Equal(false),
+    const notCompletedApprovals = await this.paymentApprovalRepository.count({
+      where: {
+        paymentId: Equal(paymentId),
+        approved: Equal(false),
+      },
+    });
+    if (notCompletedApprovals > 0) {
+      throw new HttpException(
+        `Cannot start payment. There are ${notCompletedApprovals} approval(s) to be done for this payment.`,
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+
+    const uniqueFspConfigsForApprovedTransactions =
+      await this.transactionViewScopedRepository.getUniqueProgramFspConfigForApprovedTransactions(
+        {
+          programId,
+          paymentId,
         },
-      });
-      if (notCompletedApprovals > 0) {
-        throw new HttpException(
-          `Cannot start payment. There are ${notCompletedApprovals} approval(s) to be done for this payment.`,
-          HttpStatus.BAD_REQUEST,
-        );
-      }
+      );
+    if (uniqueFspConfigsForApprovedTransactions.length === 0) {
+      throw new HttpException(
+        'No "approved" transactions found for this payment.',
+        HttpStatus.BAD_REQUEST,
+      );
+    }
+    const programFspConfigurationNames = uniqueFspConfigsForApprovedTransactions.map(
+      (config) => config.programFspConfigurationName,
+    );
+    await this.paymentsHelperService.checkFspConfigurationsOrThrow(
+      programId,
+      programFspConfigurationNames,
+    );
 
-      // check that all FSP configurations are still valid
-      const uniqueFspConfigsForApprovedTransactions =
-        await this.transactionViewScopedRepository.getUniqueProgramFspConfigForApprovedTransactions(
-          {
-            programId,
-            paymentId,
-          },
-        );
-      if (uniqueFspConfigsForApprovedTransactions.length === 0) {
-        throw new HttpException(
-          'No "approved" transactions found for this payment.',
-          HttpStatus.BAD_REQUEST,
-        );
-      }
-
+    try {
       await this.paymentEventsService.createEvent({
         paymentId,
         userId,
         type: PaymentEvent.started,
       });
 
-      await this.markTransactionsAsFailedForNotConfiguredFsps({
-        userId,
-        programId,
-        paymentId,
-        status: TransactionStatusEnum.approved,
-      });
-
-      await this.markTransactionsAsFailedForNotIncludedRegistrations({
+      const fspConfigIds = uniqueFspConfigsForApprovedTransactions.map(
+        (p) => p.programFspConfigurationId,
+      );
+      await this.failTransactionsForNotIncludedRegistrations({
+        fspConfigIds,
         userId,
         programId,
         paymentId,
@@ -125,12 +129,14 @@ export class PaymentsExecutionService {
     });
   }
 
-  private async markTransactionsAsFailedForNotIncludedRegistrations({
+  private async failTransactionsForNotIncludedRegistrations({
+    fspConfigIds,
     userId,
     programId,
     paymentId,
     status,
   }: {
+    fspConfigIds: number[];
     userId: number;
     programId: number;
     paymentId: number;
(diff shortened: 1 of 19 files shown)


services/121-service/test/program/manage-program-fsp-configuration.test.ts
```diff
@@ -137,6 +138,101 @@ describe('Manage FSP-configurations', () => {
     expect(getResultConfig).toEqual(result.body);
   });
 
+  it('should save a program FSP-configuration without properties with state "configurationPending"', async () => {
```
RubenGeo, line 141: We now have a lot of tests in this file. And it uses a resetDB before each test, which increases the time it takes to run all of these tests. 

I think the point is now there to refactor this test file so it does not need the reset for every test. I think we can replace it by just creating a new program for every test.

This can also be a seperate PR/ task I think. But would be nice to do in this PBI

services/121-service/test/program/manage-program-fsp-configuration.test.ts
```diff
@@ -137,6 +138,101 @@ describe('Manage FSP-configurations', () => {
...
+        ],
+      };
+    const createResult = await postProgramFspConfiguration({
+      programId: programIdVisa,
+      body: createProgramFspConfigurationDtoIncomplete,
+      accessToken,
+    });
+    expect(createResult.body.state).toBe(
+      FspConfigurationStates.configurationPending,
+    );
```
RubenGeo: The asserts should be in the end beneath the 

`// assert`

unless it's really not possible. But in this case I  think it can be put at the end

services/121-service/src/program-fsp-configurations/program-fsp-configurations.service.ts
```diff
@@ -151,14 +155,22 @@ export class ProgramFspConfigurationsService {
       programFspConfigurationDto,
       programId,
     );
+
+    newConfigEntity.state = this.computeFspConfigurationState({
+      fspName: programFspConfigurationDto.fspName,
+      fspConfigurationProperties: programFspConfigurationDto.properties ?? [],
+    });
```
RubenGeo: I think it's a bit better to calculate the state before the mapper. 

You can than pass the state as param to the mapper and the mapper than adds it to the entity. 

The reason for that is that the mapper else does not return a complete entity (not complete because it does not have state), which is what a reader would expect.

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -70,33 +69,25 @@ export class PaymentsExecutionService {
...
-      const fspConfigIds = uniqueFspConfigsForApprovedTransactions.map(
-        (p) => p.programFspConfigurationId,
-      );
-      await this.markTransactionsAsFailed({
-        fspConfigIds,
+      await this.markTransactionsAsFailedForNotConfiguredFsps({
+        userId,
+        programId,
+        paymentId,
+      });
```
RubenGeo: In the refinement it says:

"When processing transactions of FSPs that are not integrated failed transaction are created"

When reading this PR, and when I saw the the `markTransactionsAsFailedForNotConfiguredFsps` next to the `markTransactionsAsFailedForNotIncludedRegistrations` I thought at first that it made sense to do it before the processing of the queue. 

But when reconsidering it I was think it's better to do during the queue processing:
1. We have to ensure that this check also works when doing a retry
2. It would also ensure that if during a payment and someone remove the fsp config properties a failed transaction is still created and not an internal server error.
3. If in the future we add extra ways to kick of transaction jobs such a check would stil work

Maybe near/in the function `logTransactionJobStart` (or in a renamed version of that function) it would make sense to do this.

What do you think?

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -70,33 +69,25 @@ export class PaymentsExecutionService {
...
+
+      await this.markTransactionsAsFailedForNotIncludedRegistrations({
         userId,
```
RubenGeo, line 91: Not related to this line.

But maybe it's still nice to have a seperate task so that PaymentsHelperService.validateFspSettings use the same logic/error message to determine if fsp configs have the right state.

Which than begs the question again -> 'is it not better to remove the check on create payment' 

I still think it's nice to keep but use the same helper

Let's discuss again, haha

services/121-service/src/program-fsp-configurations/program-fsp-configurations.service.ts
```diff
@@ -272,14 +289,18 @@ export class ProgramFspConfigurationsService {
...
     return ProgramFspConfigurationMapper.mapPropertyEntitiesToDtos(properties);
   }
 
   private async validateAllowedPropertyNames({
     propertyNames,
     fspName,
   }: {
-    propertyNames: string[];
+    propertyNames: FspConfigurationProperties[];
```
RubenGeo, line 321: (y)

services/121-service/src/program-fsp-configurations/program-fsp-configurations.service.ts
```diff
@@ -398,6 +419,57 @@ export class ProgramFspConfigurationsService {
...
+
+    const propertyNames = fspConfigurationProperties.map((p) => p.name) ?? [];
+    const hasAllRequiredProperties = requiredProperties.every((required) =>
+      propertyNames.includes(required),
+    );
+
+    return hasAllRequiredProperties
+      ? FspConfigurationStates.configured
+      : FspConfigurationStates.configurationPending;
+  }
```
RubenGeo, line 473: This function has really nice logic to 'business' logic to unit test, but it's private so it's not possible. 

We could have a seperate `ProgramFspConfigurationsHelperService` like we have in many other places with this function (public) in it, and potentially also move some other functions of this relative large class/file

Can be a seperate optional task/ PR in the PBI , what do you think? @Yuri-510 overengineering or fun?

RubenGeo (changes requested):
Nice work!

services/121-service/test/program/manage-program-fsp-configuration.test.ts
Yuri-510: I created a [separate task](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20178?System.State=Flagged%2CIn%20Progress%2CNew%2CReady%2CReview%2CTo%20Do&workitem=43070) to do in a separate PR

services/121-service/src/program-fsp-configurations/program-fsp-configurations.service.ts
Yuri-510: I always think refactoring is fun :) I created an [item](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20178?System.State=Flagged%2CIn%20Progress%2CNew%2CReady%2CReview%2CTo%20Do&workitem=43071) for this

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20178?System.State=Flagged%2CIn%20Progress%2CNew%2CReady%2CReview%2CTo%20Do&workitem=43074

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: For now, lets keep the early failures in 'start' and 'retry' for user convenience. I'll sketch up a PR for locking FSP configs when a payment is in progress.

Yuri-510: And also expand [this item](https://dev.azure.com/redcrossnl/121%20Platform/_sprints/taskboard/121%20Development%20Team/121%20Platform/Sprint%20178?System.State=Flagged%2CIn%20Progress%2CNew%2CReady%2CReview%2CTo%20Do&workitem=43074) to create a single (something like TransactionSchedulingPreflight) service that can be used during payment creation, payment start/retry and transaction job handling.

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -237,6 +301,12 @@ export class PaymentsExecutionService {
         type: PaymentEvent.retry,
       });
 
+      await this.markTransactionsAsFailedForInvalidPaymentData({
+        userId,
+        programId,
+        paymentId,
+      });
```
RubenGeo: In a retry scenario we should not fail transactions if the registration is on status included.

Because 

1. Failed transactions also count towards a registration's total transaction count
2. When a registration total transaction count == maxPayments a registration moves to status 'completed'
3. Therefore a try whould not be possible on those registrations

RubenGeo: I would remove markTransactionsAsFailedForInvalidPaymentData and:

in start payment flow call both:
- markTransactionsAsFailedForNotConfiguredFsps
- markTransactionsAsFailedForNotIncludedRegistrations


in retry flow just call: 
- markTransactionsAsFailedForNotConfiguredFsps

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -153,29 +158,88 @@ export class PaymentsExecutionService {
...
-        ),
+        errorMessages: new Map(transactionIds.map((id) => [id, errorMessage])),
       });
     }
   }
 
+  private mapTransactionIdsToFspConfigId({
+    transactions,
+  }: {
+    transactions: TransactionViewEntity[];
```
RubenGeo: Try to avoid passing complete entities between functions see B.4 of https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#b-nestjs-module-implementation.

The reasoning behind that is that is that this forces callers to construct full entity objects even when only a few fields are needed. Also for a reader it's easier to understand which fields are actually needed for that function. 

This comment also applies to a few other places

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -153,29 +158,88 @@ export class PaymentsExecutionService {
...
+    userId: number;
+    programId: number;
+    paymentId: number;
+  }) {
+    const transactionsToFail =
+      await this.transactionViewScopedRepository.getByStatusOfNonConfiguredFsps(
+        {
+          programId,
+          paymentId,
+          status: TransactionStatusEnum.approved,
```
RubenGeo: I don't think we should filter on status 'approved' here. Because on a retry these transaction are not on status approved but on status 'error'.

RubenGeo: Also a retry can done on a specific sub-set of transactionIds, so you could only retry a few transactions. but this function (getByStatusOfNonConfiguredFsps) queries all transactions of the whole payment and not just the subset. 

I think these are reasons my instinct told me to mark transaction as failed lower in the tree, than there as there are than less conditions to take into account ;)

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -153,29 +158,88 @@ export class PaymentsExecutionService {
...
+        },
       );
-      if (fspConfigTransactions.length === 0) {
-        continue;
-      }
+    await this.failTransactions({
+      transactions: transactionsToFail,
+      userId,
+      errorMessage: 'FSP configuration is not fully configured',
```
RubenGeo: Should we have the current FSPName and fspconfig.'state' in this error message? Or is that unneeded complexity for not much benefit?

RubenGeo (changes requested):

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: The FSP name is already in the table next to the reason of failure in the transaction list, so that would make this code unnecessarily complex imo.

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: Resolved because we now do early throws instead of failing transactions in the start- and retry payment process.

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: Resolved because we now do early throws instead of failing transactions in the start- and retry payment process.

services/121-service/src/payments/services/payments-helper.service.ts
```diff
@@ -40,33 +39,14 @@ export class PaymentsHelperService {
       relations: ['properties'],
```
RubenGeo: We don't need the relations here anymore

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -221,6 +221,14 @@ export class PaymentsExecutionService {
         inputReferenceIds: referenceIds,
       });
 
+      const programFspConfigurationNames = transactionDetails.map(
+        (t) => t.programFspConfigurationName,
+      );
+      await this.paymentsHelperService.checkFspConfigurationsOrThrow(
+        programId,
+        programFspConfigurationNames,
+      );
```
RubenGeo, line 225: I think we should also have unit or integration test that tests whether this check happens on a retry checkFspConfigurationsOrThrow

services/121-service/src/payments/services/payments-execution.service.ts
```diff
@@ -38,65 +38,62 @@ export class PaymentsExecutionService {
...
       const fspConfigIds = uniqueFspConfigsForApprovedTransactions.map(
         (p) => p.programFspConfigurationId,
       );
-      await this.markTransactionsAsFailed({
+      await this.failTransactionsForNotIncludedRegistrations({
         fspConfigIds,
         userId,
         programId,
         paymentId,
+        status: TransactionStatusEnum.approved,
```
RubenGeo: I don't think we need this extra status param anymore, now that we went for a different approach

services/121-service/src/payments/services/payments-helper.service.ts
```diff
@@ -16,22 +16,21 @@ export class PaymentsHelperService {
...
-      throw new HttpException(
-        `${errorMessages.join(', ')}`,
-        HttpStatus.BAD_REQUEST,
-      );
+      throw new HttpException(errorMessages.join(', '), HttpStatus.BAD_REQUEST);
     }
   }
 
-  private async validateFspSettings(
+  private async validateFspConfiguration(
```
RubenGeo:
```suggestion
  private async getFspConfigurationErrorMessage(
```

services/121-service/src/payments/services/payments-execution.service.ts
RubenGeo: Same goes for `startPayment` i think

RubenGeo (changes requested):
Nice, this is the better approach :)

(The PR description also needs an update)

services/121-service/src/payments/services/payments-execution.service.ts
Yuri-510: Added integration tests for both retry and create. `startPayment` was already covered.

RubenGeo (dismissed):

RubenGeo (approved):

PR #8386, opened 2026-06-26
feat: Update permission maintenance service

nameerhqureshi:
[AB#42961](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42961) 

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/shared/services/permission-maintenance.service.ts
+++ b/services/121-service/src/shared/services/permission-maintenance.service.ts
@@ -1,10 +1,12 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository } from 'typeorm';
+import { Equal, Repository } from 'typeorm';
 
+import { PostgresStatusCodes } from '@121-service/src/shared/enum/postgres-status-codes.enum';
 import { PermissionEntity } from '@121-service/src/user/entities/permissions.entity';
 import { UserRoleEntity } from '@121-service/src/user/entities/user-role.entity';
 import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
+import { DefaultUserRole } from '@121-service/src/user/enum/user-role.enum';
 
 @Injectable()
 export class PermissionMaintenanceService {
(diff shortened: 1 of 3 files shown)


services/121-service/src/shared/services/permission-maintenance.service.ts
```diff
@@ -47,4 +97,78 @@ export class PermissionMaintenanceService {
...
+    if (adminPermissionNames.has(permissionName)) {
+      return;
+    }
+
+    await this.userRoleRepository.query(
+      `
+        INSERT INTO "121-service"."user_role_permissions_permission" ("userRoleId", "permissionId")
+        VALUES ($1, $2)
+        ON CONFLICT ("userRoleId", "permissionId") DO NOTHING
+      `,
```
RubenGeo: We try to avoid raw queries in service code, as they are sensitive to changes in table names or column names. 
 
Our hierarchy for preferred ways to access data in the database:  

1. First, use TypeORM repository functions like find, findOne, save, etc. 
2. If that's not possible, use the QueryBuilder
3. Only if neither of those work, use raw SQL

We use raw queries in migrations as those are only run once on production and not all the time. And conversely, using TypeORM entities and repos in migrations can break when entities change after the migration was written. So there it's the other way around

In this case I think we can get by with option 1.

services/121-service/src/shared/services/permission-maintenance.service.ts
```diff
@@ -1,10 +1,11 @@
...
 import { InjectRepository } from '@nestjs/typeorm';
-import { Repository } from 'typeorm';
+import { Equal, Repository } from 'typeorm';
 
 import { PermissionEntity } from '@121-service/src/user/entities/permissions.entity';
 import { UserRoleEntity } from '@121-service/src/user/entities/user-role.entity';
 import { PermissionEnum } from '@121-service/src/user/enum/permission.enum';
+import { DefaultUserRole } from '@121-service/src/user/enum/user-role.enum';
 
 @Injectable()
```
RubenGeo, line 11: Great idea to add all these safeguards, for keeping our permission up to date!

Might be worth sharing this in a short teams post when it's merged

services/121-service/src/shared/services/permission-maintenance.service.ts
```diff
@@ -13,6 +14,55 @@ export class PermissionMaintenanceService {
...
+    );
+
+    for (const permissionName of supportedPermissions) {
+      const permissionEntity =
+        permissionsByName.get(permissionName) ??
+        (await this.ensurePermissionExists({
+          permissionName,
+          allPermissions,
+        }));
+      permissionsByName.set(permissionName, permissionEntity);
```
RubenGeo:
```suggestion
```
I don't think we need this line. Or am I missing something?

services/121-service/src/shared/services/permission-maintenance.service.ts
```diff
@@ -13,6 +14,55 @@ export class PermissionMaintenanceService {
...
+    }
+
+    const adminPermissionNames = new Set(
+      adminRole.permissions.map((permission) => permission.name),
+    );
+
+    for (const permissionName of supportedPermissions) {
+      const permissionEntity =
+        permissionsByName.get(permissionName) ??
+        (await this.ensurePermissionExists({
```
RubenGeo: We first call syncSupportedPermissions in the app module and then we call syncAdminRolePermissions in the app module. 

That means that when we run syncAdminRolePermissions all permissions should already exists. So I think ensurePermissionExists should not be called here.

Also from a single responsibility point of view it makes more sense that syncSupportedPermissions should add the permissions entities in the DB while syncAdminRolePermissions only adds existing permissions to to the adminRole

services/121-service/src/shared/services/permission-maintenance.service.ts
```diff
@@ -47,4 +97,78 @@ export class PermissionMaintenanceService {
       removedPermissions,
     );
   }
+
+  private async ensurePermissionExists({
```
RubenGeo:
```suggestion
  private async getOrCreatePermission({
```
I think this makes it clearer that this function returns the entity

RubenGeo (changes requested):

services/121-service/src/shared/services/permission-maintenance.service.ts
nameerhqureshi: Yes I will notify team once its merged

RubenGeo (approved):
Nice :)

PR #8407, opened 2026-06-30
chore: set Multer limits for uploading files to mitigate DoS attacks

Yuri-510:
[AB#43168](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43168) 

## Describe your changes

This PR sets 'sensible' file size/ amount limits to mitigate possible DoS attacks.

It introduces a single source of truth for CSV/Excel row limits for both Multer - and logic level limiting.

It also adds some tests to assert that Multer is handling the limits.

See: https://github.com/expressjs/multer#security

## Portal preview-deployment


elwinschmitz (dismissed):

services/121-service/src/programs/program-attachments/program-attachments.controller.ts
```diff
@@ -62,7 +63,9 @@ export class ProgramAttachmentsController {
     description: 'Post attachments to a program',
   })
   @Post('programs/:programId/attachments')
-  @UseInterceptors(FileInterceptor('file'))
+  @UseInterceptors(
+    FileInterceptor('file', { limits: ATTACHMENT_FILE_UPLOAD_LIMITS }),
+  )
   public async createProgramAttachment(
     @UploadedFile(
```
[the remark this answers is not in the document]
elwinschmitz: This looks like a classic "1 Megabyte is 1000 bytes" conversion error/misunderstanding.
Sounds good to align indeed.

services/121-service/src/shared/file-upload-limits.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const BYTES_PER_MEGABYTE = 1024 * 1024;
+const MAX_FILES_PER_UPLOAD = 1;
+const MAX_TEXT_FIELDS_PER_UPLOAD = 5;
+
+export const CSV_FILE_UPLOAD_LIMITS: MulterModuleOptions['limits'] = {
+  fileSize: 50 * BYTES_PER_MEGABYTE,
+  files: MAX_FILES_PER_UPLOAD,
+  fields: MAX_TEXT_FIELDS_PER_UPLOAD,
+};
```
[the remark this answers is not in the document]
elwinschmitz: Would be an interesting puzzle to test this in a nice way.. (generating a 101Mb sized file from random data or something?)

Trying to upload more then 1-file-at-a-time should be easily doable though...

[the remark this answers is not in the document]
Yuri-510: This is already done in the `AddAttachmentsOnProgramLevel` e2e test.

services/121-service/src/programs/program-attachments/program-attachments.controller.ts
```diff
@@ -1,31 +1,29 @@
...
+    Patch,
+    Post,
+    Req,
+    Res,
+    UploadedFile,
+    UseGuards,
+    UseInterceptors,
 } from '@nestjs/common';
 import { FileInterceptor } from '@nestjs/platform-express';
 import {
```
RubenGeo, line 19: Weird linting in the diff

services/121-service/src/shared/file-upload-limits.ts
```diff
@@ -0,0 +1,51 @@
...
+type FileUploadLimits = NonNullable<MulterModuleOptions['limits']>;
+
+// Limits for multipart/form-data uploads, passed to multer (via FileInterceptor)
+// to help protect against denial-of-service (DoS) attacks.
+// See: https://github.com/expressjs/multer#security
+
+const BYTES_PER_MEGABYTE = 1024 * 1024;
+const MAX_FILES_PER_UPLOAD = 1;
+const MAX_TEXT_FIELDS_PER_UPLOAD = 5;
+const BYTES_PER_ALLOWED_ROW = 1024;
```
RubenGeo: This may be too tight. 

Quote from claude:
```
ASCII text (Latin letters, digits, commas): 1 byte/char → ~1,024 characters per row, including delimiters, quotes, and the newline

So for a registration import row with, say, 20 columns, that's an average budget of ~50 ASCII characters per field 

```

Sometimes programs have aprox 100 questions so maybe we should increase this limit 10 fold to be on the safe side. Or do you think this may be too large and it will impact performance too much?

services/121-service/src/registration/controllers/registrations.controller.ts
```diff
@@ -177,7 +185,11 @@ export class RegistrationsController {
   @Patch('programs/:programId/registrations')
   @ApiConsumes('multipart/form-data')
   @ApiBody(FILE_UPLOAD_WITH_REASON_API_FORMAT)
-  @UseInterceptors(FileInterceptor('file'))
+  @UseInterceptors(
+    FileInterceptor('file', {
+      limits: REGISTRATION_BULK_PATCH_CSV_FILE_UPLOAD_LIMITS,
```
RubenGeo, line 192: It would be nice to enforce with typing or eslint that we always use the same config file. So in the future we do not forget that it exists

services/121-service/src/shared/file-upload-limits.ts
```diff
@@ -0,0 +1,51 @@
...
+
+type FileUploadLimits = NonNullable<MulterModuleOptions['limits']>;
+
+// Limits for multipart/form-data uploads, passed to multer (via FileInterceptor)
+// to help protect against denial-of-service (DoS) attacks.
+// See: https://github.com/expressjs/multer#security
+
+const BYTES_PER_MEGABYTE = 1024 * 1024;
+const MAX_FILES_PER_UPLOAD = 1;
+const MAX_TEXT_FIELDS_PER_UPLOAD = 5;
```
RubenGeo: I think it better to rely on the DTO for validating the fields of a body, as we can use that on any body validation and not just body's that are uploaded besides a file

services/121-service/test/registrations/import-registration.test.ts
```diff
@@ -38,6 +40,23 @@ import {
...
   let accessToken: string;
 
+  it('should return an error when the import file exceeds the upload size limit', async () => {
+    // Arrange
+    await resetDB({ seedScript: SeedScript.nlrcMultiple });
+    accessToken = await getAccessToken();
+    const fileSizeLimit = REGISTRATION_IMPORT_CSV_FILE_UPLOAD_LIMITS.fileSize!;
+
+    // Act
+    const response = await getServer()
```
RubenGeo: There is a helper function to make this API request. It does not accept a buffer.alloc however.

I do think it would probably be better to extend that function with an optional parameter. So if we change the api path in the future it will be easier to update. 


The same comment applies to `excel-reconciliation`

RubenGeo (changes requested):
Great initiative to centralize this

services/121-service/src/shared/file-upload-limits.ts
Yuri-510: I think so too. I removed that limit.

services/121-service/src/registration/controllers/registrations.controller.ts
Yuri-510: I created a wrapper that enforces limits and an ESLint rule that gets angry when trying to use the "normal" interceptor.

services/121-service/src/shared/file-upload-limits.ts
Yuri-510: I think we're safe with 10mb per row.

services/121-service/test/registrations/import-registration.test.ts
Yuri-510: For now I created more explicit and clean helpers. I think the optional parameter is a bit messy and I think our future selves will not be hindered much by this.

RubenGeo (approved):

PR #8431, opened 2026-07-02
 New fsp flow: backend

nielsVoogt:
[AB#43252](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43252) 

## Describe your changes

Backend for new FSP flow

(https://github.com/global-121/121-platform/blob/main/services/121-service/README.md#api)

## Portal preview-deployment


services/121-service/swagger.json
```diff
@@ -636,6 +636,20 @@
       "programId"
     ]
   },
+  {
```
RubenGeo, line 639: You are adding 2 endpoints in this PR:

1. POST /fsp-configurations/fsps which and creates a config for each, throwing 409 CONFLICT if one already exists.
2.  PATCH /fsp-configurations/fsps which overwrites the whole set of fsp configuration. It deletes  deleting and creates the ones that are missing. A PATCH request should be a Partial update. So I think this endpoint should actually be a PUT as it is an overwrite.

I think however that for the our use case we don't actually need the seperate POST, because we don't really need the "fail if it already exists" behavior. So I think we only need the PUT endpoint.

RubenGeo: I think we can than also use the API path `fsp-configurations` (without fsps), so it follow to more conventional API pattern, and than the endpoint can accept and array of object rather than an array of strings. Something like this
```
[
 {
name: 'debitCards'
fspName: 'Intersolve-visa'
label: 'Debit card'
}
]
```

Name and/or label could potentially be optional.

RubenGeo (changes requested):

services/121-service/swagger.json
Yuri-510: I squashed the two endpoints to one PUT endpoint. 

Accepting an array of config objects however, would make the service method unnecessarily bloated. The only thing the frontend has access to is FSP name and label. FSP name we have and label is already automatically set to default english with FSP name.

RubenGeo:
```
Accepting an array of config objects however, would make the service method unnecessarily bloated. The only thing the frontend has access to is FSP name and label. FSP name we have and label is already automatically set to default english with FSP name.
```

Making the bodies match a bit more between the POST, GET, PATCH and PUT would make our API a bit more resource-focused (REST) over use-case-focused. Which is better practice.

However considering the time we spend on this and the fact that the GET endpoint already returns a different body (because it also return the properties). I think it's fine to merge like this

RubenGeo (approved):

PR #8437, opened 2026-07-02
test: add checks on create program for users without permission

arsforza:
[AB#42719](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42719) 

- [ ] I have checked [the list of integrations](https://github.com/global-121/121-platform/wiki/Integrations-with-external-systems#api-endpoints-used-by-external-systems) with the 121 API for [changed endpoints](https://github.com/global-121/121-platform/blob/main/services/121-service/README.md#api)

## Portal preview-deployment


services/121-service/test/program/create-program.test.ts
```diff
@@ -242,4 +244,79 @@ describe('Create program', () => {
     // A new program should not have been created
     expect(getProgramResponse.statusCode).toBe(HttpStatus.NOT_FOUND);
   });
+
+  it('should not be able to post a program without correct permissions', async () => {
+    // Arrange
+    const users = [
```
elwinschmitz: This would be a perfect candidate for the [`it.each`-pattern](https://devdocs.io/jest/api#each), right? :)

Also, what's up with the snapshot-file? Was it added by accident?

And; Also, wouldn't it be 'enough' to try and `POST` only a single program? That would save some time for the test-runs(s), right?

elwinschmitz (changes requested):

services/121-service/test/program/create-program.test.ts
```diff
@@ -242,4 +291,33 @@ describe('Create program', () => {
     // A new program should not have been created
     expect(getProgramResponse.statusCode).toBe(HttpStatus.NOT_FOUND);
   });
+
+  it.each(usersWithoutPermission)(
```
elwinschmitz: I think it would be fine (better even) to include the array here, in-line.
As this set of users, these "users-without-permission", are only relevant _for this test_... Not necessarily for the other test-scenarios in this file....

And, naming-wise... They don't technically "lack the correct permission", they "miss a User-entity flag/property". ;)
The whole "creating a program" is kind-of the exception-to-the-rule regarding our "permissions-system"...

services/121-service/test/program/create-program.test.ts
```diff
@@ -242,4 +291,33 @@ describe('Create program', () => {
...
+
+      await logoutUser(accessToken);
+      accessToken = await getAccessToken(email, password);
+      if (!email || !password) {
+        throw new Error(
+          'Missing USERCONFIG_121_SERVICE_* user credentials in env; required for create-program permission test.',
+        );
+      }
+
+      for (const seedProgram of seedPrograms) {
```
elwinschmitz: Maybe only trying 1 program, could safe some time in these tests... (and another loop in the code)

elwinschmitz (dismissed):
Good. The comments are optional. ;) (But would still be nice)

elwinschmitz (approved):

PR #8438, opened 2026-07-02
feature: backend for duplicating program

Yuri-510:
## Describe your changes

Added a helper that clones any entity and relations according to a configuration.

Whenever a property is added to the program entity, the program duplication configuration will error on compile time; this configuration must be explicitly maintained alongside the program entity.

**note**
The duplication helper works like copying a folder and its subfolders: it walks parent → child → grandchild and copies each down the tree.

Threshold approvers don't fit that shape right now. An approver link is a sideways pointer between two things that both get copied — an aidworker-assignment row points at a threshold row. They're siblings, not parent-and-child and therefore approvers cannot be duplicated with the helper. Hence the explicit threshold duplication method in the threshold service.

## Portal preview-deployment

https://happy-rock-0411d2003-8438.westeurope.3.azurestaticapps.net


services/121-service/src/programs/program-duplication.const.ts
```diff
@@ -0,0 +1,46 @@
+import { ProgramEntity } from "@121-service/src/programs/entities/program.entity";
+
+export const propertiesToDuplicate: Record<keyof ProgramEntity, boolean> = {
```
RubenGeo: This approach is really nice and readable

services/121-service/src/programs/programs.controller.ts
```diff
@@ -153,10 +163,21 @@ You can also leave the body empty.`,
...
     koboAssetId: string,
 
+    @Query('copyFromProgramId', new ParseIntPipe({ optional: true }))
+    copyFromProgramId: number | undefined,
+
     @Req() req: ScopedUserRequest,
   ): Promise<ProgramEntity> {
     const userId = RequestHelper.getUserId(req);
 
+    if (copyFromProgramId) {
```
RubenGeo, line 176: Should we throw if both copyFromProgramId and the body is set? 

Not sure if this relevant/too detailed yet for this POC, but something to consider on the actual implementation

services/121-service/src/programs/programs.service.ts
```diff
@@ -238,6 +241,51 @@ export class ProgramService {
...
+        manager: queryRunner.manager,
+        entity: ProgramEntity,
+        id: copyFromProgramId,
+        propertiesToDuplicate,
+        overrides,
+      });
+
+      await queryRunner.commitTransaction();
+    } catch (err) {
+      await queryRunner.rollbackTransaction();
```
RubenGeo: Should the whole open close transaction be part of the duplicateEntity function?

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,263 @@
...
+  FindOptionsWhere,
+  ObjectLiteral,
+} from 'typeorm';
+
+import { InverseRelationWithJoinColumnsInterface } from '@121-service/src/utils/entity-duplication/interfaces/inverse-relation-with-join-columns.interface';
+import {
+  RelationWithInverseInterface,
+} from '@121-service/src/utils/entity-duplication/interfaces/relation-with-inverse.interface';
+
+export async function duplicateEntity<T extends ObjectLiteral>({
```
RubenGeo: In general the functions in this file look good.

They are just a bit more abstract than most of our code since they are generic helpers and not really related to actual business. (Which makes a lot of sense in this case)

I'll leave a more detailed review on this for the none POC version. 

But in general I agree with this approach from an architectural point of view. (Like that these are plain helper functions instead of a nest.js intectable service)

Unless you have specific question

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,263 @@
...
+}: {
+  metadata: EntityMetadata;
+  source: ObjectLiteral;
+  overrides: Record<string, unknown>;
+  propertiesToDuplicate?: Record<string, boolean>;
+  excludeForeignKeyColumns?: boolean;
+}): Record<string, unknown> {
+  const result: Record<string, unknown> = {};
+  for (const column of metadata.columns) {
+    if (isPrimaryOrAuditColumn(column)) {
```
RubenGeo, line 108: `id` and such are also listed in the config so maybe isPrimaryOrAuditColumn is not needed

Yuri-510: In the program config they are, but that is not a guarantee for other future entity duplication.

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      newProgramId,
+      accessToken,
+    );
+    expect(persistedProgramResponse.statusCode).toBe(HttpStatus.OK);
+    expect(persistedProgramResponse.body.id).toBe(newProgramId);
+    expect(persistedProgramResponse.body.titlePortal).toStrictEqual(
+      sourceProgram.titlePortal,
+    );
+    expect(persistedProgramResponse.body.currency).toBe(sourceProgram.currency);
+  });
```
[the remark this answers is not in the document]
Yuri-510: Added a test for asserting FSP config relation is also duplicated

services/121-service/src/programs/programs.service.ts
Yuri-510: Ah yes, that is quite nice.

services/121-service/src/programs/programs.controller.ts
Yuri-510: The body can contain overwrites for the duplicate program.

services/121-service/src/programs/program-duplication.const.ts
```diff
@@ -0,0 +1,46 @@
+import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
+
+export const propertiesToDuplicate: Record<keyof ProgramEntity, boolean> = {
+  // Primary and audit columns
```
RubenGeo: I think it's missing a copy of grand-child properties:
- So program-fsp-configurtion-properties
- And also approverThresholds 

This may make this whole thing more complicated :) 


Maybe we can do the configuration it in a similar way as Typeorm find queries work where you can select nested relations

services/121-service/src/programs/program-duplication.const.ts
```diff
@@ -0,0 +1,46 @@
...
+  endDate: true,
+  currency: true,
+  distributionFrequency: true,
+  distributionDuration: true,
+  fixedTransferValue: true,
+  paymentAmountMultiplierFormula: true,
+  targetNrRegistrations: true,
+  description: true,
+  validation: true,
+  fullnameNamingConvention: true,
```
RubenGeo: fullnameNamingConvention and messageTemplates

The copied naming convention and template placeholders can reference attributes that don't exist on the new program.

We can definitely skip messageTemplates as I think it's only used in the Netherlands and in the drc by an INGO. 

Maybe we should refrain from copying these for now? And put fullnameNamingConvention on: ['fullName'] for now? We should probably discuss this with somebody from the business

services/121-service/src/programs/programs.controller.ts
```diff
@@ -153,10 +163,30 @@ You can also leave the body empty.`,
...
+        { skipMissingProperties: true },
+      );
+
+      if (duplicationErrors.length > 0) {
+        throw new HttpException(duplicationErrors, HttpStatus.BAD_REQUEST);
+      }
+
+      return this.programService.duplicateProgram({
+        copyFromProgramId,
+        overrides: programData,
```
RubenGeo: I think the portal will only override the title for now. I am wondering if having the generic overrides property in `duplicateEntity` function add to much complexity to the overall flow. 

We could also do the the update of the portal title in another PATCH API call? 

What do you think?

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,277 @@
+import { HttpException, HttpStatus } from '@nestjs/common';
+import {
+  DataSource,
```
RubenGeo: Does it make sense to have some unit tests for the functions in this file?

services/121-service/src/utils/entity-duplication/interfaces/relation-with-inverse.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { InverseRelationWithJoinColumnsInterface } from '@121-service/src/utils/entity-duplication/interfaces/inverse-relation-with-join-columns.interface';
+
+export interface RelationWithInverseInterface {
```
RubenGeo: According to claude we can use an already defined type from typeorm. Which seems to work for me locally.

Your type is more specific but it's maybe also less type safe than using the actual type from typeorm I think

```
import { RelationMetadata } from 'typeorm/metadata/RelationMetadata';
```

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,277 @@
...
+    const primaryColumn = metadata.primaryColumns[0].propertyName;
+    const source = await repository.findOne({
+      where: { [primaryColumn]: Equal(id) } as FindOptionsWhere<T>,
+      relations: buildRelationLoadTree<T>({ metadata, relationNamesToDuplicate }),
+    });
+    if (!source) {
+      throw new HttpException(
+        { errors: `No ${metadata.name} found with id ${id}` },
+        HttpStatus.NOT_FOUND,
+      );
```
RubenGeo:
```
if (!source) {
      throw new HttpException(
        { errors: `No ${metadata.name} found with id ${id}` },
        HttpStatus.NOT_FOUND,
      );
    } 
```

I think this 'helper' function should not be concerned about HttpSpecifics. 

I think this http throw should happen in the programService. 

So I would:
- Check if the program exists in the 121-service, throw if 404 if it does not exist
- Change findOne to findOneOrFail

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,97 @@
...
+describe('Duplicate program', () => {
+  const copyFromProgramId = 2;
+  let accessToken: string;
+
+  beforeEach(async () => {
+    await resetDB({ seedScript: SeedScript.nlrcMultiple });
+    accessToken = await getAccessToken();
+  });
+
+  it('should duplicate a program and persist it', async () => {
```
RubenGeo: I think both tests use the same code path, so I think it can be one code path with multiple valdiations:
- The root colums
- Configured fsp configs
  - fsp config properties
- Configured team members
- Configured approver thresholds

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    );
+
+    // Assert
+    expect(duplicateResponse.statusCode).toBe(HttpStatus.CREATED);
+    expect(newProgramId).toBeDefined();
+    expect(newProgramId).not.toBe(copyFromProgramId);
+    expect(persistedProgramResponse.body.titlePortal).toStrictEqual(
+      sourceProgram.titlePortal,
+    );
+    expect(persistedProgramResponse.body.currency).toBe(sourceProgram.currency);
```
RubenGeo: I think we should validate some relevant columns that copied and some that should explicitly not be copied (which are dangerous to copy)

```
// Columns that SHOULD be copied
expect(duplicated.titlePortal).toStrictEqual(source.titlePortal);
expect(duplicated.currency).toBe(source.currency);
expect(duplicated.budget).toBe(source.budget);
// ... cover each meaningful column

// Things that SHOULD NOT be copied
expect(duplicated.id).not.toBe(source.id);
expect(duplicated.registrations).toHaveLength(0);
expect(duplicated.payments).toHaveLength(0);
```

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+    tree[relationName] =
+      Object.keys(manyToManySubRelations).length > 0
+        ? manyToManySubRelations
+        : true;
+  }
+  return tree as FindOptionsRelations<T>;
+}
+
+function cloneColumns({
```
RubenGeo:
```suggestion
function pickValuesFromSource({
```

Or `pickValues`?

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,277 @@
...
+    childColumns[parentForeignKeyProperty] = newParentId;
+
+    // Re-link many-to-many relations (e.g. roles) instead of cloning targets.
+    for (const childRelation of childMetadata.relations) {
+      if (childRelation.isManyToMany && child[childRelation.propertyName]) {
+        childColumns[childRelation.propertyName] =
+          child[childRelation.propertyName];
+      }
+    }
+    return childRepository.create(childColumns as DeepPartial<ObjectLiteral>);
```
RubenGeo: I think this function would be more readable if we split off selected part in a seperate function.

services/121-service/src/programs/programs.controller.ts
Yuri-510: I think overrides/overall-complexity ratio is very low. So we can keep it.

services/121-service/src/programs/program-duplication.const.ts
Yuri-510: Leaving this open till we are set on the implementation.

services/121-service/src/programs/program-duplication.const.ts
Yuri-510: Everything to be copied is decided via the config file now.

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
Yuri-510: I added unit tests. It also helps understanding the helper better.

services/121-service/test/program/duplicate-program.test.ts
Yuri-510: I added helpers that loop over configuration to assert everything that should be copied, is copied

services/121-service/src/programs/programs.service.ts
```diff
@@ -238,6 +241,34 @@ export class ProgramService {
...
+      );
+    }
+
+    const newProgram = await duplicateEntity({
+      dataSource: this.dataSource,
+      entity: ProgramEntity,
+      id: copyFromProgramId,
+      propertiesToDuplicate,
+      overrides,
+    });
```
RubenGeo, line 280: I see that in 83536737b9  it was removed that the user that is is copying is also assigned as the Admin role for that program. 

I think it does make sense to keep that. Else you may end up in a situation where someone copies a program of which he only has program.read permission, and than nobody can assign new users in that program. 

(also a test to cover this scenario)

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,629 @@
...
+  }
+
+  return manyToManySubRelations;
+}
+
+// ---------------------------------------------------------------------------
+// Copy-columns phase: copy the entity's own column values.
+// ---------------------------------------------------------------------------
+
+export function copyColumnValues({
```
RubenGeo, line 96: This function has 2 optional properties which are sort of a smell to suggest it can be more readable if split up in multiple functions:

```
copyRootColumnValues()
   // 1) filters out properties that are not selected in propertiesToDuplicate
  // 2) runs copyColumnValues()
  // 3) over writes the overrides

copyChildrenColumnValues
  // 1) Skips nullabele FKs
  // 2) runs copyColumnValues()

pickColumns() () // contains all generic functionalityshared core: skip primary/audit


```

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+): RelationMetadata {
+  return {
+    isOneToMany: false,
+    isOneToOne: false,
+    inverseRelation: undefined,
+    ...overrides,
+     
+  } as any;
+}
+
```
RubenGeo, line 46: Functions in this file are exported solely because the spec file imports them directly, they are not used anywhere else in production code.

This feels a bit as the same anti-pattern as making a NestJS service method public just so a unit test can call it. 

It leaks implementation details into the public API and couples tests to internal structure, making refactoring harder. However I do think there are practical drawbacks of testing everything through `duplicateEntity` it would require you to mock a lot and it would maybe make the tests a bit readable. 

What do you think?

services/121-service/src/programs/program-duplication.const.ts
RubenGeo: Let's discuss with erna in the office tomorrow

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,213 @@
...
+        programId: duplicatedProgramId,
+        accessToken,
+      })
+    ).body;
+    const duplicatedThresholds = (
+      await getProgramApprovalThresholds({
+        programId: duplicatedProgramId,
+        accessToken,
+      })
+    ).body;
```
RubenGeo, line 132: This is part of the assert. The act should contain only the 'process /action' we are testing

services/121-service/src/programs/programs.service.ts
```diff
@@ -238,6 +241,34 @@ export class ProgramService {
...
+  }): Promise<ProgramEntity> {
+    const programExists = await this.programRepository.exists({
+      where: { id: Equal(copyFromProgramId) },
+    });
+    if (!programExists) {
+      throw new HttpException(
+        { errors: `No program found with id ${copyFromProgramId}` },
+        HttpStatus.NOT_FOUND,
+      );
+    }
```
RubenGeo: This could also use `findProgramOrThrow`

It think it should also be covered by a unit test

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      await getProgramFspConfigurations({
+        programId: copyFromProgramId,
+        accessToken,
+      })
+    ).body;
+    // The source program has data worth duplicating.
+    expect(sourceUsers.length).toBeGreaterThan(0);
+    expect(sourceThresholds.length).toBeGreaterThan(0);
+    expect(sourceProgram.fspConfigurations.length).toBeGreaterThan(0);
+
```
RubenGeo: I think it's better to avoid these halfway assertions if we can. I think in this case it makes sense if we:
1. seed the admin setup. 
2. And than explicitely seed the program, fspConfigurations, programAsssignments, programApprovalThresholds. 

This way we have more control over the input (the program we are about to duplicate) and we can avoid these assertion. 

It would also allow us to validate the resulting program based on the the data that we POSTed instead of what comes back from the getProgram endpoint.

services/121-service/test/program/duplicate-program.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    const duplicatedThresholds = (
+      await getProgramApprovalThresholds({
+        programId: duplicatedProgramId,
+        accessToken,
+      })
+    ).body;
+
+    // Assert
+    expect(duplicateResponse.statusCode).toBe(HttpStatus.CREATED);
+    expectDuplicatedColumnsMatchSource({ source: sourceProgram, duplicatedProgram });
```
RubenGeo: The test assertion helpers in this file contain 'conditional test logic'. Which I think would have given an eslint error (jest/no-conditional-expect) if they were not in helper functions.

I think we can get away with an explicit subset with hardcoded expectations instead of dynamically asserting using propertiesToDuplicate:
- We don't need to assert all ~20 columns since the copy mechanism is generic and unit tested and relies on the config. 
- The purpose of the integrations test mostly the highover flow, e.g. is a new program created and persisted in the database, are the relational entities persisted in the database without validation each column explicitely
- The test is now testing "the code does what the config says". Which is a bit like writing a test for each Permission() decorator we add to each controller
- We don't need conditional logic than if we god for a more hardcoded approach

services/121-service/src/utils/entity-duplication/duplicate-entity.helper.ts
```diff
@@ -0,0 +1,629 @@
...
+    }
+  }
+
+  return childRepository.create(childColumns as DeepPartial<ObjectLiteral>);
+}
+
+// ---------------------------------------------------------------------------
+// Remap phase: re-point foreign keys that reference another duplicated entity.
+// ---------------------------------------------------------------------------
+
```
RubenGeo, line 148: I am having a hard time reviewing this extra code and am sort of thinking that in the end a more specific duplicate program approach would have been better. But I also never saw that prototype so there could be a lot of complexity with that approach that I don't know anything about now 

But I also feel the sunk cost of switching approaches again now. So maybe it's not important enough for me to 100% understand everything going on here. As long as it's covered properly with tests. So let's stick with the current approach for now.

Not sure if we need to discuss this further but I just wanted to share my thought to be transparant and as learning to ourselves for future PRs

RubenGeo (changes requested):

PR #8444, opened 2026-07-02
Add unit and integration tests for upload image from kobo

Piotrk39:
[AB#43082](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43082) 

## Portal preview-deployment

https://happy-rock-0411d2003-8444.westeurope.3.azurestaticapps.net


services/121-service/src/kobo/services/kobo-image.service.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+      programRegistrationAttributesService.getAttributes.mockResolvedValue([
+        { name: mockAttributeName, type: RegistrationAttributeTypes.koboImage },
+      ] as any);
+      registrationsService.getOnePaginatedRegistrationByReferenceId.mockResolvedValue(
+        { [mockAttributeName]: mockImageUrl } as any,
+      );
+      httpService.getStream.mockResolvedValue({
+        headers: { 'content-type': 'image/jpeg' },
+        data: mockStream,
+        status: 200,
```
RubenGeo:
```suggestion
        status: HttpStatus.OK,
```

services/121-service/src/kobo/services/kobo-image.service.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+      await expect(
+        service.getKoboImageStream({
+          programId: mockProgramId,
+          referenceId: mockReferenceId,
+          attributeName: mockAttributeName,
+        }),
+      ).rejects.toThrow(BadRequestException);
+    });
+
+    it('should pass the Kobo token in the Authorization header', async () => {
```
RubenGeo: I don't think we need this one.

services/121-service/test/kobo/kobo-image.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+  });
+
+  it('should stream the image bytes with correct content-type', async () => {
+    // Act
+    const response = await getServer()
+      .get(
+        `/programs/${programId}/registrations/${submissionUuid}/kobo-images/photo`,
+      )
+      .set('Cookie', [accessToken]);
```
RubenGeo: I would create a helper method for this API request, so we can re-use it across these methods

services/121-service/test/kobo/kobo-image.test.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    // Act
+    const response = await getServer()
+      .get(
+        `/programs/${programWithoutKoboId}/registrations/${submissionUuid}/kobo-images/photo`,
+      )
+      .set('Cookie', [accessToken]);
+
+    // Assert
+    expect(response.status).toBe(HttpStatus.NOT_FOUND);
```
RubenGeo: I think we only need the: `should return 404 when the attribute does not exist on the program` as that one is not covered in unit tests I think

I think all of these other scenario you have already covered very well with the unit tests, so we don't need them. 

And for the integrations tests I think it's enough to have 1 happy flow and 1 unhappy flow (to validate that the errors actually end up at the consumer

RubenGeo (changes requested):
Very well done @Piotrk39 . Your unit testing & integration testing improved since MTN.

services/121-service/src/kobo/services/kobo-image.service.spec.ts
RubenGeo: Also in a few other places: (the "should stream an image successfully" test, the "mimetype not allowed" test, and the "Authorization header" test)

services/121-service/test/kobo/kobo-image.test.ts
Piotrk39: Good point! Removed!

services/121-service/src/kobo/services/kobo-image.service.spec.ts
Piotrk39: Yes got it. Changed!

services/121-service/src/kobo/services/kobo-image.service.spec.ts
Piotrk39: Agree, removed!

services/121-service/test/kobo/kobo-image.test.ts
Piotrk39: Added a helper!

RubenGeo (approved):

PR #8570, opened 2026-07-27
chore: delete payments async

Yuri-510:
[AB#43478](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43478) 

## Describe your changes

Deleting a payment with a large amount of relation rows (transactions/ transaction events) could result in the client timing out after 4 minutes of no response.

We now flag the payment as deleted and handle deletion of transactions with a processor.

## Portal preview-deployment


RubenGeo:
Just to couble check, did you test if it was possible by doing this in a synchronize way by optimizing the query?

I think in total this query should delete aprox 500.000 records:
1. Max 100.000 transaction (because that is what we support)
2. Aprox 300.000 transaction events? (created + 2 approval steps per transactions)
3. 100.000 latest transaction events

I think deleting 500.000 records should not take so long, I am guessing this is just because the 'CASCADE' delete that happens now is just inefficient.

I did not test this so it could be that I am wrong



Update: I tried the code in this [branch](https://github.com/global-121/121-platform/tree/rubengeo/delete-payment-fast). It deletes a payment of 65.000 registrations in ~3 seconds. The code on that branch is a mess though, let's discuss when you're back

PR #8574, opened 2026-07-27
E2E: Duplicate program

nielsVoogt:
[AB#43511](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43511) 

## Describe your changes

Added E2E test for new duplicate program flow

- [x] I have added tests


e2e/portal/pages/ProgramOverviewPage.ts
Yuri-510, line 1: This is what the "landing page" looks like on initial production state, so here I would expect a `openCreateNewProgram`. However, I don't expect/see any duplication functionality here. Maybe we should create a "ProgramsOverviewPage"? What do you think?

Image

e2e/portal/tests/CreateProgram/program-data.ts
```diff
@@ -0,0 +1,30 @@
+import { CurrencyCode } from '@121-service/src/exchange-rates/enums/currency-code.enum';
+import { Fsps } from '@121-service/src/fsp-integrations/shared/enum/fsp-name.enum';
+
+import { getFspLabels } from '@121-e2e/portal/helpers/get-fsp-labels';
+
+export const getProgramData = ({
```
Yuri-510: Consider a more apt naming like `getCreateProgramFormValues`

e2e/portal/tests/CreateProgram/program-data.ts
Yuri-510, line 1: I see you created a couple of files in the e2e folder using kebab-case, while the standard in this folder looks to be PascalCase. I think it's better to keep it uniform. It also separates clearly what is an e2e file. What do you think? You can specify what kind of file it is by using `*.helper.ts` or `*.const.ts`.

e2e/portal/tests/CreateProgram/DuplicateProgramSuccessfully.spec.ts
Yuri-510, line 1: Consider moving the 'Duplicate the program' test to the create program test file, so we only have to create the program once for both the creation and duplication flow and don't need the shared helper.

e2e/portal/pages/ProgramOverviewPage.ts
nielsVoogt: @Yuri-510 Yes, I agree and actually started renaming this myself as well before opening the PR. However, it is used in a lot of places where this name change would make less sense. And keep in mind, this the root view after login so I kind of get understand it was named `HomePage`.

Come to think of it, in other places where this is used it's mostly for stuff like this:
```
    await homePage.selectAccountOption('Change password');
    await homePage.selectAccountOption('Logout');
```

I can maybe change these from `homePage` to just use `basePage` instead, since these actions are not specificly tied to the "homepage". When that is done, the `programOverviewPage` change would make sense. 

What are your thoughts on this?

e2e/portal/tests/CreateProgram/DuplicateProgramSuccessfully.spec.ts
nielsVoogt: @Yuri-510 
In a following test - for excel edge cases - I also use the createProgram to get a clean program from initialState. See changes [here](https://github.com/global-121/121-platform/pull/8575/changes#diff-6eb0d565f16cea457cdae977bc0f4a92e42921756282daf03543b1bca2f39bfcR25).  I can also use a seed there, but then I'd have to edit the test to first remove existing FSP's.

What do you prefer?

e2e/portal/pages/ProgramOverviewPage.ts
Yuri-510: I would create a new page file `ProgramsOverviewPage` next to `HomePage`.

e2e/portal/tests/CreateProgram/DuplicateProgramSuccessfully.spec.ts
Yuri-510: I think your suggestion is a good idea. That seems to be the pattern we're currently using.

e2e/portal/pages/ProgramOverviewPage.ts
nielsVoogt: @Yuri-510 
I feel that leaving in `HomePage` is incorrect. Since the actions (`selectAccountOption`, etc) are already inside `BasePage` I see no value in adding a empty `HomePage` file that only extends `BasePage`.

Piotrk39: Feels like ```HomePage``` is at this point more of a remnant of what we believed at this time this landing page is also it feels like this is what ```BasePage``` actually does. I agree with moving this now to ```BasePage``` as we keep logic in one place and that we kind of have a home page but in reality it is just a landing page for selecting a program as Portal grows we need to avoid too much complexity in E2E tests.

e2e/portal/tests/CreateProgram/program-data.ts
nielsVoogt: Not relevant anymore - see changes

Yuri-510 (approved):

PR #8602, opened 2026-07-30
Start payment faster by adding jobs in bulk

RubenGeo:
[AB#43726](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43726) 

## Describe your changes
Reason: 

Elwin found out that start payment took around 4 minutes on NLRC production. This is a small attempt to make it better.

 
Fix: 
- Locally I found that: startPayment was taking 10+ seconds for large payments (~32k registrations) 
- Jobs are added to the queue one by one which results in a lot of awaits. 
- I changed it to add jobs in bulk

Contemplation:
- Not sure if this will fix  the issue completely on NLRC as there the 4 minutes mark is much larger than my local 10 second mark, not sure what the cause for that is. 
- My current fix is a low effort attempt to make it a bit better
- I suggest we'll put in more effort If our users ever start complaining about this

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/fsp-integrations/transaction-queues/transaction-queues.service.ts
+++ b/services/121-service/src/fsp-integrations/transaction-queues/transaction-queues.service.ts
@@ -1,4 +1,5 @@
 import { Inject, Injectable } from '@nestjs/common';
+import { Queue } from 'bull';
 import Redis from 'ioredis';
 
 import { AirtelTransactionJobDto } from '@121-service/src/fsp-integrations/transaction-queues/dto/airtel-transaction-job.dto';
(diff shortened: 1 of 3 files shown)


services/121-service/src/fsp-integrations/transaction-queues/transaction-queues.service.ts
```diff
@@ -29,124 +31,121 @@ export class TransactionQueuesService {
...
+      transactionJobs.map((transactionJob) => ({
+        name: JobNames.default,
+        data: transactionJob,
+      })),
+    );
+
+    const jobIds = jobs
+      .map((job) => job.id)
+      .filter((id) => id !== undefined && id !== null)
+      .map((id) => String(id));
```
Yuri-510: It's Monday morning and I feel like nitpicking :p 

The first map looks to be redundant.

Consider a little more readable version:

```javascript
const jobIds = jobs
      .filter((job) => job.id != null)
      .map((job) => String(job.id));
```

nameerhqureshi (approved):

PR #8626, opened 2026-08-04
Add twilio disabled mode

RubenGeo:
[AB#43766](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43766) 

## Describe your changes

Add a `TWILIO_MODE` env variable with three values: `EXTERNAL`, `MOCK`, `DISABLED`. It replaces the old `TWILIO_MOCK` boolean.

`DISABLED` mode lets an instance run without Twilio. No messages are sent. WhatsApp, SMS, lookup, and incoming-message handling all check this mode and skip real Twilio calls.

Two new startup checks were added:
- `TwilioEnvVariableValidationService` checks Twilio env vars are set, unless `TWILIO_MODE` is `DISABLED`.
- `FspEnvVariableValidationService` now also checks that FSPs which need Twilio (e.g. Intersolve Voucher) aren't enabled while `TWILIO_MODE` is `DISABLED`. This prevents silent failures at payment time.

## Why this prework is needed

[AB#42840](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42840) wants every program to require a `phoneNumber` attribute, but only when Twilio is enabled for the instance.

There is currently no clear, instance-level setting for "is Twilio on or off". The old `TWILIO_MOCK` boolean only chose between real and mock Twilio. It could not express "Twilio is off".

This PR adds that missing setting. [AB#42840](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/42840) can then use `TWILIO_MODE !== DISABLED` as the single source of truth for whether a program must have a phoneNumber attribute.

Please do not merge until `TWILIO_MODE` is set on each server.

## Why did I chose to silently skip sending messages in Disabled mode

The main reason is because it was easiest to implement, and it's an improvement on the curent behavior. Because instances that do no use Twilio now are on mock mode, which is even worse. Because in mock mode the server pretends to send message while it does not actually send them. 

I did add task [Throw errors/ hide message buttons from portal if Twilio is disabled?](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/43770), as I think it's the desired behaviour. However I actually think this may be out of scope for this item.

- [x] ~~I do not need any deviation from [our PR guidelines](https://github.com/global-121/121-platform/blob/main/docs/CONTRIBUTING.md#pull-request-guidelines)~~  ~~Please do not merge until we set the TWILIO_MODE on each server~~ Twilio mode has been set on all instances

## Portal preview-deployment


services/121-service/src/migration/1742574202000-message-template-content-sid.ts
```diff
@@ -71,7 +72,7 @@ export class MessageTemplateContentSid1742574202000 implements MigrationInterfac
     // Migrate instance that are in mock mode: important for demo and training
     // This will set the contentSid to a mock value related to the language and type
     // It does not take into account the mock message we have for PV and OCW, it seemed not worth the effort
-    if (env.MOCK_TWILIO) {
+    if (env.TWILIO_MODE === TwilioMode.mock) {
```
Yuri-510: This is really prone to errors, since it needs to be actively maintained. Is there a better solution?

services/121-service/src/notifications/lookup/lookup.service.ts
```diff
@@ -16,6 +18,10 @@ export class LookupService {
     phoneNumber: string,
     throwNoException?: boolean,
   ): Promise<string | undefined> {
+    // When Twilio is disabled, skip the carrier lookup and pass the number through.
```
Yuri-510: I think the code reads exactly the same as this comment, making the comment adding noise rather than adding value.

services/121-service/src/fsp-integrations/shared/dto/fsp-env-variables.dto.ts
```diff
@@ -5,4 +5,8 @@ export class FspEnvVariablesDto {
   readonly mode: FspMode;
 
   readonly variables: EnvVariablesReadOnly;
+
+  // When true, this FSP deliveres message via WhatsApp/SMS via Twilio
+  // and therefore cannot function when Twilio is disabled (TWILIO_MODE=DISABLED).
```
Yuri-510, line 10: Can it not be derived from `TWILIO_MODE` then?

services/121-service/src/notifications/message-queues/message-queues.service.spec.ts
```diff
@@ -82,6 +97,27 @@ describe('MessageQueuesService', () => {
     );
   });
 
+  it('should not add message to queue when TWILIO_MODE is disabled', async () => {
```
Yuri-510, line 101: This reads as if the `message-queues` service is tightly coupled with Twilio, which would suggest this service actually is a Twilio messages queue service. Or am I missing something?

services/121-service/src/payments/services/fsp-env-variable-validation.service.ts
```diff
@@ -6,78 +6,89 @@ import {
...
-      case FspMode.disabled:
-      case FspMode.mock:
-        return;
-      case FspMode.external:
-        break;
-      default: {
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars -- exhaustive check necessary
-        const exhaustiveCheck: never = mode;
-      }
+    // A disabled FSP has nothing to validate.
```
Yuri-510: Nitpick: I think also for this block, the code the describes exactly what is does, so in my opinion the comment adds noise instead of value.

services/121-service/src/payments/services/fsp-env-variable-validation.service.ts
```diff
@@ -6,78 +6,89 @@ import {
...
-    const envVariableIsSet = (envVar: unknown): boolean =>
-      envVar !== undefined && envVar !== null && envVar !== '';
+    const messages: string[] = [];
 
-    const missingVariables = Object.entries(variables).filter(
-      ([_key, value]) => !envVariableIsSet(value),
-    );
+    // FSPs that deliver to registrations over WhatsApp/SMS via Twilio (e.g.
+    // Intersolve Voucher) cannot function when Twilio is disabled. This applies
+    // whenever the FSP is enabled (MOCK or EXTERNAL).
```
Yuri-510: The code is perfectly understandable in itself. I'm not saying this to blindly force a certain dogma. It's just that to read what is happening we now have to read two blocks of text/code, so if every block of code had a comment describing what it does (and sometimes this _is_ needed) it would take twice the amount of time reading through code to understand what is happening. It just adds up.

services/121-service/src/notifications/twilio.client.ts
```diff
@@ -24,14 +25,17 @@ class MockTwilioRequestClient extends RequestClient {
...
 
-if (env.MOCK_TWILIO) {
+if (env.TWILIO_MODE === TwilioMode.mock) {
   mockClient = {
     httpClient: new MockTwilioRequestClient(`${env.MOCK_SERVICE_URL}/api`),
   };
 }
 
+// TWILIO_SID/TWILIO_AUTHTOKEN are optional because they are not required when
+// TWILIO_MODE=DISABLED. Fall back to placeholder values so the client can still
```
Yuri-510, line 35: Is there a way to not construct the Twilio client when twilio mode is disabled?

services/121-service/src/notifications/message.module.ts
```diff
@@ -46,8 +48,36 @@ import { AzureLogService } from '@121-service/src/shared/services/azure-log.serv
...
+    const validationResult =
+      this.twilioEnvVariableValidationService.validateTwilioEnvVariables({
+        mode: env.TWILIO_MODE,
+        variables: {
+          TWILIO_SID: env.TWILIO_SID,
+          TWILIO_AUTHTOKEN: env.TWILIO_AUTHTOKEN,
+          TWILIO_WHATSAPP_NUMBER: env.TWILIO_WHATSAPP_NUMBER,
+          TWILIO_MESSAGING_SID: env.TWILIO_MESSAGING_SID,
+        },
+      });
```
Yuri-510, line 71: It looks like this validation is also done when twilio mode is disabled. This, together with the fact that Twilio client is also constructed when mode is disabled got me thinking it might me a better idea to create a twilio service, which would look something like this:

```js
@Injectable()
export class TwilioService implements OnModuleInit {
  private client: ReturnType<typeof twilio> | undefined;

  constructor(
    private readonly validationService: TwilioEnvVariableValidationService,
  ) {}

  onModuleInit() {
    // fail-fast validation, colocated with the thing it validates
    const result = this.validationService.validateTwilioEnvVariables({ ... });
    // log / throw ...

    // construct only when enabled — never in disabled mode
    if (env.TWILIO_MODE !== TwilioMode.disabled) {
      this.client = twilio(env.TWILIO_SID, env.TWILIO_AUTHTOKEN, this.buildOpts());
    }
  }

  private getClient() {
    if (!this.client) {
      throw new Error('Twilio client requested while TWILIO_MODE=DISABLED.');
    }
    return this.client;
  }

  // wrapped operations
  createMessage(payload) { return this.getClient().messages.create(payload); }
  lookupPhoneNumber(number) { return this.getClient().lookups.v1.phoneNumbers(number).fetch({ type: ['carrier'] }); }
  // ...
}
```

Yuri-510: Then this could be injected in the other services (messages etc.).

services/121-service/src/notifications/services/twilio-env-variable-validation.service.ts
```diff
@@ -0,0 +1,42 @@
...
+export class TwilioEnvVariableValidationService {
+  // A pure function.
+  public validateTwilioEnvVariables({
+    mode,
+    variables,
+  }: {
+    mode: TwilioMode;
+    variables: Record<string, string | undefined>;
+  }): { ok: boolean; messages: string[] } {
+    // When Twilio is disabled, no environment variables are required.
```
Yuri-510: See my other comments on these type of comments. I think these are often generated to generate more output tokens and therefore more revenue. In the long run I think they are not going to make our lives easier.

Yuri-510 (commented):
Nice work overall. I have some design questions and I see a lot of comments that seem to describe the same as the code does, so you'll see some nitpicking here and there. I'm starting to think copilot generates these because it means more output tokens to be consumed, thus more revenue. However, I strongly believe they take up more of our time in understanding what the code is saying, so it might be a good idea to update our agent.md's on this, but lets discuss.

services/121-service/src/migration/1742574202000-message-template-content-sid.ts
RubenGeo: It's an old migration file that had a type error since I changed the env variables. 

I'll just comment that section out instead since this migration is already in the past

services/121-service/src/notifications/message-queues/message-queues.service.spec.ts
RubenGeo: Yes, we only send text messages via twilio. We could rename all whatsapp, message, sms, related services to something with twilio, but I would say that's out of scope

services/121-service/src/fsp-integrations/shared/dto/fsp-env-variables.dto.ts
RubenGeo: This is a property in an interface to mark whether FSP is dependencd on Twilio

It's used to mark whether validation on startup needs to happen if twilio is enabled when an FSP is enabled

services/121-service/src/payments/services/fsp-env-variable-validation.service.ts
RubenGeo: No worries, agreed that I left too many comments from our friend claude in this PR

services/121-service/src/notifications/twilio.client.ts
RubenGeo: I'll reply to https://github.com/global-121/121-platform/pull/8626#discussion_r3727559610, I did not find another good way to do this.

services/121-service/src/notifications/message.module.ts
RubenGeo: > This, together with the fact that Twilio client is also constructed when mode is disabled got me thinking it might me a better idea to create a twilio service, which would look something like this.

I think it's a cool idea and was already inclined to accept it at first. 

However when I started working on it I sort of backtracked again.

Writing down my thoughts on this:

What problems are we trying to solve with your solution?

**1**

> It looks like this validation is also done when twilio mode is disabled.

Not sure if I understand what you mean here. I re-used the pattern as we have for `validateFspEnvVariableSettings` where we are validating the Fsp env variables based on the FspMode. Because validateTwilioEnvVariables takes in the TwilioMode I think it's clear enough that validation depends on the Twilio mode.

**2**
> Twilio client is also constructed when mode is disabled

I don't think this will have any noticable performance implications for our code. So the main reason to introduce a Twilio clients wrapper is readability of the code.

This wrapper would resolve the weirdness around the comment in file `twilio.client.ts`
image

Which I don't think is a big enough problem to warrant this extra service. Because it would come with the burden of:
- if we want to use new methods from the sdk that we are currently not using, the person developing needs to understand that this wrapper and having to update it 
- And it adds a small maintenance burden of having to maintain the wrapper which needs to match the twilio sdk client. So if they ever change the methods of the sdk we need to update the wrapper as well

RubenGeo:
Thanks for the review @Yuri-510 , good attention to details (Y)

services/121-service/src/fsp-integrations/shared/dto/fsp-env-variables.dto.ts
Yuri-510: Then I would just leave out this comment. The comment is likely to just raise questions while the code itself is straightforward.

services/121-service/src/notifications/lookup/lookup.service.ts
```diff
@@ -16,6 +18,9 @@ export class LookupService {
     phoneNumber: string,
     throwNoException?: boolean,
   ): Promise<string | undefined> {
+    if (env.TWILIO_MODE === TwilioMode.disabled) {
+      return this.sanitizePhoneNrExtra(phoneNumber).replace(/\D/g, '');
```
Yuri-510, line 22: A question: we do need the plus-sign (+) for phone numbers, right? Your regex would also remove that.

RubenGeo: It's horrible, but now we always store all our phonenumbers without the + and than append them while sending it with twilio

RubenGeo: It's maybe not so horrible as the requirement per FSP also differs, so maybe it is okay to always start with the most trimmed down version in the DB. Without 00 and +

Yuri-510 (approved):

PR #8638, opened 2026-08-06
Remove allow empty phone number setting

RubenGeo:
[AB#43768](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43768) 

## Describe your changes

- Program-level "allow empty phone number" toggle is removed entirely. Programs can no longer be configured to allow empty phone numbers via this setting.

Whether a phone number is required is now driven by two things instead of the program flag:

- The phoneNumber program registration attribute's own isRequired property (per-attribute), and
- The global TWILIO_MODE environment variable (instance-wide).
- Kobo form validation now keys off Twilio mode. A phoneNumber survey field is required when TWILIO_MODE !== DISABLED (previously it was required unless allowEmptyPhoneNumber was true).

Program creation auto-adds non-required phoneNumber attribute only when Twilio is enabled. 

Registration import/update phone-number validation is now generic. The dedicated validatePhoneNumberEmpty check was removed; the required-empty error is now emitted by the generic required-tel-attribute path. 

cleanCustomDataIfPhoneNr is removed. As this enforcment already happens in the input validator, this part was sort of double validation

## Portal preview-deployment

https://happy-rock-0411d2003-8638.westeurope.3.azurestaticapps.net

### DIFF
--- a/services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
+++ b/services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
@@ -457,7 +457,7 @@ describe('ProgramRegistrationAttributesService', () => {
       mockEnv.TWILIO_MODE = TwilioMode.mock;
     });
 
-    it('adds a required phoneNumber attribute when it is missing and Twilio is enabled', async () => {
+    it('adds a non-required phoneNumber attribute when it is missing and Twilio is enabled', async () => {
       // Act
       const result =
         await programRegistrationAttributesService.applyProgramRegistrationAttributesFallbackIfNecessary(
@@ -474,6 +474,7 @@ describe('ProgramRegistrationAttributesService', () => {
       );
       expect(phoneNumberAttribute).toBeDefined();
       expect(phoneNumberAttribute?.isRequired).toBe(false);
+      expect(phoneNumberAttribute?.type).toBe(RegistrationAttributeTypes.tel);
     });
 
     it('does not duplicate the phoneNumber attribute when it is already present', async () => {
--- a/services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
+++ b/services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
@@ -13,7 +13,10 @@ import {
 import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
 import { ProgramRegistrationAttributeEntity } from '@121-service/src/programs/entities/program-registration-attribute.entity';
 import { RegistrationViewEntity } from '@121-service/src/registration/entities/registration-view.entity';
-import { RegistrationAttributeTypes } from '@121-service/src/registration/enum/registration-attribute.enum';
+import {
+  DefaultRegistrationDataAttributeNames,
+  RegistrationAttributeTypes,
+} from '@121-service/src/registration/enum/registration-attribute.enum';
 import { generateMockCreateQueryBuilder } from '@121-service/src/utils/test-helpers/createQueryBuilderMock.helper';
 
 jest.mock('@121-service/src/env', () => ({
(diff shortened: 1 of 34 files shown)


services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
```diff
@@ -448,4 +451,74 @@ describe('ProgramRegistrationAttributesService', () => {
...
+        );
+
+      // Assert
+      const phoneNumberAttribute = result.find(
+        (attr) =>
+          attr.name === DefaultRegistrationDataAttributeNames.phoneNumber,
+      );
+      expect(phoneNumberAttribute).toBeDefined();
+      expect(phoneNumberAttribute?.isRequired).toBe(false);
+    });
```
Yuri-510: So the attribute is required, but the value is not? It's a bit confusing for me that the test tests that it "adds a required phoneNumber", but then asserts that it's not required.

services/121-service/src/program-registration-attributes/program-registration-attributes.service.ts
```diff
@@ -226,15 +226,19 @@ export class ProgramRegistrationAttributesService {
...
 
-    // make sure phoneNumber is in programRegistrationAttributes
-
+    // Twilio needs a required phoneNumber to send messages, so only add it when Twilio is enabled.
     if (
-      !programRegistrationAttributes.find((attr) => attr.name === 'phoneNumber')
+      env.TWILIO_MODE !== TwilioMode.disabled &&
+      !programRegistrationAttributes.find(
+        (attr) =>
+          attr.name === DefaultRegistrationDataAttributeNames.phoneNumber,
```
Yuri-510: I know this is not your code, but lets make it a little nicer by introducing a const that reads something like `phoneNumberAttributeMissing` so the `if` statement would be:

```js
const phoneNumberAttributeMissing = !programRegistrationAttributes.find((attr) => attr.name === 'phoneNumber');

if (
  env.TWILIO_MODE !== TwilioMode.disabled &&
  phoneNumberAttributeMissing
)
```

services/121-service/src/registration/services/registrations-creation.service.spec.ts
```diff
@@ -24,9 +24,7 @@ describe('RegistrationsCreationService', () => {
     const programService = unitRef.get(ProgramService);
     jest
       .spyOn(programService as any, 'findProgramOrThrow')
-      .mockImplementation(() => ({
-        allowEmptyPhoneNumber: false,
-      }));
+      .mockImplementation(() => ({}));
```
Yuri-510: You can just remove this block

services/121-service/src/program-registration-attributes/program-registration-attributes.service.spec.ts
RubenGeo: Good question

So when Twilio is enabled on an instance we automatically add a phonenumber attribute.

The main goals of adding that question is to ensure that a user realizes phonenumber is used to send messages when using twilio. So it also appears in the registrations import csv template etc. 

However it does not mean that all people in a program need to have a phonenumber. We have seen that both in the DRC and Boodschappenkaarten programma.

I discussed with James that it makes more sense to have the default on non required, else you would get errors if kobo submission come in from a kobo form that has a non required phonenumber question.

RubenGeo: This comment did inspire me to add this commit. Please take a look at it. [`56339a2` (this PR)](https://github.com/global-121/121-platform/pull/8638/changes/56339a2a863673af3dee5e06c79a3ca1c7ae6aa7)

Yuri-510 (approved):

PR #8658, opened 2026-08-13
fix: Twilio Error 63021 AH payment

nameerhqureshi:
[AB#43910](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43910) 

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
+++ b/services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
@@ -17,7 +17,6 @@ import {
   TwilioStatus,
   TwilioStatusCallbackDto,
 } from '@121-service/src/notifications/dto/twilio.dto';
-import { TwilioMessageEntity } from '@121-service/src/notifications/entities/twilio.entity';
 import { MessageContentType } from '@121-service/src/notifications/enum/message-type.enum';
 import { ProgramNotificationEnum } from '@121-service/src/notifications/enum/program-notification.enum';
 import { MessageQueuesService } from '@121-service/src/notifications/message-queues/message-queues.service';
@@ -49,8 +48,6 @@ export class IntersolveVoucherService {
   public readonly transactionRepository: Repository<TransactionEntity>;
   @InjectRepository(ProgramEntity)
   public readonly programRepository: Repository<ProgramEntity>;
-  @InjectRepository(TwilioMessageEntity)
-  private readonly twilioMessageRepository: Repository<TwilioMessageEntity>;
 
   private readonly fallbackRegistrationPreferredLanguage =
     RegistrationPreferredLanguage.en;
@@ -122,7 +119,7 @@ export class IntersolveVoucherService {
         );
         return;
       } else {
-        // .. if existing voucher is found, then continue with that one, and don't create new one
+        // .. if existing voucher found, then continue with that one, and don't create new one
         paResult.status = TransactionStatusEnum.success;
       }
     } else {
@@ -458,38 +455,31 @@ export class IntersolveVoucherService {
   public async updateTransactionProgressBasedOnInitialMessage({
     transactionId,
     newTransactionStatus,
-    messageSid,
     errorMessage,
     userId,
     programFspConfigurationId,
   }: {
     transactionId: number;
-    newTransactionStatus: TransactionStatusEnum;
-    messageSid?: string;
+    newTransactionStatus:
+      | TransactionStatusEnum.waiting
+      | TransactionStatusEnum.error;
     errorMessage?: string;
     userId: number;
     programFspConfigurationId: number;
   }): Promise<void> {
-    await this.transactionsService.saveProgress({
-      context: {
-        transactionId,
-        userId,
-        programFspConfigurationId,
-      },
+    const context = {
+      transactionId,
+      userId,
+      programFspConfigurationId,
+    };
+
+    await this.transactionsService.saveProgressUnlessTransactionFailed({
+      context,
       description:
         TransactionEventDescription.intersolveVoucherInitialMessageSent,
       newTransactionStatus,
       errorMessage,
     });
-
-    if (messageSid) {
-      await this.twilioMessageRepository.update(
-        { sid: messageSid },
-        {
-          transactionId,
-        },
-      );
-    }
   }
 
   public async getVoucherImage(
@@ -680,13 +670,11 @@ export class IntersolveVoucherService {
     transactionId,
     newTransactionStatus,
     errorMessage,
-    messageSid,
     intersolveVoucherId,
   }: {
     transactionId: number;
     newTransactionStatus: TransactionStatusEnum;
     errorMessage: string | null;
-    messageSid?: string;
     intersolveVoucherId: number;
   }): Promise<void> {
     await this.transactionsService.saveProgressFromExternalSource({
@@ -703,15 +691,6 @@ export class IntersolveVoucherService {
       });
     intersolveVoucher.send = true;
     await this.intersolveVoucherScopedRepository.save(intersolveVoucher);
-
-    if (messageSid) {
-      await this.twilioMessageRepository.update(
-        { sid: messageSid },
-        {
-          transactionId,
-        },
-      );
-    }
   }
 
   public async getVouchersWithBalance(
--- a/services/121-service/src/notifications/whatsapp/whatsapp.service.ts
+++ b/services/121-service/src/notifications/whatsapp/whatsapp.service.ts
@@ -52,6 +52,7 @@ export class WhatsappService {
     messageProcessType,
     existingSidToUpdate,
     userId,
+    transactionId,
     firstAttempt = true,
   }: {
     message?: string;
@@ -63,6 +64,7 @@ export class WhatsappService {
     messageProcessType?: MessageProcessType;
     existingSidToUpdate?: string;
     userId: MessageSenderUserId;
+    transactionId?: number;
     firstAttempt?: boolean; // Controls retry logic for Twilio media errors (63021)
   }): Promise<string | undefined> {
     const payload = {
@@ -110,6 +112,7 @@ export class WhatsappService {
         message: messageToStore,
         userId,
         registrationId,
+        transactionId,
         mediaUrl,
         messageContentType,
         messageProcessType,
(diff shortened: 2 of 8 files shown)


services/121-service/src/payments/transactions/transactions.service.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  beforeEach(() => {
+    const { unit, unitRef } = TestBed.create(TransactionsService).compile();
+
+    transactionsService = unit;
+    transactionRepository = unitRef.get(TransactionRepository);
+    transactionEventsService = unitRef.get(TransactionEventsService);
+  });
+
+  describe('saveWaitingProgressUnlessTransactionFailed', () => {
```
RubenGeo: It's a bit better to describe behavior or scenario, not which function is called in the describe block.  https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#test-description-naming

services/121-service/src/payments/transactions/transaction.repository.ts
```diff
@@ -47,6 +47,25 @@ export class TransactionRepository extends Repository<TransactionEntity> {
       .execute();
   }
 
+  public async updateStatusUnlessIn({
```
RubenGeo, line 50: Smart way to prevent race conditions 👍

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
```diff
@@ -118,11 +121,12 @@ export class WhatsappService {
     } catch (error) {
       if (
         isSameAsString(error.code, TwilioErrorCodes.mediaUrlInvalid) &&
-        mediaUrl &&
+        (mediaUrl || contentSid) &&
```
RubenGeo: I tried to find the failed message for # 29077, but was not able to find it.

So I was wondering did you find an example of this happening where a twilio 63021 error occured on a message where no media url was send?

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
```diff
@@ -118,11 +121,12 @@ export class WhatsappService {
...
         firstAttempt
       ) {
         firstAttempt = false;
```
RubenGeo, line 127: Maybe a pattern like this can be a bit more readable: 

```suggestion
     const messageContainsMedia = Boolean(mediaUrl) || Boolean(contentSid);
      const isRetryableMediaError =
        isSameAsString(error.code, TwilioErrorCodes.mediaUrlInvalid) &&
        messageContainsMedia;
        firstAttempt = false;
```

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
RubenGeo: The previous registration (# 28705) for which a message failed  due to 63021 did have a media url

services/121-service/src/payments/transactions/transactions.service.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    it('should not create an event when a failed transaction cannot be set to waiting', async () => {
```
RubenGeo, line 29: The added tests are missing - // Arrange / // Act / // Assert

services/121-service/src/payments/transactions/transactions.service.ts
```diff
@@ -120,6 +120,34 @@ export class TransactionsService {
...
+  }): Promise<void> {
+    const wasWaitingStatusSet =
+      await this.transactionRepository.updateStatusUnlessIn({
+        transactionId: context.transactionId,
+        newTransactionStatus: TransactionStatusEnum.waiting,
+        excludedStatuses: [
+          TransactionStatusEnum.error,
+        ],
+      });
+    if (!wasWaitingStatusSet) {
```
RubenGeo: This early return means that we skip setting the initial transaction event:  'initial WhatsApp message was send'. So this is silently dropped when a failure callback landed first.

I was thinking if that makes sense but I am unsure. What do you think is more confusing for a user if the order of the events does not make sense, or if one is silently ignored?

RubenGeo (changes requested):
Good find on the race condition, hope that solve it.

The rest of the review contains some nitpicking and open questions

nielsbom:
I was experimenting with https://github.com/rcieri/glab-tui and accidentally marked this pull request as draft. I marked it for review again to fix that. Sorry!

services/121-service/src/payments/transactions/transactions.service.ts
nameerhqureshi: Good question. When the failure callback lands first, the transaction is already in `error` and an error event exists. Writing the 'initial WhatsApp message was sent' event after that would be factually wrong (the message failed to send) and would produce a confusing timeline: `error → waiting`. So we intentionally skip it, the final state the user sees is the error, which is the accurate one. The event isn't silently dropped in the sense that the outcome is still fully visible via the error event.

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
nameerhqureshi: I actually couldn't verify #29077's exact message either (same as you). What I can say from the code: registration messages are sent via Twilio Content templates (`contentSid`), whose header media is configured on Twilio's side, so a 63021 can occur even when we never pass our own `mediaUrl`. The old retry condition only covered `mediaUrl`, which meant any template message with header media would never be retried on 63021. The `contentSid` addition closes that gap regardless of #29077's specifics. If you know how to retrieve #29077's actual Twilio message, happy to double-check whether it really was a 63021 on a template send.

services/121-service/src/payments/transactions/transactions.service.ts
RubenGeo: you are right

services/121-service/src/notifications/whatsapp/whatsapp.service.ts
nameerhqureshi: Yes you are right. The message for `#29077` did had a mediaUrl when I dig, so I have reverted back this code to just to `mediaUrl` only. I have created a separate [task](https://dev.azure.com/redcrossnl/121%20Platform/_workitems/edit/44053) in which will try to fix the issue that the message had an `errorCode` 63021 but its status was updated as `sent`.

services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
```diff
@@ -458,38 +455,41 @@ export class IntersolveVoucherService {
...
     userId: number;
     programFspConfigurationId: number;
   }): Promise<void> {
+    const context = {
+      transactionId,
+      userId,
+      programFspConfigurationId,
+    };
+
+    if (newTransactionStatus === TransactionStatusEnum.waiting) {
```
RubenGeo: It took me some time to understand what was happening here as the logic with messages, transactions and race conditions is really complex. 

But I think it can potentially be simplified here, as I think in this function newTransactionStatus can only be waiting or error. And error does not need a completely seperate flow I think. So could we refactor it like this:



```ts
  public async updateTransactionProgressBasedOnInitialMessage({
    transactionId,
    newTransactionStatus,
    errorMessage,
    userId,
    programFspConfigurationId,
  }: {
    transactionId: number;
    newTransactionStatus:
      TransactionStatusEnum.waiting | TransactionStatusEnum.error;
    errorMessage?: string;
    userId: number;
    programFspConfigurationId: number;
  }): Promise<void> {
    const context = {
      transactionId,
      userId,
      programFspConfigurationId,
    };

    await this.transactionsService.saveProgressUnlessTransactionFailed({
      context,
      newTransactionStatus,
      description:
        TransactionEventDescription.intersolveVoucherInitialMessageSent,
      errorMessage,
    });
  }
```

```ts
// transactions service
/// rename saveWaitingProgressUnlessTransactionFailed to saveProgressUnlessTransactionFailed
  public async saveProgressUnlessTransactionFailed({
    context,
    description,
    newTransactionStatus,
    errorMessage,
  }: {
    context: TransactionEventCreationContext;
    description: TransactionEventDescription;
    newTransactionStatus:
      TransactionStatusEnum.waiting | TransactionStatusEnum.error;
    errorMessage?: string;
  }): Promise<void> {
    const wasStatusSet =
      await this.transactionRepository.updateStatusUnlessIn({
        transactionId: context.transactionId,
        newTransactionStatus,
        excludedStatuses: [TransactionStatusEnum.error],
      });
    if (!wasStatusSet) {
      return;
    }

    await this.transactionEventsService.createEvent({
      context,
      description,
      errorMessage,
    });
  }
```

- I removed the extra branch in updateTransactionProgressBasedOnInitialMessage for non-waiting. Because I think wasStatusSet will always be true for an newTransactionStatus of error. As there are no status callback from twilio that can cause the race condtion if the initial message request to twilio gives an error
- I added more explicit typing for newTransactionStatus

What do you think?

RubenGeo (changes requested):

services/121-service/src/fsp-integrations/integrations/intersolve-voucher/services/intersolve-voucher.service.ts
nameerhqureshi: Thank you for the suggestion, I have implemented. The branch in `updateTransactionProgressBasedOnInitialMessage` is removed and both `waiting` and `error` now go through a single `saveProgressUnlessTransactionFailed` helper, with the explicit `waiting | error` typing you proposed.

One small note on your assumption -> wasStatusSet (`wasTransactionStatusSet`) is not always true for `error`, if a Twilio failure callback has already landed and set the transaction to `error`, the atomic `updateStatusUnlessIn` guard will skip the second write (and the duplicate event). That's intentional because it protects against out-of-order callbacks in both directions. Also adjusted `errorMessage` to optional (`errorMessage?: string`) since the waiting case passes `undefined`.

RubenGeo: I think the error only occurs when messageSid is undefined, i.e. when twilioClient.messages.create(...) threw in whatsapp.service.ts:88. 

I think this means that:

- Twilio never accepted the message, so it will never emit a status callback for it.
- Even if it somehow did, the callback lookup matches on sid = callbackData.MessageSid

But that guard still does not hurt there so it think the current state of the code is good

RubenGeo (approved):

PR #8668, opened 2026-08-17
feature: AlFouad job processors

Yuri-510:
[AB#43836](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43836) 

## Describe your changes

AlFouad job processors

## Portal preview-deployment

https://happy-rock-0411d2003-8668.westeurope.3.azurestaticapps.net


interfaces/portal/src/assets/fsps/alfouad.jpg
RubenGeo: Still worth it to ask for an official logo

services/121-service/src/fsp-integrations/reconciliation/alfouad/alfouad-reconciliation.service.ts
```diff
@@ -0,0 +1,110 @@
...
+  public async doAlfouadReconciliation(): Promise<number> {
+    const transactionIds =
+      await this.transactionRepository.getWaitingTransactionIdsByFsp({
+        fspName: Fsps.alfouad,
+      });
+
+    for (const transactionId of transactionIds) {
+      try {
+        await this.reconcileTransaction(transactionId);
+      } catch (error) {
```
RubenGeo, line 31: This catch-all swallows every error the same way, whether it's an expected/transient failure (e.g. a timeout from Al Fouad's API) or an unexpected bug (e.g. a TypeError introduced by a future refactor). I'd suggest distinguishing the two: let unexpected errors propagate/fail loudly (so we notice regressions) and only swallow-and-continue for known transient/external failures. Maybe distinquish using Al Fouad errors?

services/121-service/src/fsp-integrations/reconciliation/alfouad/alfouad-reconciliation.service.ts
```diff
@@ -0,0 +1,110 @@
...
+      requestIdentity,
+    });
+
+    if (!transactionState) {
+      throw new Error(
+        `Al Fouad transaction not found for referenceNumber ${referenceNumber}`,
+      );
+    }
+
+    const newTransactionStatus = mapAlfouadStateToTransactionStatus({
```
RubenGeo: Is it possible to trigger different states somehow with the staging environment of Al Fouad?, just wondering if they for example send some different body back or something on different states. 

The doc doesn't show an actual example response for State=4 (Hold) or State=5 (Canceled), so it's possible Message carries more detail in practice than the doc's Approved example suggests. (in section 4.7)

Also in general it's good to validate the non-happy flow against the staging environment

services/121-service/src/fsp-integrations/settings/alfouad/alfouad-env-variable-settings.const.ts
```diff
@@ -0,0 +1,11 @@
+import { alfouadEnvVariablesSchema, env } from '@121-service/src/env';
+import { getEnvVariablesForFsp } from '@121-service/src/fsp-integrations/settings/get-env-variables-for-fsp';
+import { FspEnvVariablesDto } from '@121-service/src/fsp-integrations/shared/dto/fsp-env-variables.dto';
+
+export const ALFOUAD_ENV_VARIABLE_SETTINGS: FspEnvVariablesDto = {
```
RubenGeo, line 5: alfouadEnvVariablesSchema bundles the account/branch/username/password/public-key vars in with ALFOUAD_MODE/ALFOUAD_API_URL, and ALFOUAD_ENV_VARIABLE_SETTINGS uses all of them for the required-at-startup check. But AlfouadService.getAlfouadFspConfig reads those credentials from ProgramFspConfigurationRepository, not env, so in production they're never actually used at runtime.

So as it's written now, enabling ALFOUAD_MODE=EXTERNAL in production without also setting these env vars will just fail app startup, while the intent is that a user configures these per program via the portal.

MTN ran into the same thing and solved it by splitting the DB-backed credentials into a separate mtnOptionalEnvVariablesSchema, excluded from MTN_ENV_VARIABLE_SETTINGS's required-variables list. I think Al Fouad could do the same for ALFOUAD_ACCOUNT, ALFOUAD_BRANCH_ID, ALFOUAD_USERNAME, ALFOUAD_PASSWORD, ALFOUAD_PUBLIC_KEY. That solution is probably GEFN

Separately, I think it'd be worth a follow-up task to make the distinction between "optional FSP env vars" and "required-at-startup FSP env vars" more explicit/robust in general, instead of relying on which schema object a var happens to be placed in. I think we also had a discussion on this once during architecture chit-chat.

services/121-service/src/fsp-integrations/settings/alfouad/alfouad-settings.const.ts
```diff
@@ -0,0 +1,57 @@
...
+      isRequired: true,
+    },
+    {
+      name: FspAttributes.phoneNumber,
+      isRequired: true,
+    },
+    {
+      name: FspAttributes.addressCity,
+      isRequired: true,
+    },
```
RubenGeo: I don't think this is used at the moment in the Al Fouad integration, so I think it can be removed from here

services/121-service/src/fsp-integrations/transaction-jobs/services/transaction-jobs-alfouad.service.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    service = new TransactionJobsAlfouadService(
+      alfouadService,
+      transactionJobsHelperService,
+      transactionEventScopedRepository,
+      transactionsService,
+    );
+  });
+
+  describe('processTransactionJob', () => {
```
RubenGeo: It would be better to not have function names here in describe (in accordance to the [guidelines](https://github.com/global-121/121-platform/wiki/Guidelines-Software-Architecture#-describing-behavior))

services/121-service/src/fsp-integrations/reconciliation/alfouad/alfouad-reconciliation.service.ts
```diff
@@ -0,0 +1,110 @@
...
+    const failedTransactionAttempts =
+      await this.transactionEventScopedRepository.countFailedTransactionAttempts(
+        transactionId,
+      );
+
+    return computeTransactionReference({
+      referenceId,
+      transactionId,
+      failedTransactionAttempts,
+    });
```
RubenGeo: You could move generateReferenceNumber from TransactionJobsAlfouadService to AlfouadService and than use it in both the TransactionJobsAlfouadService and here in the AlfouadReconciliationService, to make this a bit more dry

services/121-service/src/payments/services/transaction-jobs-creation.service.ts
```diff
@@ -650,6 +659,48 @@ export class TransactionJobsCreationService {
...
+      registrationViews.map(
+        (registrationView): AlfouadTransactionJobDto => {
+          const base = sharedJobsByReferenceId.get(
+            registrationView.referenceId,
+          );
+          return {
+            ...base!,
+            beneficiaryFullName: registrationView[FspAttributes.fullName]!,
+            beneficiaryPhoneNumber:
+              registrationView[FspAttributes.phoneNumber]!,
```
RubenGeo: I think it would be better to do this mapping to `beneficiary` variable names in a deeper layer as it naming specific to the Al Fouad api. 

probably a bit better to leave it here on 


```suggestion
            fullName: registrationView[FspAttributes.fullName]!,
            phoneNumber:
              registrationView[FspAttributes.phoneNumber]!,
```

or 
```suggestion
            registrationFullName: registrationView[FspAttributes.fullName]!,
            registrationPhoneNumber:
              registrationView[FspAttributes.phoneNumber]!,
```
If you need this clearer distinction between the senderFullName used later on

RubenGeo (changes requested):
Great work, curious to see the integration tests 👍 

I think when that is finished I'll also do some functional test to see if I can get it running locally against the staging environment

services/121-service/src/fsp-integrations/shared/enum/fsp-name.enum.ts
```diff
@@ -10,4 +10,5 @@ export enum Fsps {
   nedbank = 'Nedbank',
   onafriq = 'Onafriq',
   mtn = 'MTN',
+  alfouad = 'AlFouad',
```
RubenGeo, line 13:
```suggestion
  alfouad = 'Al-fouad',
```
Seem more consitent

RubenGeo: This also made me realize the class/file naming should follow the same multi-word convention as Cooperative Bank of Oromia / Commercial Bank Ethiopia — i.e. AlFouad (not Alfouad) for class/identifier names, and al-fouad (not alfouad) for file and folder names.

If we want to apply that, it's probably better done as a separate follow-up task once everything here is merged into main, rather than now — a rename across ~45 files would create a large, disruptive diff on top of your current work.

services/121-service/src/fsp-integrations/reconciliation/alfouad/alfouad-reconciliation.service.ts
Yuri-510: I tried asking them during our call, but got a vague response. I will ask them on Whatsapp again.

services/121-service/src/fsp-integrations/shared/enum/fsp-name.enum.ts
Yuri-510: I created an extra task

RubenGeo (approved):

PR #8682, opened 2026-08-20
chore: redact sensitive alfouad properties

Yuri-510:
[AB#43978](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/43978) 

## Describe your changes

Added a class `SensitiveValue`. Sensitive values can now be "annotated" (and therefore recognized as such) in the payload as: `secretName: new SensitiveValue(nameValue)`

## Portal preview-deployment


services/121-service/src/shared/services/custom-http.service.ts
```diff
@@ -418,6 +438,9 @@ export class CustomHttpService {
     return cloneDeepWith(
       data,
       (value: unknown, key: string | number | undefined) => {
+        if (value instanceof SensitiveValue) {
+          return '**REDACTED**';
+        }
         if (isSensitiveProperty(key)) {
```
elwinschmitz:
```suggestion
        if ((value instanceof SensitiveValue) || isSensitiveProperty(key)) {
```

?

services/121-service/src/shared/services/custom-http.service.ts
```diff
@@ -418,6 +438,9 @@ export class CustomHttpService {
     return cloneDeepWith(
       data,
       (value: unknown, key: string | number | undefined) => {
+        if (value instanceof SensitiveValue) {
```
RubenGeo: I think we could consider a distinction between:
1. username/password type of values
2. and PII that we do not want to log because of local laws. 

For username password it's good if we redact everything.

But for PII it is maybe valuable if we can maybe retain like the last 3 values of a phonenumber/name in our logging so the logs are a bit more valuable while debugging. I think this depends on the correspondence we hear back from our email conversation. 

I don't think this is blocking for now as we could add this distinction later in a new PR

RubenGeo (changes requested):
Nice solution 👍 ,  just elwin's comment. My comment is not blocking atm

services/121-service/src/shared/services/custom-http.service.ts
Yuri-510: created an extra task for this

RubenGeo (dismissed):

RubenGeo (approved):

PR #8698, opened 2026-08-25
Alfouad integration tests

Yuri-510:
[AB#44001](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44001) 

## Describe your changes

Alfouad integration tests

## Portal preview-deployment


services/121-service/src/seed-data/program/program-alfouad.json
```diff
@@ -0,0 +1,101 @@
...
+      "label": {
+        "en": "City"
+      },
+      "type": "text",
+      "options": null,
+      "includeInTransactionExport": true,
+      "scoring": {},
+      "editableInPortal": true,
+      "showInPeopleAffectedTable": false
+    }
```
RubenGeo: We don't need address city I think

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
+import { HttpStatus } from '@nestjs/common';
```
RubenGeo: I think we should still have unit test for some things in the alfouad-reconciliation.service.ts as they are not hit by the integration test

- getTransactionStateByRef returns undefined → throws AlfouadApiError → caught, logged, loop continues to the next transaction. 
- Non-AlfouadApiError is rethrown and aborts the batch.

services/121-service/src/seed-data/program/program-alfouad.json
```diff
@@ -0,0 +1,101 @@
...
+          "name": "branchIdAlfouad",
+          "value": "1"
+        },
+        {
+          "name": "usernameAlfouad",
+          "value": "Red Crescent"
+        },
+        {
+          "name": "passwordAlfouad",
+          "value": "mock-password"
```
RubenGeo: You can use this here `ALFOUAD_PASSWORD` to seed the values that are in the .env from the user. 
It should match the env variable name. We also have this pattern for program-nlrc-pv.json for example.

This way you could have the actual staging environment values in your .env and when you than hit the database reset endpoint you do not need to add them manually through the program fsp configuration interface.

This also applies to the other env variables in this file

services/mock-service/src/fsp-integration/alfouad/alfouad.mock.service.ts
```diff
@@ -0,0 +1,121 @@
...
+  [AlfouadMockPhoneNumber.stateApproved, AlfouadMockTransactionState.approved],
+  [AlfouadMockPhoneNumber.stateHold, AlfouadMockTransactionState.hold],
+  [AlfouadMockPhoneNumber.stateCanceled, AlfouadMockTransactionState.canceled],
+]);
+
+@Injectable()
+export class AlfouadMockService {
+  // In-memory only: correlates created transactions with later status lookups,
+  // as the real Al Fouad API does. Nothing is persisted.
+  private readonly transactionStateByReferenceNumber = new Map<
```
RubenGeo: Our vision on the Mock service is that it is supposed to be stateless. The reason for that is that there is no hidden state that can leak between tests or cause flaky tests depending on execution order. Also when somebody like account management is giving a demo this stateful approach can result in an error when they import somebody with the same referenceId after they reset the 121-service and not the mock-service.

The other FSP mocks (Airtel, Safaricom, Onafriq, Nedbank, Coop Bank of Oromia, MTN) use magic input values to control scenario outcomes instead of storing state.

This transactionStateByReferenceNumber Map makes Al Fouad stateful again, which also introduces statefulness between requests.

I think we can use a similair approach here as we have for MTN where we hack the reference id creation in mock mode:

Image

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
...
+
+    const transactionEventDescriptions = await getTransactionEventDescriptions({
+      programId,
+      transactionId: transaction.id,
+      accessToken,
+    });
+    expect(transactionEventDescriptions).toEqual(expectedEventsForSyncError);
+  });
+
+  it('should keep the transaction on waiting while Al Fouad reports a non-final state', async () => {
```
RubenGeo: So when I was reviewing the Airtel FSP integration that Niels made we had a few discussions around the balance between integration and unit test for FSP integrations. 

The conclusion was sort of that Integration Tests should be used to proves something than cannot be done in a unit test so database logic, http logic and database states that effect the outcome of later logic in the code. 

In this case I think this test does not add that much value above the test that we already have in `should successfully initiate a payment and reconcile it to success` it only checks that the newTransactionStatus in reconcileTransaction is waiting the saveProgressFromExternalSource should not be called. Which I think we could do in a unit test instead

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
...
+
+    const transactionEventDescriptions = await getTransactionEventDescriptions({
+      programId,
+      transactionId: transaction.id,
+      accessToken,
+    });
+    expect(transactionEventDescriptions).toEqual(expectedEventsForSyncError);
+  });
+
+  it('should set the transaction to error when Al Fouad reports it as canceled', async () => {
```
RubenGeo: I think the same comments applies here https://github.com/global-121/121-platform/pull/8698/changes#r3854872012

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
...
+    const paymentId = doPaymentResponse.body.id;
+
+    await waitForPaymentAndTransactionsToComplete({
+      programId,
+      paymentReferenceIds,
+      paymentId,
+      accessToken,
+      maxWaitTimeMs: 10_000,
+      completeStatuses: [TransactionStatusEnum.waiting],
+    });
```
RubenGeo: We can use doPaymentAndWaitForCompletion here instead which does both things. Also in a few other places in this test file

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
...
+
+    const transactionEventDescriptions = await getTransactionEventDescriptions({
+      programId,
+      transactionId: transaction.id,
+      accessToken,
+    });
+    expect(transactionEventDescriptions).toEqual(expectedEventsForSuccess);
+
+    // A second reconciliation run should not touch the completed transaction
+    await runCronJobDoAlfouadReconciliation();
```
RubenGeo: If the cronjob returned the number of transactions still not in a final state (like MTN's does), we could assert on that here too.

So the first time we hit the cronjob it should return 1 and the second time it should return 0.

services/121-service/test/payment/fsp-integrations/do-payment-fsp-alfouad.test.ts
```diff
@@ -0,0 +1,509 @@
...
+
+    await waitForPaymentAndTransactionsToComplete({
+      programId,
+      paymentReferenceIds,
+      paymentId,
+      accessToken,
+      maxWaitTimeMs: 10_000,
+      completeStatuses: [TransactionStatusEnum.success],
+    });
```
RubenGeo:
```suggestion
```
I think this part is a bit redundant with the first test, where we already asserted moving from waiting to success after the cron

Could be that I am missing something, so please push back if that is the case

RubenGeo (changes requested):

RubenGeo (approved):

PR #8705, opened 2026-08-26
chore: harden alfouad redaction

Yuri-510:
[AB#44138](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44138) 

## Describe your changes

- Converted the auth identity interface to a class (`alfouad-auth-identity.class.ts`) whose `toJSON()` returns '**REDACTED**'. The credential fields stay fully readable for building the auth header, but any accidental serialization (`JSON.stringify`, an error message, another logger, or nested inside a logged payload) now yields **REDACTED**

- Added authorization to the redaction key-list in `custom-http.service.ts` and made the key matching case-insensitive. So if anyone ever adds header logging (or routes headers through `redactSensitiveDataProperties`), Authorization: `Bearer …` is redacted regardless of casing.

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/fsp-integrations/integrations/alfouad/interfaces/alfouad-auth-identity.class.spec.ts
+++ b/services/121-service/src/fsp-integrations/integrations/alfouad/interfaces/alfouad-auth-identity.class.spec.ts
@@ -1,0 +1,31 @@
+import { AlfouadAuthIdentity } from '@121-service/src/fsp-integrations/integrations/alfouad/interfaces/alfouad-auth-identity.class';
+
+describe('AlfouadAuthIdentity', () => {
+  const authIdentity = new AlfouadAuthIdentity({
+    account: '161010004501',
+    branchId: '1',
+    username: 'Red Crescent',
+    password: 'secret',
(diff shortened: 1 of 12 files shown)


RubenGeo:
I think this PR is a bit of overengineering. 
- We currently do not log headers of our outgoing API requests
- The password value is already RSA encrypted
- For other FSP we do not have something similar

If we want this, I think we should implement it in a more generic way. 

We could consider the like hidden classes that Niels demonstraded in the latest lightning talk. Have the ProgramFspConfigurationRepository public methods convert properties to SecretConfigValue based on FspConfigurationPropertyVisibilityMap. and based on that make decisions in the http service. But I do not think this feature is a must for the current work item

Regardless of the above, the custom-http.service.ts changes (adding `authorization` to the redacted keys and making the key matching case-insensitive) look good to me as-is and I'd be fine merging those independently of the AlfouadAuthIdentity class discussion.

RubenGeo (changes requested):
see comment

PR #8736, opened 2026-08-28
chore(test): Fix flaky export payment check

RubenGeo:
[AB#44191](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44191)

## Describe your changes

Fix flaky

## Portal preview-deployment

### DIFF
--- a/e2e/portal/tests/ViewPayments/ExportPayments.spec.ts
+++ b/e2e/portal/tests/ViewPayments/ExportPayments.spec.ts
@@ -103,7 +103,6 @@ test('View available actions for a "view only" user', async ({
 
     await loginPage.login({
       skipNavigateToLogin: true,
-      skipUrlCheck: true,
       username: env.USERCONFIG_121_SERVICE_EMAIL_USER_VIEW ?? '',
       password: env.USERCONFIG_121_SERVICE_PASSWORD_USER_VIEW ?? '',
     });


elwinschmitz (commented):
I think the idea around that `skip`-flag was to "speed things up", as its not really important _in this case_ that "the URL matches something specific"... because the next step in the test is doing a navigation to another page again.

So I'm really wondering if this _really_ reduces the flakyness...
Unless the bug is actually _INSIDE_ the `login`-method's logic around the `skipUrlCheck`-flag... But then I would expect some code-change there...

~~Can you (let some robot) explain this?~~ (already checked)

elwinschmitz (changes requested):
I think the idea around that `skip`-flag was to "speed things up", as its not really important _in this case_ that "the URL matches something specific"... because the next step in the test is doing a navigation to another page again.

So I'm really wondering if this _really_ reduces the flakyness...
Unless the bug is actually _INSIDE_ the `login`-method's logic around the `skipUrlCheck`-flag... But then I would expect some code-change there...

@copilot: Can you explain this?

elwinschmitz:
I think the idea around that `skip`-flag was to "speed things up", as its not really important _in this case_ that "the URL matches something specific"... because the next step in the test is doing a navigation to another page again.

So I'm really wondering if this _really_ reduces the flakyness...
Unless the bug is actually _INSIDE_ the `login`-method's logic around the `skipUrlCheck`-flag... But then I would expect some code-change there...

@copilot: Can you explain this?

PR #8770, opened 2026-09-02
fix: Scope column appears twice

nameerhqureshi:
[AB#44269](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44269) 

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/kobo/services/kobo.service.spec.ts
+++ b/services/121-service/src/kobo/services/kobo.service.spec.ts
@@ -14,7 +14,10 @@ import { ProgramRegistrationAttributesService } from '@121-service/src/program-r
 import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
 import { ProgramService } from '@121-service/src/programs/programs.service';
 import { ProgramRepository } from '@121-service/src/programs/repositories/program.repository';
-import { RegistrationAttributeTypes } from '@121-service/src/registration/enum/registration-attribute.enum';
+import {
+  GenericRegistrationAttributes,
+  RegistrationAttributeTypes,
+} from '@121-service/src/registration/enum/registration-attribute.enum';
 import { RegistrationPreferredLanguage } from '@121-service/src/shared/enum/registration-preferred-language.enum';
 
 describe('KoboService', () => {
@@ -359,6 +362,16 @@ describe('KoboService', () => {
           name: 'fsp', // This should be filtered out
           type: RegistrationAttributeTypes.text,
           label: { en: 'Financial Service Provider' },
+        },
+        {
+          name: GenericRegistrationAttributes.scope, // This should be filtered out as it maps to the entity-level registration.scope column
+          type: RegistrationAttributeTypes.text,
+          label: { en: 'Scope' },
+        },
+        {
+          name: GenericRegistrationAttributes.maxPayments, // This should be filtered out as a generic registration attribute
+          type: RegistrationAttributeTypes.numeric,
+          label: { en: 'Maximum payments' },
         },
       ];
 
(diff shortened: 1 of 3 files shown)


services/121-service/src/kobo/services/kobo.service.spec.ts
```diff
@@ -360,6 +360,11 @@ describe('KoboService', () => {
           type: RegistrationAttributeTypes.text,
           label: { en: 'Financial Service Provider' },
         },
+        {
+          name: 'Scope', // This should be filtered out as it maps to the entity-level registration.scope column
```
elwinschmitz: Should it be "capital-S"-scope? Not `scope`?

So: 

```suggestion
          name: GenericRegistrationAttributes.scope, // This should be filtered out as it maps to the entity-level registration.scope column
```

services/121-service/src/migration/1788361774000-remove-duplicate-scope-registration-attributes.ts
```diff
@@ -0,0 +1,22 @@
...
+       WHERE "programRegistrationAttribute"."programId" = "program"."id"
+         AND LOWER("programRegistrationAttribute"."name") = $1
+         AND "program"."enableScope" = true`,
+      ['scope'],
+    );
+  }
+
+  public down(): Promise<void> {
+    return Promise.resolve();
+  }
```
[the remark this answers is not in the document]
elwinschmitz: We should maybe find another way to 'formalise' this "we don't do down-migrations"-rule...

Because now all migration-scripts each have their own fancy way of "not doing the thing that is required somewhere else"...

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -319,9 +320,16 @@ export class KoboService {
       return attr.name !== fspQuestionName;
     };
 
+    // Filter out 'scope' as its value is stored on the entity-level registration.scope column,
+    // not as a program registration attribute
+    const isNotScope = ({ name }: { name: string }) => {
+      return name.toLowerCase() !== GenericRegistrationAttributes.scope;
```
elwinschmitz: Could we make this into a more 'generic' check?

Like: "we should not be touching _any_ `GenericRegistrationAttribute`", so:

`.filter(isNotGenericAttribute)` and:

```suggestion
    const isNotGenericAttribute = ({ name }: { name: string }) => {
      return !Object.values(GenericRegistrationAttributes).includes(name.toLowerCase());
```
(Or similar; I haven't checked this syntax)

Also;
I'm _not_ really a fan of these "double negative" method-names + return-statements, but maybe we should _not_ worry about it in this case too much, as the alternative would be maybe more confusing/verbose, I'd guess... You'd think _not_? Or would you not think this wouldn't be a problem we will not have? :P

RubenGeo: > Like: "we should not be touching any GenericRegistrationAttribute", so: so: .filter(isNotGenericAttribute)


Agreed with this

services/121-service/src/kobo/services/kobo.service.ts
```diff
@@ -319,9 +320,20 @@ export class KoboService {
...
     };
 
+    const genericRegistrationAttributeNames = new Set(
+      Object.values(GenericRegistrationAttributes).map((name) =>
+        name.toLowerCase(),
+      ),
+    );
+
+    const isCustomRegistrationAttribute = ({ name }: { name: string }) => {
+      return !genericRegistrationAttributeNames.has(name.toLowerCase());
```
elwinschmitz, line 330: Aha. yes. "its custom, when its not generic". 👍 

(This in combination with the `isNotFsp` and the other `isNot..` is _still_ somewhat confusing.. But lets tackle that some other time.. ;) )

elwinschmitz (approved):
If @RubenGeo gave his blessing on the migration-script: yay! \o/

RubenGeo (approved):

PR #8834, opened 2026-09-11
refactor: Use a stable sequence for the computed "RegistrationProgramId"s

RubenGeo:
[AB#44393](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44393) 

## Describe your changes

Okay in the end I went the more robust but more code route. This guarantees parallel processes can never be assigned the same registrationProgramId.

## Portal preview-deployment

### DIFF
--- a/services/121-service/src/programs/entities/program-registration-program-id-counter.entity.ts
+++ b/services/121-service/src/programs/entities/program-registration-program-id-counter.entity.ts
@@ -1,0 +1,24 @@
+import { Column, Entity, Index, JoinColumn, OneToOne, Relation } from 'typeorm';
+
+import { Base121Entity } from '@121-service/src/base.entity';
+import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
+
+@Entity('program_registration_program_id_counter')
+export class ProgramRegistrationProgramIdCounterEntity extends Base121Entity {
+  @OneToOne(
+    () => ProgramEntity,
+    (program) => program.registrationProgramIdCounter,
+    {
+      onDelete: 'CASCADE',
+    },
+  )
+  @JoinColumn({ name: 'programId' })
+  public program: Relation<ProgramEntity>;
+
+  @Index({ unique: true })
+  @Column({ type: 'int', nullable: false })
+  public programId: number;
+
+  @Column({ type: 'int', default: 0, nullable: false })
+  public lastRegistrationProgramId: number;
+}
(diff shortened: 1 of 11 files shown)


services/121-service/src/programs/entities/program-registration-program-id-counter.entity.ts
```diff
@@ -0,0 +1,24 @@
+import { Column, Entity, Index, JoinColumn, OneToOne, Relation } from 'typeorm';
+
+import { Base121Entity } from '@121-service/src/base.entity';
+import { ProgramEntity } from '@121-service/src/programs/entities/program.entity';
+
+@Entity('program_registration_program_id_counter')
+export class ProgramRegistrationProgramIdCounterEntity extends Base121Entity {
```
elwinschmitz: I think there is still time to come up with a _better_ name for this new magical entity... ;)

"ProgramRegistrationProgramId" isn't really helpful...

I guess its also part of our public API (property/parameter names etc.) So maybe not possible to rename _all that_...

But this name is just "hard to reasonably explain"...

What about "(Program)RegistrationIndex(Entity)"? Or "(Program)RegistrationOrder(Entity)"?

services/121-service/test/registrations/assign-registration-program-id.test.ts
```diff
@@ -40,16 +47,52 @@ const createRegistrations = ({
...
+  const programId = createProgramResponse.body.id;
+
+  await postProgramFspConfiguration({
+    programId,
+    body: newProgramFspConfiguration,
+    accessToken,
+  });
+
+  return programId;
+};
```
elwinschmitz, line 83: Maybe we should be 'upgrading' these helpers an let every test-suite create '_their own programs_'... (Not necessarily in _this_ PR...)

Maybe that is enough to run our tests concurrently... (also a good load-test at the same time. :D )

elwinschmitz (changes requested):
"Naming things is hard". ;)

services/121-service/src/programs/entities/program-registration-program-id-counter.entity.ts
RubenGeo: Good one, I am going for RegistrationProgramIdSequenceEntity. The word program is very double.

I prefer Sequence over index as an index is something different. I think we should explicitly keep the word Program in the entity name, as the purpose of this entity is to have a sequence per program

RubenGeo: > I guess its also part of our public API (property/parameter names etc.) So maybe not possible to rename all that..

Yes we cannot 'just' rename RegistrationEntity.registrationProgramId.

services/121-service/test/registrations/assign-registration-program-id.test.ts
RubenGeo: This is what I usually do when I do not want tests to restart the DB between every run. I could have co-pilot draft a PR to make these re-usable helpers after this gets merged in

elwinschmitz (approved):

PR #8842, opened 2026-09-14
fix: [Possible BUG] - Cannot Link Kobo Form to 121 When There Are Mul…

nameerhqureshi:
[AB#44467](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44467) 

## Portal preview-deployment


nielsVoogt (dismissed):

services/121-service/src/migration/1789000000000-add-program-registration-attributes-read-permission.ts
```diff
@@ -0,0 +1,43 @@
...
+      FROM "121-service"."user_role" ur, "121-service"."permission" p
+      WHERE ur."role" IN (
+        'program-admin',
+        'view',
+        'cva-manager',
+        'cva-officer',
+        'finance-manager',
+        'finance-officer',
+        'approver'
+      )
```
RubenGeo: I think it would be better to backfill against any role that currently has the RegistrationPersonalREAD, instead of listing the default role here. So any custom roles that have been created by our users also get this permission.

services/121-service/test/program/get-program-attributes.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  accessToken: string;
+}) {
+  return getServer()
+    .get(`/programs/${id}/attributes`)
+    .query({ includeProgramRegistrationAttributes: true })
+    .set('Cookie', [accessToken]);
+}
+
+describe('Get program attributes', () => {
+  beforeEach(async () => {
```
RubenGeo: We don't currently have integration tests that verify the permission configuration on every individual endpoint.

The internal logic of the permission check (in cookie-jwt.strategy.ts / the .strategy.ts files) is covered by tests, but which specific permission(s) each endpoint's @AuthenticatedUser() decorator requires isn't tested per-endpoint.

Doing that consistently would mean adding an integration test for every @AuthenticatedUser({ permissions: [...] }) usage across the codebase, which I think is not really worth it.

RubenGeo (changes requested):

services/121-service/src/migration/1789000000000-add-program-registration-attributes-read-permission.ts
nameerhqureshi: It was with the 2 permissions RegistrationREAD and RegistrationPersonalREAD. Do you think I use both or single RegistrationPersonalREAD would be enough?

RubenGeo: I think personal read, because that is related to program registation attribute entities. Regsitration.read is more related to reading the properties on the registartion entity.

RubenGeo (approved):

PR #8850, opened 2026-09-15
Add onafriq currency code per program config

RubenGeo:
[AB#44605](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44605) 

## Describe your changes

Add onafriq currency code per program config

## Portal preview-deployment

https://happy-rock-0411d2003-8850.westeurope.3.azurestaticapps.net


services/121-service/src/migration/1789475507841-add-currency-code-onafriq-fsp-config-prop.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+import { env } from '@121-service/src/env';
+
+export class AddCurrencyCodeOnafriqFspConfigProp1789475507841 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    if (!env.ONAFRIQ_CURRENCY_CODE) {
+      // No Onafriq currency configured in this environment, so there is nothing to backfill
+      return;
+    }
```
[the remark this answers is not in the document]
RubenGeo: this currency code is there, else onafriq would never have worked

services/121-service/src/fsp-integrations/reconciliation/onafriq/onafriq-reconciliation.mapper.ts
```diff
@@ -6,6 +6,7 @@ export class OnafriqReconciliationMapper {
   public static mapTransactionToReportItem(
     onafriqTransaction: OnafriqTransactionEntity,
     corporateCode: string,
+    currencyCode: string,
```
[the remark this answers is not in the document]
RubenGeo: resolved

interfaces/portal/src/app/domains/fsp-configuration/fsp-configuration.helper.ts
```diff
@@ -24,6 +24,7 @@ export const FSP_CONFIGURATION_PROPERTY_LABELS: Record<
   [FspConfigurationProperties.corporateCodeOnafriq]: $localize`Corporate code`,
   [FspConfigurationProperties.passwordOnafriq]: $localize`Password`,
   [FspConfigurationProperties.uniqueKeyOnafriq]: $localize`Unique key`,
+  [FspConfigurationProperties.currencyCodeOnafriq]: $localize`Currency code`,
```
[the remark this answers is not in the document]
RubenGeo: resolved

nielsVoogt (approved):
I'm not seeing a whole lot that can go wrong here, but I recommend a second review/approval by another team member.

services/121-service/test/onafriq/export-reconciliation-report.test.ts
```diff
@@ -16,6 +16,7 @@ import {
 describe('Export reconciliation report', () => {
   const programId = 1;
   const transferValue = 12327;
+  const currencyCodeOnafriq = env.ONAFRIQ_CURRENCY_CODE;
```
elwinschmitz, line 19: Do we still need the ENV-variable, _just for this test_?

I'd think its safe to hard-code a value in here.
And get rid of the whole `env.ts`-part for this variable.

services/121-service/src/fsp-integrations/shared/consts/fsp-configuration-property-types.const.ts
```diff
@@ -12,6 +12,7 @@ export const fspConfigurationPropertyTypes = {
   [FspConfigurationProperties.corporateCodeOnafriq]: 'string',
   [FspConfigurationProperties.passwordOnafriq]: 'string',
   [FspConfigurationProperties.uniqueKeyOnafriq]: 'string',
+  [FspConfigurationProperties.currencyCodeOnafriq]: 'string',
```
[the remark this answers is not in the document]
elwinschmitz: Will the UI for this FSP-configuration-setting, use the "Currency-dropdown"?

Or will it use "the currency configured on the program" instead, without the need for the UI-part?

services/121-service/src/fsp-integrations/reconciliation/onafriq/onafriq-reconciliation.service.ts
```diff
@@ -190,16 +190,25 @@ export class OnafriqReconciliationService {
...
+          FspConfigurationProperties.currencyCodeOnafriq,
+        ],
       });
-    const corporateCode = programFspConfigProperties[0].value as string;
+    const corporateCode = programFspConfigProperties.find(
+      (c) => c.name === FspConfigurationProperties.corporateCodeOnafriq,
+    )?.value as string;
+    const currencyCode = programFspConfigProperties.find(
+      (c) => c.name === FspConfigurationProperties.currencyCodeOnafriq,
+    )?.value as string;
```
elwinschmitz, line 203: This would be the correct place (on the edge) to map/parse this value into a `CurrencyCode` (from: `services/121-service/src/exchange-rates/enums/currency-code.enum.ts`), right?

So that `OnafriqReconciliationMapper.mapTransactionToReportItem` could 'just assume' that its value will be a "three uppercase characters"-string-representing-a-currency....

services/121-service/src/fsp-integrations/integrations/onafriq/interfaces/onafriq-request-identity.interface.ts
```diff
@@ -2,4 +2,5 @@ export interface OnafriqRequestIdentity {
   readonly corporateCode: string;
   readonly password: string;
   readonly uniqueKey: string;
+  readonly currencyCode: string;
```
elwinschmitz, line 5: Would/could/should this be a `CurrencyCode`? (See comment below.)

elwinschmitz (changes requested):

services/121-service/src/fsp-integrations/shared/consts/fsp-configuration-property-types.const.ts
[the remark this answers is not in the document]
RubenGeo: I think these are very good points @elwinschmitz, thanks for raising them.  

> Will the UI for this FSP-configuration-setting, use the "Currency-dropdown"?

No it's a free text field. We don't support the dropdown type for program fsp configuration properties. 

I also don't think it's worth it:
1. This interface is not often used
2. We don't know that OnAfriq supports a specific ISO code of currencies. Also I don't think they support all available currencies so I think a big dropdown is more confusing

> Or will it use "the currency configured on the program" instead, without the need for the UI-part?

I did think about that. However I think it's better to follow the normal process for program fsp configuration properties to make this more an explicit decision. 

In practice process for setting up the currency code is: Gabriele emailing OnAfriq what currency letters we should use and than him settings this up. Also I think they have to configure something on their side to set up this currency code. So using the currency code from the program feel more like a hidden decision, also because it differs per fsp if the currency code from a program is used or not

Let me know what you think about this reationale

services/121-service/test/onafriq/export-reconciliation-report.test.ts
RubenGeo: We also use it in services/121-service/src/seed-data/program/program-onafriq.json

services/121-service/src/seed-data/program/program-onafriq.json

In program seed files, property values matching environment variable names are dynamically resolved from process.env during seeding. Which makes it very easy when developing locally to have proper setup when you want to test with against staging/acceptance environment of an FSP, instead of that you have to reconfigure the FSP on every seed

services/121-service/src/fsp-integrations/reconciliation/onafriq/onafriq-reconciliation.service.ts
RubenGeo: If we go the the route of having a dropdown in the interface when selecting the currency, I think this should be a `as CurrencyCode`

If there is no validation when this input enters the system it's better to assume a string here. And we keep everything dumb and just pass along what the user put in there

services/121-service/src/fsp-integrations/integrations/onafriq/interfaces/onafriq-request-identity.interface.ts
RubenGeo: See other comment https://github.com/global-121/121-platform/pull/8850#discussion_r4035150419

services/121-service/test/onafriq/export-reconciliation-report.test.ts
elwinschmitz: Too bad we cannot remove more code :)

services/121-service/src/fsp-integrations/shared/consts/fsp-configuration-property-types.const.ts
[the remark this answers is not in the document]
elwinschmitz: As it is currently very tightly 'scoped/coupled' for Onafriq _only_, I don't think its a big risk to us it this way..

I was hoping we could 'align' some of these FSP-integrations a bit with some 'shared' stuff... That would be nice... It would _also_ have a 'more complexity'-downside, but still. ;)

We can always 'align more' later, if there are more (clear/important) use-cases indeed.

RubenGeo:
image
I think you still need to approve it @elwinschmitz

elwinschmitz (approved):

PR #8857, opened 2026-09-16
feat: Add transferValue to registration view

RubenGeo:
[AB#44658](https://dev.azure.com/redcrossnl/7d17f2a9-3f18-40f1-8c29-98568fbca7e4/_workitems/edit/44658) 

## Describe your changes

- Add a computed transferValue column (paymentAmountMultiplier * program fixedTransferValue) to the registration view, also exposed via the API
- The refinement also mentioned `Also use this when creating transactions ` however I decided against this
  - It's possible to pass in the transfer amount when create a payment (dry run). So than the transferValue in the registraiton view cannot be used because than the program fixedTransferValue is not perse used.

## Portal preview-deployment

https://happy-rock-0411d2003-8857.westeurope.3.azurestaticapps.net


RubenGeo:
Re: the "Nit" suggestion on `filter-operation.const.ts:38` to add dedicated filter/sort test cases for `transferValue` — I considered this but decided against adding them, and removed the two I'd initially added (`sort-registration.test.ts` / `filter-registration-range.test.ts`).

Reasoning:
- `registration_view` is a real Postgres `VIEW`, so once created, `transferValue` is just a normal output column of that view. Sorting/filtering on it goes through the exact same nestjs-paginate + TypeORM path as any plain column like `paymentAmountMultiplier` — there's no distinct "computed column" mechanism being exercised at query time.
- `transferValue = paymentAmountMultiplier × program.fixedTransferValue`, and for the test program `fixedTransferValue` is a fixed positive constant. Multiplying by a positive constant is a strictly monotonic transform, so a dedicated DESC-sort test on `transferValue` produces the *identical* ordering as the existing `paymentAmountMultiplier` sort test on the same registrations, and a dedicated `GT` filter test selects the *identical* reference IDs as the existing `paymentAmountMultiplier` `GT` test. They don't exercise anything `transferValue`-specific — not the formula, not null-handling (the one real behavioral difference, since `transferValue` can be `null` when `fixedTransferValue` is unset, while `paymentAmountMultiplier` never is — neither test happens to touch that case anyway).

What still covers this newly-exposed API surface without duplicating existing coverage:
- The `transferValue` entry now added to `filterAssertionConfig` in `filter-registration.test.ts`, which smoke-tests filtering on every generic filterable attribute (including this one).
- `calculate-payment-amount-multiplier.test.ts`, which is the actual unique test of the `transferValue` formula.
- `pagination-data.ts`'s `expectedAttributes`, which asserts `transferValue` is present in the API response shape.
- `create-program.test.ts`'s snapshot, which confirms `transferValue` is correctly registered as filterable/sortable metadata.

Happy to add the dedicated tests back if there's a scenario I'm missing, but as-is they felt like pure duplication of the `paymentAmountMultiplier` coverage rather than new confidence.

